PR #440, opened 2022-10-18
Feature/cloudinary react sdk v2

mattwendzina:
Closes # N/A

## 📝 Description

This is the second implementation option for images. Cloudinary have commented that this appears to be a great workflow option. 

Rather than marketing creating transformations on the fly in the Cloudinary portal, they would simply choose the images they want and then instruct us in Sanity what changes they want. These will then show up in the preview window. 

I have created a group of test transformations in Sanity as examples of how this would work but don't presume that they would want to be able to add effects (which is one of the transformation tests I've included). We would need to discuss with them what they wanted and add these in manually.

This option enables us to take advantage of their <AdvancedImage /> component and the plugins. There are four - 1) Lazy loading, 2) Responsive, 3) Placeholder, 4) Accessibility. In reality we would only use Responsive but could use accessibility too if we wanted. The lazy loading plugin won't work because Remix is SSR and Cloudinary transformations are client side. The placeholder functionality is being emulated in this PR due to another slight complication of using the responsive plugin. You can watch the attached video to understand this more.

N/A

## 🚀 New behavior

https://user-images.githubusercontent.com/36510083/196472002-7c7b6458-33af-403e-850e-5c428895deab.mp4


mattwendzina:
The file with the Cloudinary transformations could be a lot cleaner. And thinking about it, they really all need to be in one go otherwise they count as multiple transformations. If we think this approach is the best, which I think it is, then we can refine this.

stevenbister:
Sorry @mattwendzina don't hate me 😅 

I think I'm leaning away from this approach at the moment. 

I understand the approach but it feels like we're working around the platform by moving the controls into Sanity and that `cloudinary.ts` file has the potential to become a bit of a beast, I think.

I'm also not keen on limiting the freedom that the editors will have by moving the controls away from Cloudinary and I guess there could be some confusion around where to make edits as the Cloudinary editor will still have those fields available 🤔

I think with a bit of tweaking #434 would be my preferred approach at the moment. Ideally I'd just want to grab the transformed url from Cloudinary and pass it into a component.

We could put some more work into #434 and build our own component from the ground up, maybe taking inspiration from  ol' Kent?
https://kentcdodds.com/blog/building-an-awesome-image-loading-experience
https://github.com/kentcdodds/kentcdodds.com/blob/f46a387894db3569dca74ad30d7e4900c1a62a19/app/images.tsx

Alternatively we could look at [remix-image](https://remix-image.mcfarl.in/docs/intro) it's pretty straightforward and I've managed to throw together a (very) [small demo](https://codesandbox.io/s/spring-snowflake-inoipq?file=/app/routes/index.jsx). It would need some tweaking on the responsive images side of things. Might be less effort but is another dependency 🤷 

Not sure if @hi-im-si has any more thoughts?

hi-im-si:
> Sorry @mattwendzina don't hate me 😅
> 
> I think I'm leaning away from this approach at the moment.
> 
> I understand the approach but it feels like we're working around the platform by moving the controls into Sanity and that `cloudinary.ts` file has the potential to become a bit of a beast, I think.
> 
> I'm also not keen on limiting the freedom that the editors will have by moving the controls away from Cloudinary and I guess there could be some confusion around where to make edits as the Cloudinary editor will still have those fields available 🤔
> 
> I think with a bit of tweaking #434 would be my preferred approach at the moment. Ideally I'd just want to grab the transformed url from Cloudinary and pass it into a component.
> 
> We could put some more work into #434 and build our own component from the ground up, maybe taking inspiration from ol' Kent? https://kentcdodds.com/blog/building-an-awesome-image-loading-experience https://github.com/kentcdodds/kentcdodds.com/blob/f46a387894db3569dca74ad30d7e4900c1a62a19/app/images.tsx
> 
> Alternatively we could look at [remix-image](https://remix-image.mcfarl.in/docs/intro) it's pretty straightforward and I've managed to throw together a (very) [small demo](https://codesandbox.io/s/spring-snowflake-inoipq?file=/app/routes/index.jsx). It would need some tweaking on the responsive images side of things. Might be less effort but is another dependency 🤷
> 
> Not sure if @hi-im-si has any more thoughts?

I would agree strongly with this. I think from having a play about with #434 and seeing you can add all the transformations via the cloudinary media manager in the CMS and return the transformed image to the frontend without the need for replicating all the extra controls - that was exactly how I was expecting this to work. 

As @stevenbister mentioned I think #434 just needs the extra responsive image work (`srcset` and `sizes`) in the same way as implemented by KCD - once that is done then this is pretty much good to go. An example of the output of KCD's image works as follows:

```html
<img 
    title="Image Title" 
    alt="This is an Alt Tag" 
    src="https://res.cloudinary.com/kentcdodds-com/image/upload/c_fill,w_955,ar_3:4,q_auto,f_auto,b_rgb:e6e9ee/unsplash/photo-1474718723952-48d2a016108f" 
    srcset="https://res.cloudinary.com/kentcdodds-com/image/upload/c_fill,w_280,ar_3:4,q_auto,f_auto,b_rgb:e6e9ee/unsplash/photo-1474718723952-48d2a016108f 280w, https://res.cloudinary.com/kentcdodds-com/image/upload/c_fill,w_560,ar_3:4,q_auto,f_auto,b_rgb:e6e9ee/unsplash/photo-1474718723952-48d2a016108f 560w, https://res.cloudinary.com/kentcdodds-com/image/upload/c_fill,w_840,ar_3:4,q_auto,f_auto,b_rgb:e6e9ee/unsplash/photo-1474718723952-48d2a016108f 840w, https://res.cloudinary.com/kentcdodds-com/image/upload/c_fill,w_1100,ar_3:4,q_auto,f_auto,b_rgb:e6e9ee/unsplash/photo-1474718723952-48d2a016108f 1100w, https://res.cloudinary.com/kentcdodds-com/image/upload/c_fill,w_1300,ar_3:4,q_auto,f_auto,b_rgb:e6e9ee/unsplash/photo-1474718723952-48d2a016108f 1300w, https://res.cloudinary.com/kentcdodds-com/image/upload/c_fill,w_1650,ar_3:4,q_auto,f_auto,b_rgb:e6e9ee/unsplash/photo-1474718723952-48d2a016108f 1650w" 
    sizes="(max-width:639px) 80vw, (min-width:640px) and (max-width:1023px) 40vw, (min-width:1024px) and (max-width:1620px) 25vw, 420px" 
    class="class-name-1  class-name-2  class-name-3" 
    loading="lazy">
```

I think based on the above I also don't feel the Remix Image component would technically be necessary here and its always good to knock-off a dependency too. Based on both myself and @stevenbister thoughts I feel we close this PR and continue with #434

mattwendzina:
So unfortunately this still does not seem to be simple = (

After rejigging kents utilities to get them to work I discovered that it still basically relies on using the Cloudinary ID in order to create the transformed URLs. In other words, in basically won't support our approach.
Eventually it runs the following on each image:
```
buildImageUrl(id, { transformations });
```
buildImageUrl is the Cloudinary plugin that builds the images based off of the ID and transformations.

So I've had a look at Remix-Images. I'm not convinced about how this works... I'm pretty sure it's not actually transforming the images with Cloudinary. I don't know exactly how it's built in transforming works but in the advanced section of the docs it talks about creating a custom transformer and that Cloudinary may be an example of when you'd want to do this:

https://github.com/Josh-McFarlin/remix-image/blob/master/docs/docs/advanced/creating-a-transformer.md

There is an example of using a custom transformer - there are a couple that have been built. One is for using Sharp - I've had a look at this and it also looks like it operates in a similar way to how Cloudinary transformations work. IE - Create a new instance of the image with Sharp and then tap into methods to transform them... 
https://github.com/Josh-McFarlin/remix-image/blob/master/packages/remix-image-sharp/src/sharpTransformer.ts

So this means we'd still need to pass in the image ID in order to get the image urls with all the different sizes which puts us back where we were = (

PR #443, opened 2022-10-19
Feature/sanity seo settings

### LINKED ISSUE #151: Title & meta description
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/150

### Description

n/a

### LINKED ISSUE #152: Social images
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/150

### Description

n/a

### LINKED ISSUE #155: Index - No index settings
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/150

### Description

n/a

stevenbister:
- Closes #151 
- Closes #152 
- Closes #155 

## 📝 Description

Adds SEO settings to Sanity and feeds them through to osc-ecommerce. I've separated them out into "global" settings which sit under the settings page:

image

The global settings include:
- site title
- title separator
- org name and logo
- org social channels
- global noindex toggle so we can block the entire site

And page level settings which sit on individual pages (home, pages, collections, etc.) I've also installed a plugin which is built on Yoast to give the site editors some SEO suggestions.

image

Page level settings include:
- title
- canonical url
- keywords & synonyms (this is specifically for the seo plugin and doesn't change anything on the front end)
- description
- social image
- page noindex toggle

You'll notice I've added an `Access-Control-Allow-Origin` response header to ecommerce. You can see this in the `entry.server.ts` file. This is because the seo plugin needs to fetch the site and as Sanity is an SPA it runs it in the browser which means... **CORS**! So there's a small script in there to generate the appropriate studio url.

N/A

## 🚀 New behavior

SEO fields within Sanity studio which pulls through to osc-ecommerce.

## 📝 Additional Information

I've added a couple of pages to Sanity; one with no SEO changes and one with all of the custom settings, so we can test the settings with Playwright. Please don't delete or edit those pages 😅 

As we don't have the deployment of Sanity in this branch if you wanna have a look at it locally, pull this branch and make sure to run both `osc-studio` and `osc-ecommerce`. If you aren't running `osc-ecommerce` then you will get a "failed to fetch" error when looking at the SEO pane in studio.


packages/osc-ecommerce/app/root.tsx
```diff
@@ -48,15 +50,38 @@ export const links: LinksFunction = () => {
     ];
 };
 
-export const meta: MetaFunction = () => ({
-    charset: 'utf-8',
-    title: 'OSC Academic Hub',
-    viewport: 'width=device-width,initial-scale=1'
-});
+export const meta: MetaFunction = ({ data }) => {
```
mattwendzina, line 53: Just a thought really on this. Wondering whether it reads better if loader functions come before the data that actually consumes them? I'm still getting my head around it all, but took a moment to realise the data here is what is being fed from the server/LoaderFunction (I think), so just wonder if it would read better if these sat above?

mattwendzina:
Added a couple of comments, but other than those LGTM

packages/osc-ecommerce/app/root.tsx
stevenbister: Yeah that's a good point, I do agree not sure why I left it this way around! 

I'll swap them over

packages/osc-ecommerce/app/utils/metaTags/buildHtmlMetaTags.ts
```diff
@@ -0,0 +1,41 @@
...
+        title = pageData?.seo?.title;
+    } else {
+        title = pageData?.title;
+    }
+
+    const description = pageData?.seo?.description;
+    const image = pageData?.seo?.image;
+
+    const meta: HtmlMetaDescriptor = {
+        title: `${title} ${globalData.titleSeparator} ${globalData.siteTile}`,
```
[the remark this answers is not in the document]
stevenbister: Good spot! They should be optionally chained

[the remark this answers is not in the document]
stevenbister: Done!

packages/osc-ecommerce/app/root.tsx
stevenbister: Done!

PR #454, opened 2022-10-24
Feature/redirects

### LINKED ISSUE #153: Redirects
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/150

### Description

n/a

stevenbister:
- Closes #153 

## 📝 Description

I've added a new document to Sanity called redirects which will allow us to create a redirect on a page by page basis. In Sanity it works a bit like adding new pages. I've done it this way rather than another tab in settings as it makes it easier to search and also gives editors a quick way to add them here:

https://user-images.githubusercontent.com/23461173/197565218-04d47206-2ca7-472e-81b6-e8545297d7f9.mov

N/a

## 🚀 New behavior

A new document type (redirects) has been added to `osc-studio` along with appropriate validation.

Within `osc-ecommerce/sanity.server` I've added a utility function to create the url paths. Which is used by a new function called `shouldRedirect()`.

`shouldRedirect()` is then used on each of the `routes`. 

If no page data is found we'll check if we shouldRedirect it. If `shouldRedirect` returns something then we run it, otherwise it'll return undefined and we throw a 404.

## 📝 Additional Information

You should be able to see the redirects in action by opening your network tab and navigating to https://pr-454-osc-ecommerce.fly.dev/what-page or https://pr-454-osc-ecommerce.fly.dev/blog/page-2.

https://user-images.githubusercontent.com/23461173/198242112-b639feb8-b2ae-418d-b799-45d6a359d373.mov

The Studio should be at https://pr-454-osc-studio.fly.dev


packages/osc-ecommerce/app/models/sanity-server.spec.ts
```diff
@@ -1,131 +1,206 @@
...
+                },
+                {
+                    _key: 'c342a5c0-0353-51fd-9983-6a30913bef08',
+                    _ref: 'shopifyProductVariant-42016931512575',
+                    _type: 'reference',
+                    _weak: true
+                }
+            ],
+            vendor: 'openstudydev'
+        });
```
Tom-Pearce, line 131: Is there a more dynamic way we can confirm the object returned contains relevant data? Just thinking going forward products are subject to a lot of regular change, particularly title, price, image url, tags and once live the `id`/`createdAt` fields will also change. Even if tests are only run against the Dev store/data, they are likely to change occasionally when testing other features.

Could cause headaches if we had to change this for every test - and changing it for every test would beat the point of the test since we'd probably use the API to dump this object out anyway?

Perhaps just checking certain key fields are present and are of the expected type (string/num/etc)?

stevenbister: Yeah that's a good point! I've updated the test to check for what I think will be the key properties and am now just checking that the value exists/matches a type.

Whatcha think?

Tom-Pearce: LGTM

Tom-Pearce (approved):

PR #469, opened 2022-10-27
Feature/content component

### LINKED ISSUE #221: [S13.1]: Content
### Link to Epic

### Description

Make a component using Sanity's portable text including:

- containers component (sm/md/lg)
- background colour
- padding/margin controls
- paragraphs
- headings
- blockquote
- image
- list items
- links
- buttons

### Linked Tasks

- [x] #222
- [x] #223 
- [ ] #600

### LINKED ISSUE #222: Content make dynamic with Sanity
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/221

### Description

n/a

### LINKED ISSUE #223: Build Content component
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/221

### Description

Add testing
E2E Integration

stevenbister:
- Closes #221 
- Closes #222 
- Closes #223 

## 📝 Description

Adds a `<Content />` component to `osc-ui` and a module to `osc-studio` where you can edit it.

The content component still needs a `<Button />` and `<Image />` component adding but I figure this can be added retrospectively as this component is required for most other components.

Something to note with the WYSIWYG editor in Sanity. You will see that there is a text colour dropdown rather than an option to highlight and select the text colour like you might expect from something like WordPress:

image

Due to how this would have to be setup there is currently no way to add a link and a colour to the same piece of text :/ 

It's an open issue with Sanity: https://github.com/sanity-io/sanity/issues/2956 but hasn't had any activity for a while.

To work around this I've opted for being able to colour the entire block rather than omit it entirely.

I've also implemented `dynamicLinks` from `remix-utils` to `osc-ecommerce` so we can conditionally load the stylesheet depending on when the component is used on a page.

It's a bit boilerplate-y as you have to export a `dynamicLinks` and a `handle` from each of the `routes` files (e.g. `$slug.tsx`). You have to do this as you need to be able to access the loaderData for each page so we can decide whether the component needs to be loaded. But I've added a function into the `Module.tsx` file so we don't have to edit every page whenever we add a new component. I've added some details around using it in the Additional Information section.

N/a

## 🚀 New behavior

Adds a `<Content />` component to `osc-ui`
Adds `dynamicLinks` to the `routes` files in `osc-ecommerce` so we can conditionally import stylesheets
Adds a content module to `osc-studio` so editors can update their content.

## 📝 Additional Information

To render the styles from your component in `osc-ecommerce` you will now need to load them into the `component/Module.tsx` and add a case to the getComponentStyles function.

```ts
import contentStyles from 'osc-ui/dist/content.css';

export const getComponentStyles = (data: SanityPage) => {
    const styles: LinkDescriptor[] = [];

    for (const module of data.modules) {
        switch (module._type) {
            case 'module.content':
                styles.push({ rel: 'stylesheet', href: contentStyles });
                break;
        }
    }

    return styles;
};
```

This will then pass the new styles through to the individual routes without you needing to touch them.


packages/osc-ui/src/index.tsx
```diff
@@ -1,7 +1,9 @@
 import './styles/main.scss';
-export { Header } from './components/Header/Header';
+
+export { Badge } from './components/Badge/Badge';
+export { Content } from './components/Content/Content';
 export { Footer } from './components/Footer/Footer';
+export { Header } from './components/Header/Header';
```
mattwendzina, line 6: I'm a fan of the alphabetising and think it's nice to do this wherever we can! = )

packages/osc-ecommerce/app/components/Module.tsx
```diff
@@ -1,8 +1,45 @@
...
+    contentModule,
+    module,
+    SanityPage,
+    trustpilotModule
+} from '~/types/sanity';
+
+// So we can dynamically add the styles of each component into remix we need to create an array of stylesheet objects.
+// We will then call this function in each of the `route` files where we have a dynamicLinks.
+// It's a bit boiler plate-y but is currently the only way to dynamically insert styles.
+export const getComponentStyles = (data: SanityPage) => {
```
mattwendzina, line 15: Is there a reason why this lives as an export in the `Module.tsx` component?

packages/osc-ecommerce/app/root.tsx
```diff
@@ -147,6 +148,7 @@ const Document = withEmotionCache(({ children }: DocumentProps, emotionCache) =>
                     <link rel="canonical" href={canonical} />
                 )}
                 {typeof document === 'undefined' && <Links />}
+                {typeof document === 'undefined' && <DynamicLinks />}
```
mattwendzina, line 151: In the docs it says to add <DynamicLinks/> before the <Links/> component, though I can see it says you can put it after so perhaps in our use case it's better for it to be second?

packages/osc-ecommerce/app/routes/$slug.tsx
```diff
@@ -46,6 +47,13 @@ export const loader: LoaderFunction = async ({ request, params }) => {
     });
 };
 
+// https://github.com/sergiodxa/remix-utils#dynamiclinks
+export const dynamicLinks: DynamicLinksFunction = ({ data }) => {
```
mattwendzina, line 51: Is there a reason why this function is always being exported in the routes?

mattwendzina:
Great work mate. I've added a few comments and had a bit of a play around with it in Sanity.

I have one question about the implementation of this Background Colour/Text Colour theme stuff. It seems to me like it's going to be a faff for marketing to have to always adjust one with the other. IE, if they change Background Colour to 'secondary' then they're going to need to change Text Colour to 'primary' every time. 

If I understand this theme stuff correctly I would argue that we should just give them the theme options and let this adjust both rather than them controlling both individually. I think this would be a better UX for them than always having to adjust both individually.

packages/osc-ecommerce/app/components/Module.tsx
stevenbister: Yes, because it's related to to the Module component (module component outputs the individual modules and getComponentStyles outputs individual module styles) I thought it made sense to group them together as you'll likely be editing both when you add a new module to the switch statement

packages/osc-ecommerce/app/root.tsx
stevenbister: Yeah I toyed around with doing it the other way but settled on this so the styles cascade in after the ecommerce styles. But thinking about it again we probably want ui styles to be higher up the cascade so we can overwrite them in ecommerce when necessary

stevenbister:
Yeah I think you're right, I discovered the colorScheme stuff after I'd done this and forgot about changing it in here. I'll see if anyone else has any thoughts but having just one dropdown would definitely be better!

packages/osc-ecommerce/app/root.tsx
stevenbister: Sorted :D

packages/osc-ecommerce/app/routes/$slug.tsx
stevenbister: Done

stevenbister:
I think I'll leave the colour scheme dropdown for now until we've ironed out how it'll work exactly 🤔 Perhaps it'll need splitting out of the button component and into it's own separate PR and I can tackle them all together

packages/osc-ecommerce/app/components/Module.tsx
mattwendzina: Okay cool. Thanks for the explanation!

hi-im-si (approved):

PR #476, opened 2022-11-01
Feature/breadcrumbs component

### LINKED ISSUE #449: [S32.1]: Breadcrumbs
### Link to Epic

### Description

n/a

### Linked Tasks

#450 

_No response_

### LINKED ISSUE #450: Build Breadcrumbs Component
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/449

### Description

n/a

mattwendzina:
* Closes #449 
* Closes #450 

## 📝 Description

Create Chakra Tag component, tests and stories. 

N/A

## 🚀 New behaviour

Adds Breadcrumb Component to the OSC-UI library

## 📝 Additional Information

The implementation relies on receiving a match `pathname` and `title` as part of a `matches` props. The Remix docs give an example of how to use breadcrumbs with the `useMatches` hook and `handle` function.
(See 'useMatches' section https://remix.run/docs/en/v1/api/remix)
I've attached a small screen recording demonstrating how this could be achieved. It's slightly different than they suggest - which is to pass the link in directly to the handle function. This approach just passes a title to the handle function which can then get picked up in the `root.tsx`. 
I believe the point of this approach is that it gives flexibility on whether or not to have breadcrumbs on any given route. 
To work it will require this `handle` function to be exported on any route where we would want breadcrumbs.

The screen recording will hopefully make more sense of this explanation..! 

https://user-images.githubusercontent.com/36510083/199284395-566c175d-03a9-4ba0-9f2a-cbb62a4d3575.mp4

### DIFF
--- a/packages/osc-ui/src/components/Breadcrumb/Breadcrumb.tsx
+++ b/packages/osc-ui/src/components/Breadcrumb/Breadcrumb.tsx
@@ -1,0 +1,40 @@
+import React from 'react';
+import type { FC } from 'react';
+import {
+    Breadcrumb as ChakraBreadcrumb,
+    BreadcrumbItem as ChakraBreadcrumbItem,
+    BreadcrumbLink as ChakraBreadcrumbLink
+} from '@chakra-ui/react';
+import { Link } from '@remix-run/react';
+
+interface Match {
+    pathname: string;
+    title: string;
+}
+
+export interface Props {
+    className?: string;
+    matches: Match[];
+    separator: any;
+}
+
+export const Breadcrumb: FC<Props> = (props: Props) => {
+    const { className, matches, separator } = props;
+
+    return (
+        <ChakraBreadcrumb className={className ? className : ''} separator={separator}>
+            {matches.map((match, idx) => {
+                return (
+                    <ChakraBreadcrumbItem
+                        key={idx}
+                        isCurrentPage={matches.length === idx + 1 ? true : false}
+                    >
+                        <ChakraBreadcrumbLink as={Link} to={match.pathname}>
+                            {match.title}
+                        </ChakraBreadcrumbLink>
+                    </ChakraBreadcrumbItem>
+                );
+            })}
+        </ChakraBreadcrumb>
+    );
+};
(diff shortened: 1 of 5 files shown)


packages/osc-ui/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,58 @@
...
+export interface Props {
+    className?: string;
+    matches: Match[];
+    separator: any;
+}
+
+export const Breadcrumb: FC<Props> = (props: Props) => {
+    const { className, matches, separator } = props;
+
+    const shouldRenderALinkOrNot = (matches, match, idx) => {
```
stevenbister, line 26: Perhaps `idx` should be `index` instead? Would be a bit clearer and as we're not abbreviating anything else it would match the style a bit better, I think.

packages/osc-ui/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import type { FC } from 'react';
+import './breadcrumb.css';
```
stevenbister, line 3: Looks like this is based off an older branch, when you rebase this will need updating to `.scss`

stevenbister (changes requested):
The styles haven't worked in Chromatic but it looks like this is based off an old branch. 

You'll need to update the base branch to main anyway so when you rebase hopefully the newer stuff will filter in

packages/osc-ui/src/components/Breadcrumb/Breadcrumb.tsx
mattwendzina: Updated

packages/osc-ui/src/components/Breadcrumb/Breadcrumb.tsx
mattwendzina: Done

packages/osc-ui/src/components/Breadcrumb/Breadcrumb.tsx
stevenbister: Noice

stevenbister (approved):

PR #542, opened 2022-11-10
Feature/alerts component

mattwendzina:
Closes # 

## 📝 Description

Create Alerts component

N/A

## 🚀 New behavior

Alerts component added

## 📝 Additional Information
This components uses an 'alert' prop to determine the relevant icon/colour. Maybe this will need updating so that it sets the backgroundColour based on our themes.


packages/osc-ui/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,74 @@
+import type { FC } from 'react';
+import React from 'react';
+import './alert.scss';
```
stevenbister, line 3: I'm getting an error when I run storybook because this file doesn't exist

packages/osc-ui/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,74 @@
...
+    description?: string;
+    icon?: boolean;
+    status: 'info' | 'warning' | 'success' | 'error' | 'loading';
+    title: string;
+}
+
+export const Alert: FC<Props> = (props: Props) => {
+    const {
+        classes,
+        customComponent,
```
stevenbister, line 32: If we want to include a custom component, rather than passing it as a prop should we make the component so that it accepts children? Might make it a bit more composable?

Alternatively is there a use case where we would be adding a component here that isn't a close button? And should this just be a boolean or something that allows us to toggle the usage?

packages/osc-ui/src/components/Alert/Alert.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { CloseButton } from '@chakra-ui/react';
+
+export default {
+    title: 'Alert',
+    component: Alert
+} as Meta;
+
+const AlertTemplate: Story = ({ items }) => {
+    return (
+        <div style={{ width: '20%' }}>
```
stevenbister, line 14: Is there a reason this is set to 20% width?

packages/osc-ui/src/index.tsx
```diff
@@ -8,3 +8,4 @@ export { Carousel } from './components/Carousel/Carousel';
 export { OSCModal as Modal } from './components/Modal/Modal';
 export { Tag } from './components/Tag/Tag';
 export { Breadcrumb } from './components/Breadcrumb/Breadcrumb';
+export { Alert } from './components/Alert/Alert';
```
stevenbister, line 11: Think this should be alphabetised, I know the others aren't but should probably start trying to do it :S

stevenbister (changes requested):
Hey man looks really good! Couple of comments for you, hopefully nothing major :)

PR #550, opened 2022-11-14
Feature/replace Chakra UI with Radix UI

stevenbister:
## 📝 Description

Replaces [Chakra UI](https://chakra-ui.com/) as our primitives library with [Radix UI Primitives](https://www.radix-ui.com/).

This should give us a few benefits:

- Remove our dependency on Emotion, which should in turn fix the "green flash" issue
- Remove Chakra theming issues in Storybook
- Remove need for Emotion workaround setup in Remix entry files 
- Remove Chakra dependency which was accounting for roughly 200kb JS, Radix on the same page return 160kb

### What this PR is _not_ doing
- Styling/theming
   Styles and theming will be handled as a separate PR as they require a more thought out approach. There is a `useTheme` hook that has been added but that is purely placeholder for the time being in an attempt to replicate the colour settings in components such as `<Badge />`. Please don't take this as the way we'll be handling this in future.
- Documentation in Storybook
  This was [discussed in Discord](https://discord.com/channels/915621085872029817/1014094093636354068/1040664537940099072) but I think this should be separate from this as there's already a lot happening here.
    - https://github.com/Open-Study-College/osc/issues/552

Styles and components are handled by Emotion and Chakra UI respectively

## 🚀 New behavior

Styles will be handled by our own stylesheets and component primitives will be built up from Radix UI.

I've set up some basic hooks to handle styling, my thinking is we will use these when we want to add utility classes (`useFontSize` for example) or modifiers (`useVariant`).

Unlike Chakra, styles will be completely handled by our stylesheets and the use of classnames. To make the setting of these a little easier I've added a utility function called `classNames` which takes an unlimited number class names, filters out any falsy values and sticks the remaining classes back together. We can then use this as the className rather than having to use a bunch of if elses and fallbacks. e.g.
```ts
const fontSizeClass = useFontSize(fontSize);
const vairant = useVariant('c-badge', variant);
const themeClass = useTheme(theme);
const classes = classNames('c-badge', vairant, fontSizeClass, themeClass);

return <span className={classes}>{badgeName}</span>;
```

Rather than importing components from Chakra we will now import them from Radix. Unlike Chakra, Radix separates primitives into their own packages so we will have to import them when we need them (not a bad thing) e.g.
```bash
npm install @radix-ui/react-accordion
```
And styles should be handled completely by class names & Scss.

This won't be relevant for all of our components but be aware that when extending a component Radix recommends using `forwardRef`. This will probably be most appropriate for things like buttons or inputs or any other interactive component that we might need to access outside: https://www.radix-ui.com/docs/primitives/overview/styling#extending-a-primitive

## 📝 Additional Information

I've added a couple of extra components which were needed by existing components
- `<Avatar />` (just the component and the story as this was needed by the Tag component)
- `<Icon />` (just a wrapper for the [accessible icon primitive](https://www.radix-ui.com/docs/primitives/utilities/accessible-icon))
- `<Switch />` (just the component and the story as this was within osc-ecommerce and academic hub)

Radix has a dialog (Modal in our lib) and an alert dialog which is separate and has slightly different features. We might need to add an alert dialog/modal to our roadmap.


packages/osc-ui/src/components/Modal/modal.scss
```diff
@@ -0,0 +1,75 @@
...
+    &--lg {
+        max-width: 1024px;
+    }
+
+    &--xl {
+        max-width: 1400px;
+    }
+
+    &--full {
+        max-width: 100%;
```
aaacafe786, line 51: osc-ui-modal--full might need abit of padding on the X button, i am on a 13" screen in chrome

stevenbister: Yeah I agree but this PR isn't handling styling so that'll be tackled when we start adding styles in later PRs :)

aaacafe786 (approved):

packages/osc-ui/src/utils/classNames.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Joins a list of classes together
+ *
+ * @param classes A comma separated list of classNames
+ * @returns A string of classNames
+ * @example classNames('c-badge', vairant, fontSizeClass, themeClass, className)
```
mattwendzina, line 6: Type on `variant`

packages/osc-ui/src/components/Tag/Tag.tsx
```diff
@@ -1,20 +1,17 @@
 import type { FC } from 'react';
 import React from 'react';
-import {
-    Tag as ChakraTag,
-    TagLabel as ChakraTagLabel,
-    TagLeftIcon as ChakraTagLeftIcon,
-    TagRightIcon as ChakraTagRightIcon
-} from '@chakra-ui/react';
-
-import { themeOptions } from '../../constants';
```
mattwendzina, line 10: Looks like the content of this constants file isn't being used now so perhaps can be deleted for now?

packages/osc-ecommerce/app/components/ThemeSwitcher/ThemeSwitcher.tsx
```diff
@@ -0,0 +1,30 @@
...
+            <Switch
+                checked={isChecked}
+                onCheckedChange={(isChecked) => onToggle(isChecked)}
+                id="color-mode-toggle"
+            />
+
+            <SunIcon />
+        </>
+    );
+};
```
mattwendzina, line 30: Styling on the switchers is slightly out of alignment. There's a few places where styling needs tweaking though so I'm guessing these are tasks for another time.

packages/osc-ui/src/components/Tag/Tag.tsx
stevenbister: These are the old lines, it should be gone :)

packages/osc-ecommerce/app/components/ThemeSwitcher/ThemeSwitcher.tsx
stevenbister: Yeah I'm saying styling is out of scope of this PR I've put some initial styles in but will defo need tightening up when we get to styling things :)

packages/osc-ui/src/components/Badge/Badge.tsx
```diff
@@ -1,26 +1,27 @@
...
+    fontSize?: Sizes;
+    theme?: Themes;
+    variant?: Variants;
 }
 
 export const Badge: FC<Props> = (props: Props) => {
-    const { color, variant, fontSize, className, badgeName } = props;
+    const { badgeName, className, fontSize, theme = 'primary', variant } = props;
+    const fontSizeClass = useFontSize(fontSize);
+    const vairant = useVariant('c-badge', variant);
```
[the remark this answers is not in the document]
stevenbister: Oops! Fixed

mattwendzina:
Great work mate! 

I've noticed that on the `osc-ecommerce` website that when you change the theme you get a flash in the url of `actions/changeTheme`. This is much more pronounced if you slow down the network speed. Not sure but is this something that needs to be triggered as a server call rather than something in local state? I'm guessing the latter given that this is how remix works right. Just don't think it's particularly great UX, especially given that the button doesn't change instantly - this only happens after the successful server call. 

I've added a few other small comments, but other than this LGTM!

packages/osc-ui/src/components/Tag/Tag.tsx
mattwendzina: Sorry, I meant the actual constants file itself.

packages/osc-academic-hub/app/components/ThemeSwitcher/ThemeSwitcher.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { Switch } from 'osc-ui';
+import type { FC } from 'react';
+
+export interface Props {
+    label: string;
+    onToggle: (isChecked: boolean) => void;
+    isChecked: boolean;
+}
+
+export const ThemeSwitcher: FC<Props> = (props) => {
```
mattwendzina, line 11: Not sure if it actually matters too much as VS code still seems to pick up the types, but we've typically been defining the prop types on the component arguments- ie - `...FC<Props> = (props: Props) => { ...`

packages/osc-ui/src/components/Tag/Tag.tsx
stevenbister: Ahhh gotcha, yeah good spot!

packages/osc-academic-hub/app/components/ThemeSwitcher/ThemeSwitcher.tsx
stevenbister: Yeah we should probably keep to the convention

stevenbister:
> I've added a few other small comments, but other than this LGTM!

Thanks man 😎 I'll get those changes sorted

> I've noticed that on the `osc-ecommerce` website that when you change the theme you get a flash in the url of `actions/changeTheme`. This is much more pronounced if you slow down the network speed. Not sure but is this something that needs to be triggered as a server call rather than something in local state? I'm guessing the latter given that this is how remix works right. Just don't think it's particularly great UX, especially given that the button doesn't change instantly - this only happens after the successful server call.

Yeah that bothers me as well 😅  The way it works at the moment is it set's a cookie, so the option is "saved", which is handled by the server a la the "Remix way" and because it's on a different route to the page it has to post the form there.
If we swap the `useSubmit` hook with `useFetcher` we can do a form submit without having to do a page reload which would probably be better, we could do some optimistic UI as well then so we don't have to wait for the response.

I'll open an issue 👍

packages/osc-ui/src/components/Tag/Tag.tsx
stevenbister: Sorted mate :)

packages/osc-academic-hub/app/components/ThemeSwitcher/ThemeSwitcher.tsx
stevenbister: Sorted :)

mattwendzina:
LGTM

PR #551, opened 2022-11-14
Feature/alerts component radix

### LINKED ISSUE #544: [S41.1]: Alerts
### Link to Epic

### Description

N/a

### Linked Tasks

#545 
#546 

_No response_

### LINKED ISSUE #545: Build Alerts Component
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/544

### Description

n/a

### LINKED ISSUE #546: Alerts Colours/theme  e.g. primary, secondary, danger, warning, info
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/544

### Description

n/A

mattwendzina:
Closes #544 
Closes #545 
Closes #546 

## 📝 Description

Create Alerts component in Radix instead of Chakra

N/A 

## 🚀 New behavior

Alerts component added

## 📝 Additional Information
This moves the Alerts component over from Chakra to Radix


mattwendzina:
I think we should be careful in how much/where this is used. 

Chakra add `role='alert'` to all of the alerts. However looking at the Aria docs for roles it says the following:

```
The alert role is used to communicate an important and usually time-sensitive message to the user... The alert role should only be used for information that requires the user's immediate attention, for example:

An invalid value was entered into a form field
The user's login session is about to expire
The connection to the server was lost so local changes will not be saved
```

I don't think the Chakra docs or some of alert status' fully reflect this. E.g. I don't think using `info` would suggest it is something that requires immediate user attention. 

Chakra has five - error, success, warning, info

I'd suggest that we support error, warning, and maybe info if we want to stick within the remit of what Aria define as 'alert'. 

Alternatively, we could still use this for success/info and just not pass the role of 'alert' in those cases.

mattwendzina:
@stevenbister - This PR should address all the comments in the last PR.

In terms of the comment on using `children` - I have done this but not sure if I'd addressed this in exactly the way you were suggesting.

stevenbister:
Nice man! There might be a couple of hooks/utilities in #550 that might be useful here 😄  Also noticed you're using chakra-icons, might be worth swapping over to [radix icons](https://icons.radix-ui.com/) instead?

> In terms of the comment on using `children` - I have done this but not sure if I'd addressed this in exactly the way you were suggesting.

Yeah sorta. I was thinking more along the lines of having an `<AlertHeader>` and maybe `<AlertDescription>` as well so perhaps you would use the component like this:
```ts
// osc-ecommerce or osc-academic-hub
import {CloseButton} from 'somewhere'
import {Alert, AlertHeader, AlertDescription} from 'osc-ui'

<Alert status="warning">
    <CloseButton />
    <AlertHeader>Heads up!</AlertHeader>
    <AlertDescription>Your account is about expire, upgrade now</AlertDescription>
</Alert>
```
That way depending on whether you put the `<CloseButton />` before or after the other children, that will dictate whether it appears on the left or right and you won't need the prop that sets the position, if you follow me?

stevenbister:
Nice man! There might be a couple of hooks/utilities in #550 that might be useful here 😄  Also noticed you're using chakra-icons, might be worth swapping over to [radix icons](https://icons.radix-ui.com/) instead?

> In terms of the comment on using `children` - I have done this but not sure if I'd addressed this in exactly the way you were suggesting.

Yeah sorta. I was thinking more along the lines of having an `<AlertHeader>` and maybe `<AlertDescription>` as well so perhaps you would use the component like this:
```ts
// osc-ecommerce or osc-academic-hub
import {CloseButton} from 'somewhere'
import {Alert, AlertHeader, AlertDescription} from 'osc-ui'

<Alert status="warning">
    <CloseButton />
    <AlertHeader>Heads up!</AlertHeader>
    <AlertDescription>Your account is about expire, upgrade now</AlertDescription>
</Alert>
```
That way depending on whether you put the `<CloseButton />` before or after the other children, that will dictate whether it appears on the left or right and you won't need the prop that sets the position, if you follow me?

stevenbister:
> Alternatively, we could still use this for success/info and just not pass the role of 'alert' in those cases.

Yeah I think this makes the most sense :)

mattwendzina:
This is updated with changes for Radix. I've done a story for 'Close Button', but just currently using the Radix button- Can hopefully update this to use the Button component once that's finished. I've also commented out this test and can add back in at the same time.

packages/osc-ui/src/components/Alert/Alert.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { screen, render } from '@testing-library/react';
+
+import { Alert, AlertDescription, AlertTitle } from './Alert';
+
+describe('Alert component', () => {
+    test('should render the correct text/classes for the title and description', async () => {
+        render(
+            <Alert status="info">
+                <AlertTitle title="OSC is going live on August 30th" />
+                <AlertDescription description="Make sure you are ready!" />
```
stevenbister, line 11: Is there a reason you've gone for props like this over using children for the AlertTitle and AlertDescription? Just thinking, if we wanted to format the text with a strong tag or em would that be possible like this?

stevenbister (changes requested):
Hey man! Looks really good :D A couple of (hopefully) fairly minor things

packages/osc-ui/src/components/Alert/alert.scss
```diff
@@ -0,0 +1,60 @@
...
+    &--success {
+        background-color: #c6f6d5;
+    }
+
+    &--warning {
+        background-color: #feebc8;
+    }
+
+    &__button--right {
+        position: absolute;
```
stevenbister, line 26: I'm not sure if the class name is exactly right if we are able to put the button on the left?

Also I'm not super keen on position absolute absolute-ing things but this one can be addressed when we come to styling

packages/osc-ui/src/components/Alert/Alert.stories.tsx
```diff
@@ -0,0 +1,97 @@
+import type { Meta, Story } from '@storybook/react';
+import React from 'react';
+
+import { Alert, AlertDescription, AlertTitle } from './Alert';
+import { Cross2Icon } from '@radix-ui/react-icons';
+
+export default {
+    title: 'osc-ui/Alert',
+    component: Alert
```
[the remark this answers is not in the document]
stevenbister: Oh also in the argTypes I would hide the items as they don't really influence the component other that what's rendered.
```ts
argTypes: {
    items: {
            table: {
                disable: true
            }
        },
}
```

packages/osc-ui/src/components/Alert/Alert.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+        const { container } = render(
+            <Alert displayIcon={false} status="success">
+                <AlertTitle title="OSC is going live on August 30th" />
+                <AlertDescription description="Make sure you are ready!" />
+            </Alert>
+        );
+        const icon = container.querySelector('svg');
+        expect(icon).toBeNull();
+    });
+    test('should set aria "alert role" if status is "error" or "warning"', () => {
```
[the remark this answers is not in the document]
mattwendzina: Yeah good point. Added.

packages/osc-ui/src/components/Alert/Alert.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+        render(
+            <Alert status="error">
+                <AlertTitle title="OSC is going live on August 30th" />
+            </Alert>
+        );
+        expect(screen.getByRole('alert')).toBeInTheDocument();
+    });
+    test('should not set aria "alert role" if status is not "error" or "warning"', () => {
```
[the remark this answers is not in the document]
mattwendzina: Yep, there is success too. Updated 👍

packages/osc-ui/src/components/Alert/Alert.stories.tsx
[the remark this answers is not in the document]
mattwendzina: Updated 👍

packages/osc-ui/src/components/Alert/Alert.spec.tsx
mattwendzina: Done

packages/osc-ui/src/components/Alert/alert.scss
mattwendzina: Yeah, I can see that there are different ways, but not had much experience with this which is why I did it this way. I had a look how Chakra did it using flex and align-self but didn't like how this worked if text was shorter than the container size. e.g.
Screenshot 2022-11-16 at 11 38 29

Though I'm sure there are other ways around it.

In terms of the className - I was thinking you would pass in `&__button--left {...}` if you wanted it on the left. I did a story for this but then wasn't convinced with how it looked and questioned whether you'd ever have a close button on the right anyway and so left it out. Do you have any suggestions?

stevenbister: Ahh ignore me! I thought there's a way to do it with flexbox but I can't get it to work without absolute position and using grid is probably too complicated 😅 

I wouldn't target it by classname so instead something along the lines of 
```ts
&__button {
    position: absolute;
}

&__description + &__button {
    right: 4px
}

&__ button+&__heading {
  left: 4px
}
```

but yeah probably not one for this pr

stevenbister (approved):
LGTM

packages/osc-ui/src/components/Alert/alert.scss
mattwendzina: Haha, okay. Cool, that's good to know 👌

packages/osc-ui/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const iconClasses = classNames(`c-alert__icon c-alert__icon--${status}`, className);
+
+    const alertIcon = displayIcon ? (
+        <span className={iconClasses}>
+            <Icon width={iconSize} height={iconSize} />
+        </span>
+    ) : null;
+
+    // Only set role to alert for more critical alerts
+    const setRoleToAlert = Boolean(status === 'error' || status === 'warning');
```
hi-im-si, line 53: If i am totally honest, i think alerting a user to provide information in any shape or form is probably worthy of the role attribute. I think the user will always need to be drawn to it.

if we wanted to suppress the announcement then we can add the `aria-live="polite"` as this essentially downgrades its urgency or even go as far as setting `aria-live="off"` to completely suppress it.

packages/osc-ui/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export const Alert: FC<Props> = (props: Props) => {
+    const { className, children, displayIcon = true, iconSize = 20, status } = props;
+
+    let Icon;
+
+    switch (status) {
+        case 'error':
+            Icon = WarningIcon;
```
hi-im-si, line 27: I know Chakra re-uses the warning icon but if would probably prefer to include an error icon.

hi-im-si (changes requested):
Looks pretty good so far 👍🏼

packages/osc-ui/src/components/Alert/Alert.tsx
mattwendzina: Would you class the exclamation triangle as an error icon?

packages/osc-ui/src/components/Alert/Alert.tsx
mattwendzina: I wasn't sure whether alerts will always be used to alert the user to provide information - Like a `success` alert is presumably just alerting the user that something has been successful, and `info` is alerting them with a piece of information that most likely doesn't require immediate attention or action.  But yeah, I could set them to `alert` as default and perhaps add something like a `notUrgent` boolean flag to toggle the urgency level down?

packages/osc-ui/src/components/Alert/Alert.tsx
hi-im-si: I think its probably the triangle more than anything. Few examples from a number of design systems - you can see some use the exclamation but within a circle or an 'X'. It was more an observation more than anything to raise this.

image
image
image
image

mattwendzina: Out of the box Radix only has an exclamation in a triangle, but has a crossed circle. Shall I update to one of those? When said this was an 'observation... to raise this' - did you mean raise a separate ticket, or that it's just an observation? Ie - Do you want the icon updated to resolve this?

hi-im-si: Happy to get this updated to the crossed circle 👍🏼

mattwendzina: Cool, done.

hi-im-si (approved):
🚀

PR #554, opened 2022-11-14
Feature/auth microservice

### LINKED ISSUE #438: [S3.1]: Auth Service
### Link to Epic

### Description

n/a

### Linked Tasks

#439 

_No response_

### LINKED ISSUE #439: Create Mapping for Auth Service
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/438

### Description

Mapping data structure between third party api and OSC api

### LINKED ISSUE #491: [S3.2]: Auth endpoint
### Link to Epic

### Description

n/a

### Linked Tasks

#499 
#500 
#501 
#502 
#503 

_No response_

### LINKED ISSUE #499: Research Auth inside Apollo servers
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/491

### Description

n/a

### LINKED ISSUE #500: Create JWT Library
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/491

### Description

n/a

### LINKED ISSUE #502: build login/logout end points
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/491

### Description

n/a

### LINKED ISSUE #503: testing logging in/out and accessing restricted sections
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/491

### Description

n/a

Tom-Pearce:
- Closes #438 
- Closes #439 
- Closes #491 
- Closes #499 
- Closes #500 
- Closes #502 
- Closes #503 

## 📝 Description

This is a working Authentication API, and gateway.

None

## 🚀 New behaviour

Able to create a new user.
Able to sign in, and retrieve an `accessToken` and `refreshToken`.
- `accessToken` lasts for a short amount of time and has extra data in it's payload.
- `refreshToken` lasts longer and is used to fetch a new accessToken, allowing users to stay logged in over longer periods of time.
- durations defined in `.env`

`createUser` and `login` mutations are form validated using `graphql-constraint-directive`.
- Currently this is throwing errors incorrectly (not proper GraphQL errors). I have reported the bug and they have acknowledged it and will be looking to fix.

Fetch a list of users.
Fetch details for a specific user.
- defaults to logged in user if `null`

GraphQL shield implemented to authorise requests to restricted endpoints (e.g. user/users).

Gateway has been changed to use `apollo-server-express` and now runs on an instance of express.
Gateway decodes `JWT` and passes payload as context to all queries to subgraphs.


packages/osc-api-auth/src/utils/auth.ts
```diff
@@ -0,0 +1,16 @@
+import type { userJWT } from '~/types/general';
+
+export const userCan = (user: userJWT, desiredPermission: string | []) => {
```
stevenbister, line 3: Not sure if this is intentional or not? Below, you've stated that the `desiredPermission` should be an array of strings if it's not a string so the type here should probably be `desiredPermission: string | string[]`. 

I'm actually not sure that `[]` will allow values to be assigned to the array but I might be wrong

Tom-Pearce: You're right, mistake on my part, with PHP running through my head 😁

packages/osc-api-auth/src/resolvers/resolver.ts
```diff
@@ -1,23 +1,72 @@
...
+                : await prisma.user.findMany({
+                      skip: start,
+                      take: limit,
+                      orderBy: {
+                          [orderBy]: orderDir
+                      }
+                  });
+        },
+        user: async (_: undefined, args: getUserArgs, context: AuthContext) => {
+            return args.id ?? context.user ?? false
```
mattwendzina, line 48: I thought the nested ternary's were controversial... not seen this before lol. 
Just to be clear - this is saying if `args.id` and `context.user` are `undefined` or `null` then return false - ie `null`, otherwise if one returns true then get user?

Tom-Pearce: Even better, I can just remove it, it's comepletely redundant

packages/osc-api-auth/src/resolvers/resolver.ts
```diff
@@ -1,23 +1,72 @@
...
+                  })
+                : await prisma.user.findMany({
+                      skip: start,
+                      take: limit,
+                      orderBy: {
+                          [orderBy]: orderDir
+                      }
+                  });
+        },
+        user: async (_: undefined, args: getUserArgs, context: AuthContext) => {
```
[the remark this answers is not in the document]
Tom-Pearce: Yeah I should probably go through and make it more consistent. Was a lot of playing around at various stages so got a bit lax when something started working.

packages/osc-api-auth/src/schemas/schema.ts
```diff
@@ -4,14 +4,113 @@ export const typeDefs = gql`
...
+        user(id: Int): User
+    }
+
+    input createUserInput {
+        firstName: String! @constraint(maxLength: 128)
+        lastName: String! @constraint(maxLength: 128)
+        email: String! @constraint(format: "email", maxLength: 255)
+        ### Minimum 9 characters, at least one uppercase letter, one lowercase letter and one number. #?!@$%^&*- characters allowed but not required
+        ### Note it seems to break if we use any / in the regex, for example /d for all digits [0-9] 
+        password: String! @constraint(minLength: 9, maxLength: 32, pattern: "^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])[a-zA-Z0-9#?!@$%^&*-]{9,}$")
```
chrisui, line 103: In terms of password best practices it would be recommended never to have a max limit on password length (given it's hashed there's no real technical reason to do this). And avoid requiring any specific patterns - instead simply have a set of incrementally complex regex's with a size which is matched against user input and simply informs them of their password strength (eg. those weak/ok/strong/epic bars you see sometimes).

Ie. It's better to only enforce a minimum length and strongly advise users to use a long, unique, password through a password manager - or at least long and memorable.

Eg. a password of only 12 lower-alpha-numeric characters would be many orders of magnitude stronger than 9 characters even with a wider set of characters enforced.

packages/osc-api-auth/src/schemas/schema.ts
```diff
@@ -4,14 +4,113 @@ export const typeDefs = gql`
...
+    }
+
+    input loginInput {
+        email: String! @constraint(format: email)
+        password: String! @constraint(minLength: 9, maxLength: 32)
```
chrisui, line 108: Avoid any constraints on login - real users either know their password or have to reset. Also by removing constraints from previous comment users are less likely to be forced to use a password they might not remember/have recalled from a manager. You're only giving easier access to/narrowing access for hackers and bots.

packages/osc-api-auth/src/utils/account.ts
```diff
@@ -0,0 +1,184 @@
...
+            }
+        });
+    }
+
+    // Compare input password to hashed password
+    const passwordMatch = await password.compare(input.password, user.password);
+
+    // If passwords dont match throw error
+    // We'll hide this behind the same text as no matching user for now to prevent identifying used emails
+    if (!passwordMatch) {
```
chrisui, line 63: if you wait for 1-3 seconds when passwords don't match you can greatly reduce the speed at which any bot would be able to brute force user's passwords without damaging user experience for the vast majority of users.

Ie. the bot has to wait many many times more for a bad response before proceeding

packages/osc-api-auth/src/types/environment.d.ts
```diff
@@ -0,0 +1,14 @@
+declare global {
+    namespace NodeJS {
+        interface ProcessEnv {
```
chrisui, line 3: A type-safe and runtime-validated way to access env variables could be to use `zod` or similar validation/parsing library.

```ts
const env = z.object({
  NODE_ENV: z.enum(['development', 'production']),
  SALT_ROUNDS: z.number()
}).parse(process.env)

env.SALT_ROUNDS // guaranteed to be set and parsed as a number here
```

hi-im-si (commented):
Some feedback as promised:


- `_: undefined` is odd, I believe you can just do `(_, a: string, b: string)` if the argument is unused
- All the cases of `foo: string | null` can just be `foo?: string` unless `null` is explicitly being passed in, which it probably isn't
- Most of these models can likely be `readonly`, so `readonly foo?: string` it'll just make it more robust
- There's a comment about destructing arguments, it's probably better off done in the body of the function, it's easier to read
- Maybe instead of typing the arguments type the function soo: `type Foo = (bar: string) => string` and then `const foo: Foo = bar => { ... }` it may be easier to maintain
 - Just for clarity returning ternaries can be hard to read sometimes, just return out the function early e.g
```ts
{
    if (pagination === "cursor") {
       return await prisma.user.findMany({....});
    }

    return await prisma.user.findMany({
        skip: start,
        take: limit,
        orderBy: {
            [orderBy]: orderDir
        }
    });
}
```
- Maybe add some tests that verify that the env vars as set, otherwise exit the service on start up, last thing you want is the service running with "undefined" as the secret and not knowing about it
- Maybe break down some methods into smaller methods and have some unit tests :slightly_smiling_face:

You'll want to do some unit tests that you'll be able to do in vitest and then some integration tests to test the flow just via the APIs, then use playwright to test the the entire flow end to end. Contract tests are also worth looking into, you can use Pact for this, but it looks like this just follows Oauth2 spec so it might not be as necessary!


Your integration tests would usually spin up the service and just test against itself - while mocking most of the external things (edited) 

Lots of unit tests > less integration tests > less contract tests > less e2e tests (happy paths only really)

packages/osc-api-auth/src/resolvers/resolver.ts
```diff
@@ -1,23 +1,70 @@
...
+            await account.get(args.id ?? context.user!.id);
```
mattwendzina, line 48: Do you still need to `return` this?

mattwendzina: Presumably, unlikely as it is, if both of those values were `undefined`/`null` it would just return nothing rather than causing any errors?

Tom-Pearce: Yeah, my bad, good spot

hi-im-si (approved):

PR #557, opened 2022-11-15
Feature/countdown clock

### LINKED ISSUE #532: [S37.1]: Countdown clock
5### Link to Epic

### Description

n/a

### Linked Tasks

#533 
#594 

_No response_

### LINKED ISSUE #533: Build Countdown component
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/532

### Description

n/a

mattwendzina:
Closes #532 
Closes #533 

## 📝 Description

Create a Countdown clock component

N/A

## 🚀 New behavior

Countdown clock component added 

## 📝 Additional Information

I've created a Countdown clock container that holds the timer logic and passes this through to the `CountdownClock`. Currently I have just put this alongside the CountdownClock in OSC-UI but I wonder whether it belongs in OSC-Ecommerce as we will be passing the data into it and it will hold the state that manages the time.

Currently the styles are also commented out so you will need to comment these back in to test. This is because it's failing on the linting at the moment and I'm not sure why.


packages/osc-ui/src/components/CountdownClock/CountdownClockContainer.tsx
```diff
@@ -0,0 +1,111 @@
...
+    const [timer, setTimer] = useState<Timer>({
+        active: false,
+        days: '0',
+        hours: '0',
+        minutes: '0',
+        seconds: '0'
+    });
+
+    const now = () => new Date().getTime();
+    const distance = () => endDate - now();
```
Tom-Pearce, line 49: Would `duration` or `remainingDuration` be a more appropriate name? Distance makes me think of meters/km/etc

packages/osc-ui/src/components/CountdownClock/CountdownClock.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import type { FC } from 'react';
+import React, { Fragment } from 'react';
+
+interface Titles {
+    days: any;
+    hours: any;
+    minutes: any;
+    seconds: any;
+}
```
Tom-Pearce, line 11: Curious as the idea behind the titles as customisable props? I can see from the tests, it's for "5d, 3h", but Just thinking what happens when we reach 1 for a given section if I chose to use the full word? "1 hours" "1 days" would look a bit unpolished, but not sure how you'd manage plural/singular if the titles are supplied as props since people could pass anything.

mattwendzina: Good point. Given that we've been briefed to keep as current design I'm tempted to drop this as an customisable prop and just set as 'd,h,m,s'.  I hadn't thought about the additional logic that would have to be brought in if one were to choose other options. Feels more it would make more sense for  someone to write the logic for the particular use case if and when it occurs.

packages/osc-ui/src/components/CountdownClock/CountdownClockContainer.tsx
mattwendzina: Yeah, fair shout. I think `remainingDuration` is more clear.

packages/osc-ui/src/components/CountdownClock/CountdownClock.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export interface Props {
+    classNames?: string;
+    icon: any;
+    name?: string;
+    timer: any;
+    titles: Titles;
+}
+export const CountdownClock: FC<Props> = ({ classNames, icon, name, timer, titles }: Props) => {
```
stevenbister, line 20: Doesn't look like the classNames prop is being called anywhere?

stevenbister (changes requested):
Looks great! 

I agree with Tom's comments, I also feel like the component names (`CountdownClock` and `CountdownClockContainer`) aren't that intunitive.

I would expect to be doing this
```ts
<CountdownClock ... />
```
over this
```ts
<CountdownClockContainer ... />
```
To me `CountdownClockContainer` seems like I would be able to add stuff to it. 
Perhaps swapping the names around so `CountdownClockContainer` becomes `CountdownClock` and the current `CountdownClock` becomes `CountdownClockInner` or something.

Hopefully that makes sense? Not sure if I've explained that very well 😬

packages/osc-ui/src/components/CountdownClock/CountdownClockContainer.tsx
mattwendzina: Updated

mattwendzina:
Fair point @stevenbister on the naming of the files. I got used to naming components that handled the data the containers when using Redux, but doesn't make much sense here. I've updated 👍

packages/osc-ui/src/components/CountdownClock/CountdownClock.stories.tsx
```diff
@@ -0,0 +1,59 @@
+import type { Meta, Story } from '@storybook/react';
+import React from 'react';
+import { CountdownClock } from './CountdownClock';
+import type { Props } from './CountdownClock';
+import { ClockIcon } from '@radix-ui/react-icons';
+
+export default {
+    title: 'osc-ui/Countdown Clock',
+    component: CountdownClock
```
stevenbister, line 9: Sorry man, not sure why I missed this yesterday! You mind adding the description and args stuff here?

packages/osc-ui/src/components/CountdownClock/CountdownClock.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+const template: Story<Props> = (args) => <CountdownClock {...args} />;
+
+export const Primary = template.bind({});
+export const WithIcon = template.bind({});
+export const WithTitle = template.bind({});
+export const WithMultipleTimers = template.bind({});
+
+Primary.args = {
+    endDate: new Date('Jan 21,2023, 19:20:00'),
+    titles: { days: 'd', hours: 'h', minutes: 'm', seconds: 's' }
```
stevenbister, line 21: Think you've removed this prop from the component now? Probs worth taking it of here as it's showing up in the story

stevenbister (changes requested):
Couple of little things but otherwise LGTM :D

packages/osc-ui/src/components/CountdownClock/CountdownClock.stories.tsx
mattwendzina: No probs, thanks for the reminder. Done 🙌

packages/osc-ui/src/components/CountdownClock/CountdownClock.stories.tsx
mattwendzina: Nice spot. Done.

stevenbister (approved):
LGTM 🎉

hi-im-si (changes requested):
Looks Great!
Only thing i can see missing is an expiration message or an onComplete. This would be simply some text to inform the user for example that a sale is over or even potentially display a link to navigate them to a special offers page.

mattwendzina:
> Looks Great! Only thing i can see missing is an expiration message or an onComplete. This would be simply some text to inform the user for example that a sale is over or even potentially display a link to navigate them to a special offers page.

Added in functionality for onComplete message/link

hi-im-si (approved):

PR #567, opened 2022-11-21
Feature/buttons component

### LINKED ISSUE #217: [S12.1]: Buttons
### Link to Epic

### Description

n/a

### Linked Tasks

- [x] #218 
- [x] #219 
- [x] #248 
#598

### LINKED ISSUE #218: Build Buttons component
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/217

### Description

Conditions for button:

- Must be able to define correct button type e.g. <a href=""> or `<button>` 
- Variant (Primary, Secondary ... etc ) - this usually correlates with theme, font size and button size
- Label
- Type ( File, Email, External Link, Internal Link, No Link, Copy to Clipboard, Add to Cart)
- Links

Add testing
E2E Integration

### LINKED ISSUE #219: Buttons - Colours/theme
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/217

### Description

n/a

### LINKED ISSUE #248: Button - make dynamic with Sanity
### Link task to story

Story: #229

### Description

Button must have the following conditions controlled in the CMS

- Must be able to define correct button type e.g. <a href=""> or `<button>` 
- Variant (Primary, Secondary ... etc ) - this usually correlates with theme, font size and button size
- Label
- Type ( File, Email, External Link, Internal Link, No Link, Copy to Clipboard, Add to Cart)
- Links

This component will need to be part of the Content component

stevenbister:
- Closes #217 
- Closes #218
- Closes #219 
- Closes #248 

## 📝 Description

Adds the button component and links it up with the content component so we can add buttons using Sanity.

N/a

### Button component
I've refactored the original implementation of this button so it's a bit simpler to use, and hopefully maintain. Rather than relying on an `action` prop we now set the type of button (button, anchor, RemixLink) with an `as` prop. And the copy text option is a component in and of itself which extends the default `<Button>`.
So regular buttons would be used like this

```tsx
<Button
    as="a"
    href="https://example.com"
    size="md"
    target="_blank"
    variant="solid"
>
  External link
</Button>
```
and the copy button would be like this
```tsx
<CopyButton
    size="md"
    textToCopy="Copied text"
    variant="solid"
>
  Copy to clipboard
</CopyButton>
```

I think splitting them up like this makes the intent a bit clearer and allows us to better handle typescript and setting conditional types.

### osc-ecommerce
Inside the `<Content />` component in `osc-ui` I have added a new prop called `buttons` which takes an array of button objects (this will come out of Sanity). There is now a map in there which loops over the buttons and displays the matching type.

I also discovered that the current implementation of the `getComponentStyles` function would only return top level modules, so something like this where the buttons module is part of the content module meant that the buttons stylesheet would only be loaded if we made it dependent on another module i.e.

```ts
for (const module of moduleTypes) {
        switch (module) {
            case 'module.content':
                styles.push({ rel: 'stylesheet', href: buttonStyles });
                styles.push({ rel: 'stylesheet', href: contentStyles });
                break;
        }
    }
```

To avoid this dependency I've added a function called `getTypes` in `models/sanity.server` which will recursively search through the `sanityPage` data and return an array of all of the modules. This is then used in the `getComponentStyles` like this to remove any duplicates we've returned.

```ts
const moduleTypes = [...new Set(getTypes(data))];
```

## 📝 Additional Information

I have added a `StrictUnion` type to the `osc-ui/types` file. This is going to be a very useful helper for setting conditional types and creating unions. It's used like this:
```ts
StrictUnion<DefaultButtonProps | AnchorProps | LinkProps>;
```
And essentially looks through that union and checks for props that match in each type/interface. If the types aren't shared then typescript will error if you try and used a type from `DefaultButtonProps` that isn't in `AnchorProps`. There is probably a much better explanation here: 
- https://stackoverflow.com/questions/52677576/typescript-discriminated-union-allows-invalid-state/52678379#52678379
- [Playground](https://www.typescriptlang.org/play?#code/FAFwngDgpgBAqgOwJYHsEGkpgM4B4AqAfDALwz4xQAeIUCAJtuTAPwwDWWKAZswFwwEUAG5QATgG5QkWAGUQYpAGMQiVAgASUADbQxBADTkAgtu3EyFarQZMAhgjCtmAMhgAFO2JBI723ABKUEooYvS4AKJUStoArvRQuGpomDgEpuZGnGA85IRGQqJihMQCheJS4NAw8ooqyQgEFjUKyqrIaFq64oZ5wNLVFGQtde3quADexmWxALYARuJGAEIzC+IAvjAAPjATAMJri2IwG4RSwCEI2CAwwn5I9AJDe9MwAIwrAgBMG1JXNxgfmwKAAag8nswyAcBABmP79AD0iJgETEYlClzQgKQCHu2kezxgpFeAk+MFWMG+RkOMHhUmRaIxYix11uwJQAEk8RCidDpuTafSgA)


packages/osc-ui/src/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,159 @@
...
+    }
+};
+
+export const Anchor = Template.bind({});
+Anchor.args = {
+    ...Primary.args,
+    children: 'External link',
+    action: 'anchor',
+    href: 'https://example.com',
+    target: '_blank'
```
mattwendzina, line 121: This doesn't actually appear to be opening the link in a new window. I can see that it has the `rel="noopener noreferrer"` but doesn't set the `target` - Looking at the code it appears that you aren't setting the `target`.

packages/osc-ui/src/components/Button/Button.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+    });
+
+    test('sets the rel to noopener and noreferrer when target is "_blank"', () => {
+        render(
+            <Button action="anchor" href="https://example.com" target="_blank">
+                Test Button
+            </Button>
+        );
+
+        expect(screen.getByRole('link')).toHaveAttribute('rel', 'noopener noreferrer');
```
mattwendzina, line 109: Maybe worth testing that `target="_blank"` is only present too?

packages/osc-ui/src/components/Button/Button.stories.tsx
stevenbister: Aha yeah I'd assumed that the `{...attr}` from `ButtonHTMLAttributes` would apply it but I think because I'd specifically set it as a prop it gets skipped over. Sorted now anyway 😄

packages/osc-ui/src/components/Button/Button.spec.tsx
stevenbister: Yeah good shout!

mattwendzina:
Nice work mate. LGTM!

hi-im-si (approved):
🚀

PR #568, opened 2022-11-21
Refactor/update vitest environment

### LINKED ISSUE #547: Update vitest environment to use jsdom by default rather than happy-dom
### Description

At the moment when writing tests, to use things like react-testing-library's `user-event` we need to include this at the top of the test file:
```ts
/**
 * @vitest-environment jsdom
 */
```
I would like to update the `vite.config` / `vitest.config` to use jsdom as the environment by default rather than `happy-dom`

Using `happy-dom` makes me sad. I always forget about adding the `jsdom` comment to my tests and therefore spend a chunk of time trying to work out what is wrong before I realise what I need to do.

### Proposed Solution or API

Will need to explore if there is more to it but I believe it's just a matter of updating the vite config files like this

```ts
{
    css: true,
    reporters: ['verbose'],
    globals: true,
    environment: 'jsdom',
    setupFiles: ['./__test__/setup-test-env.ts'],
    include: ['./src/**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx,css}']
}
```

### Alternatives

_No response_

### Additional Information

_No response_

stevenbister:
- Closes #547 

## 📝 Description

Replaces happy-dom in the Vite environment with jsdom

Vite currently uses happy-dom as the default testing environment. This means that to use things just as `userEvent` from react testing library we have to manually add 
```ts
/**
 * @vitest-environment jsdom
 */
```
at the top of the file to test in jsdom

## 🚀 New behavior

I have updated the vite/vitest config files to use jsdom by default so we don't have to remember to manually add the environment setter. I've also uninstalled happy-dom as we're no longer using it.

You no longer need to add the vitest environment at the top of your test files.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17687,6 +17687,8 @@
       "resolved": "https://registry.npmjs.org/@types/concat-stream/-/concat-stream-1.6.1.tgz",
       "integrity": "sha512-eHE4cQPoj6ngxBZMvVf6Hw7Mh4jMW4U9lpGmS5GBPB9RYxlFg+CHaVN7ErNY4W9XfLIEn20b4VDYaIrbq0q4uA==",
       "dev": true,
+      "optional": true,
+      "peer": true,
       "dependencies": {
         "@types/node": "*"
       }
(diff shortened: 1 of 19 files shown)


hi-im-si (approved):
LGTM!

PR #573, opened 2022-11-23
Feature/keen slider

aaacafe786:
Closes # 

## 📝 Description

> Add a brief description

## 🚀 New behavior


.github/actions/deployment/action.yml
```diff
@@ -44,6 +44,12 @@ runs:
       env:
         FLY_API_TOKEN: ${{ inputs.FLY_API_TOKEN }}
 
+    - name: Destroy app
+      if: ${{ env.PR_CLOSED != 'true' }}
```
stevenbister, line 48: This ideally should be in a different PR

packages/osc-ecommerce/app/root.tsx
```diff
@@ -157,6 +155,7 @@ export default function App() {
         checkConnectivity(online, offline).then((data) => console.log(data));
     }, []);
 
+    // add comment to deploy
```
stevenbister, line 158: What's this?

packages/osc-studio/utils/joinWithAnd.ts
```diff
@@ -0,0 +1,5 @@
+export const joinWithAnd = (list: string[]) => {
```
stevenbister, line 1: What is this? I know I added it to an unmerged PR but don't think it's being used in here? Can you remove if not

packages/osc-studio/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "target": "ES2021",
```
stevenbister, line 3: What is this? I know I added it to an unmerged PR but don't think it's being used in here? Can you remove if not

packages/osc-ui/src/components/Carousel/Carousel.tsx
```diff
@@ -1,274 +1,368 @@
...
-    const setAriaHidden = useCallback(() => {
-        if (!emblaApi) return;
-        const slides = emblaApi.slideNodes();
-        emblaApi.slidesInView(true).forEach((indexInView) => {
-            slides[indexInView].setAttribute('aria-hidden', 'false');
-        });
-        emblaApi.slidesNotInView(true).forEach((indexNotInView) => {
-            slides[indexNotInView].setAttribute('aria-hidden', 'true');
+    const ResizePlugin: KeenSliderPlugin = (slider) => {
+        const observer = new ResizeObserver(function () {
```
stevenbister, line 86: See my comments around `trueSlidesPerPage` and `updateHeight` we might be able to remove this resize observer

packages/osc-ui/src/components/Carousel/carousel.scss
```diff
@@ -1,101 +1,231 @@
...
+    justify-content: center;
+    font-size: 50px;
+    font-weight: 500;
+}
 
-    .embla__slide_caption {
-        padding: 0;
-        margin: 0;
-    }
+.keen-slide1 {
```
stevenbister, line 15: Presume these are just here for testing?

packages/osc-ui/src/components/Carousel/carousel.scss
```diff
@@ -1,101 +1,231 @@
...
+    border-radius: 50%;
+    cursor: pointer;
+}
 
-.embla__slide_inner {
-    height: var(--embla__height);
-    margin: var(--embla__slideGap);
-    padding: var(--embla__slideGap);
+.dot:focus {
+    outline: none;
```
stevenbister, line 210: Please don't remove the focus outlines from things as it's an accessibility issue. If you don't want to see it when you use your mouse use `focus-visible` instead

packages/osc-ui/src/components/Carousel/Carousel.tsx
```diff
@@ -1,274 +1,368 @@
...
-        }
-    }
+            return false;
+        })
+    );
+    const [loaded, setLoaded] = useState(false);
+    const [sliderRef, instanceRef] = useKeenSlider<HTMLDivElement>(
+        {
+            loop,
+            drag: indicatorActive || dotsActive,
```
stevenbister, line 201: This should be true by default, we would always expect the slider to be dragable
```suggestion
            drag: true,
```

packages/osc-ui/src/components/Carousel/Carousel.tsx
```diff
@@ -1,274 +1,368 @@
...
+            dragSpeed: 0,
```
stevenbister, line 202: Do we need this? The default is 1 so not sure why you've set it to 0 here?

stevenbister (changes requested):
Hi mate, left a few comments for you. I think this can be simplified a lot if we lean more into what Keen Slider gives us. I've not reviewed your tests, Sanity or storybook in this instance I tried to focus on just the component itself. Hopefully my comments make sense!

A couple of other general things:

There are a few things in here that should be separate PRs which is making reviewing a bit hard as it's hard to tell what's required for the carousel and what's needed for something else.

Also can you add more detailed comments throughout your code. It's really helpful to not just state what something is but to maybe also explain _why_ something is done in that way. For example why you're doing the adaptive height in one way whereas Keen give an example with a different implementation that doesn't need a resize observer.
There is also some quite complex logic in some if statements that would benefit from a comment above it explaining in english what is happening.

packages/osc-ui/src/components/Carousel/Carousel.tsx
```diff
@@ -1,274 +1,368 @@
...
+    const [trueSlidesPerPage, setTrueSlidesPerPage] = useState(0);
 
-    const onSelect = useCallback(() => {
-        if (!emblaApi) return;
-        setSelectedIndex(emblaApi.selectedScrollSnap());
-    }, [emblaApi, setSelectedIndex]);
+    function updateHeight(slider: any) {
+        const largestSlide = Math.max(
+            ...Array.from(slider.slides).map(
+                (slide) => (slide as any).children[0].getBoundingClientRect().height
```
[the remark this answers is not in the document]
aaacafe786: please try this out before dropping it, i could not get the intended behaviour to work with the default

packages/osc-ui/src/components/Carousel/Carousel.tsx
```diff
@@ -1,274 +1,368 @@
...
+                {loaded &&
+                    instanceRef.current &&
+                    indicatorActive &&
+                    indicatorHeader &&
+                    dots.length > 1 && <div className={`indicator header`}>{arrows()}</div>}
+            </div>
+            {/* slider */}
+            <div className="navigation-wrapper">
+                <div ref={sliderRef} className="keen-slider">
+                    {React.Children?.map(children, (q, index) => {
```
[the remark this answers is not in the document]
aaacafe786: hey man, i think it needs an empty div to function, please test this out before dropping it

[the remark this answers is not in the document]
aaacafe786: not lazy loading slides means that you can end up loading all the slides in the carousel regardless of how large it is, the following ternary means: if the index is between the slidesInView (array) +/- trueSlidesPerPage (sorry, i know you dont like this, maybe there is an alternative), render the content of the slide in the dom otherwise render and empty div

packages/osc-ui/src/components/Carousel/Carousel.tsx
```diff
@@ -1,274 +1,368 @@
...
     } = props;
 
-    const delayInt = parseInt(delay, 10);
-
-    const trueSlidesPerPage = () => {
-        if (typeof window === 'undefined') return 1;
-        if (window.matchMedia('(min-width: 768px)').matches) return slidesPerPage;
-        if (window.matchMedia('(min-width: 640px)').matches) return 2;
+    const calcTrueSlidesPerPage = useCallback(() => {
+        if (window.matchMedia('(max-width: 640px)').matches) return 0;
```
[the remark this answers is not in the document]
aaacafe786: they also expose a prop called portion on each slide, it allows you to calculate all the slides in view, might be an easy replacement
```    
const slidesInView = (slider) => {
        const loadedSlides = slider.track.details.slides.map((slide, idx) => {
            if (slide.portion > 0) {
                return true;
            } else {
                return false;
            }
        });
        return loadedSlides;
    };
```

packages/osc-ui/src/components/Carousel/Carousel.tsx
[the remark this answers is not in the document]
aaacafe786: the reason it's not just slides in view, is because you can move up-to the number of slides per page in one swipe on desktop and these images need to be preloaded

packages/osc-ui/src/components/Carousel/Carousel.tsx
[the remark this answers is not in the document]
aaacafe786: you will still need the observer incase the page is resized, keep this as a minimum: https://codesandbox.io/s/github/rcbyr/keen-slider-sandboxes/tree/v6/misc/resize/react-typescript?file=/src/App.tsx

packages/osc-ui/src/components/Carousel/Carousel.tsx
aaacafe786: nah, this can go

packages/osc-ui/src/components/Carousel/carousel.scss
aaacafe786: yeah, came with the default keen slider setup

packages/osc-ui/src/components/Carousel/Carousel.tsx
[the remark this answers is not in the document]
stevenbister: We shouldn't be lazy loading the entire slide or it's content. The images in the slide yes, but lazy loading the slide itself or the content inside it will remove it from the DOM. This means the content won't exist for things like screen readers or search engines, which require the content to exist for them to function/rank our site.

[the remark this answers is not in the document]
aaacafe786: you would need to extract all the images from the child of carousel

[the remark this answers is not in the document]
aaacafe786: plus, there is an aria-hidden attribute which states that the slide is not in view, they can still click the button right

packages/osc-ui/src/components/Carousel/Carousel.tsx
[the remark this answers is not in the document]
stevenbister: But we're not resizing the slider in that way? If you use the `maxId` property then it automatically updates with the breakpoints, which should be handling the page resizing

packages/osc-ui/src/components/Carousel/Carousel.tsx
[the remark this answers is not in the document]
stevenbister: You _should_ be able to use the native image lazy loading so you wouldn't need to do that, and again the way Keen works is it removes it from the DOM which isn't really ideal.

Yeah they should still be able to click the button but search engines can't do that so it needs to exist, whether the aria attribute exists or not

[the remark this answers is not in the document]
aaacafe786: im pretty curious how exactly does a larger amount of slides improve the SEO score, i looked about and couldn't find anything very convincing for either side, lighthouse/webpagetest does not complain about it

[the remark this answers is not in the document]
stevenbister: It's got nothing to do with lighthouse -- it's about how Google works. Crawlers need to be able to parse the content and follow links, if the content isn't there they can't crawl the site fully and therefore can't rank us correctly. It's the same reason you need to prerender your site if you're using an SPA.

packages/osc-ui/src/components/Carousel/Carousel.tsx
```diff
@@ -1,274 +1,368 @@
...
+            if (slide.portion > 0) {
+                return true;
+            } else {
+                return slidesInView[idx];
             }
-            if (slideGap) {
-                r.style.setProperty('--embla__slideGap', slideGap + 'px');
+        });
+        console.log('this is the old slides in view', slidesInView);
+        console.log('this is the new slides in view', loadedSlides);
```
mattwendzina, line 163: Are these console logs needed now?

packages/osc-ui/src/components/Carousel/Carousel.tsx
```diff
@@ -1,274 +1,368 @@
...
+            if (slide.portion > 0) {
+                return true;
+            } else {
+                return false;
             }
-            if (axis) {
-                r.style.setProperty('--embla__axis', axis === 'x' ? 'row' : 'column');
+        });
+        console.log('this is the old slides in view', slidesInView);
+        console.log('this is the new slides in view', loadedSlides);
```
mattwendzina, line 176: Same as above comment on console logs

mattwendzina:
Having had a look at the carousel in Storybooks it looks like the arrows could do with some conditional styling to disable/grey out when it's at the end of the carousel deck. Or when it gets to the end it goes back to the start on the next click.
Perhaps the styling is out of the remit for getting this in now though so ignore if that's the case..!

packages/osc-ui/src/components/Carousel/Carousel.tsx
```diff
@@ -1,274 +1,368 @@
...
+    const AdaptiveHeight: KeenSliderPlugin = (slider) => {
+        // function updateHeight() {
+        //     const largestSlide = Math.max(
+        //         ...Array.from(slider.slides).map(
+        //             (slide) => slide.children[0].getBoundingClientRect().height
+        //         )
+        //     );
+        //     slider.container.style.height =
+        //         largestSlide > 0 ? largestSlide.toString() + 'px' : undefined;
+        // }
```
mattwendzina, line 111: Is there a reason this code is commented out or can it be deleted?

mattwendzina:
Bit confused as it looks like the `Card` component is in this PR as well as the Carousel. Are they going in together under the same PR?

PR #574, opened 2022-11-24
Refactor/set trailing commas to `es5`

### LINKED ISSUE #561: ENHANCEMENT: set `"trailingCommas": "es5"` in prettierrc.json
### Description

Change `tralingCommas` option to `es5` instead of `none` in `prettierrc.json`.

A habit I got into when coding in PHP. Trailing commas in objects/arrays allow for quick reordering, and addition of extra indexes, without risking forgetting to add the comma. It's arguably also cleaner and a bit easier on the eye

Trying to move `hello` up would start throwing errors, while an easy fix, a trailing comma means it would never have an issue since it would already have a trailing comma. So you would never miss the comma.
```ts
let arr = [
    "foo": "bar",
    "user": await prisma.user.findUnique({where: id: 1}),
    "hello": "world"
];
```

```ts
let arr = [
    "foo": "bar",
    "user": await prisma.user.findUnique({where: id: 1}),
    "hello": "world",
];
```
Let's be honest, we've all been there scouring a codebase to find out we missed a comma.

Only caveat would be that older browsers don't like this in JS and will throw an error, but then again "older browsers" generally refers to IE which is arguably dead. Not sure what the team stance for legacy browser support is, particularly with such a modern tech stack - I wonder how well it runs on IE <= 9

Additionally, it makes git commits cleaner, as when adding an item to the end of an array you no longer get the git diff highlighting the previous line (where you had to add the comma, since it's no longer the last item).

### Proposed Solution or API

In the root `prettierrc.json` change `"trailingCommas"` from `"none"` to `"es5"`

Before:
```json
"trailingCommas": "none"
```
After:
```json
"trailingCommas": "es5"
```

### Alternatives

_No response_

### Additional Information

PHP specific SO question answer:
https://stackoverflow.com/a/2829598/5350538

SO answer on above question quoting ZendFramework docs on the subject
https://stackoverflow.com/a/4850885/5350538

>When using this latter declaration, we encourage using a trailing comma for the last item in the array; this minimizes the impact of adding new items on successive lines, and helps to ensure no parse errors occur due to a missing comma.

Tom-Pearce:
- Closes #561 

## 📝 Description
Sets `trailingCommas` to `es5` in `prettierrc.json`

Trailing commas are removed. Current setting is `none`

## 🚀 New behaviour
Trailing commas are allowed in accepted areas, and prettier will add them automatically where appropriate. Typically only applies to multi line arrays/objects/function calls. Inline arrays/objects/function calls will not have a trailing comma.

This will also make Git diff logs cleaner as they will no longer show an edited line where a commas was added to allow an extra argument, since the commas will already be there.

Affected:
- Array literals
- Object literals
- Parameter definitions
- Function calls
- Named imports
- Named exports
- Array and object destructuring

## 📝 Additional Information
[MDN docs on trailing commas](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas)

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -18,6 +18,6 @@
     }
   ],
   "semi": true,
-  "trailingComma": "none",
+  "trailingComma": "es5",
   "singleQuote": true
 }


stevenbister (approved):
YES!

PR #608, opened 2022-12-05
Feature/image gallery component

### LINKED ISSUE #229: [S15.1]: Image Gallery
### Link to Epic

### Description

n/a

### Linked Tasks

- [ ] #230 
- [ ] #231

### LINKED ISSUE #231: Build Image Gallery Component
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/229

### Description

- Grid of Images (can support multiple layouts e.g. 2x2, 3x3, 4x4) 
- Each image can be clickable to display a modal/lightbox
- Support for captions

ref: https://www.youtube.com/watch?v=XJWdLbw3QjY

Build in storybook with:
- Testing (Vitest, Chromatic)

mattwendzina:
Closes #229 
Closes #231 

## 📝 Description

Add Image Gallery Component

N/A

## 🚀 New behavior

Enables use of an Image Gallery with a Lightbox

### The LightGallery plugin
- This uses LightGallery (www.lightgalleryjs.com). It requires a license to be used in production which costs $99 dollars for small teams.
- It has a number of plugins that can be attached and used in conjunction with different modes. There are also numerous lifecycle methods that can be tapped into if required.
- I have included the Thumbnail and Zoom plugins as default but these can be removed if necessary. The plugin uses tree shaking so each plugin is imported individually.

### CSS
- I've put all the css into an scss file. This includes the Core, Thumbnail and Zoom CSS.
- I've also added some additional styling (mainly for the different grids) at the top of the file.
- Not sure that this is the right/best approach to the styling, at minimum I guess the breakpoints will need swapping out for variables? Will need a bit of support when it comes to global things like the variables, correct way of setting up breakpoints etc...
- I haven't removed the CSS styles from the node_modules but I'm guessing this is something that can be done if we're just importing the styles locally?


packages/osc-ui/src/components/ImageGallery/ImageGallery.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+        // Expect this to be double as it renders one image for the initial view and one for the carousel
+        expect(screen.getAllByRole('img').length).toBe(8);
+        const images = container.getElementsByClassName('c-image-gallery__img');
+        for (const img of Array.from(images)) {
+            expect(img).toHaveClass('c-image-gallery__img--2x2-grid');
+        }
+    });
+
+    test('should open the lightgallery when the image is clicked', async () => {
+        const user = userEvent.setup();
```
stevenbister, line 34: Eslint seems unhappy about this? I think it's because the `render` function is sitting in the `setup()`. Perhaps we need something like this?
```suggestion
        // eslint-disable-next-line testing-library/render-result-naming-convention -- render is in setup function
        const user = userEvent.setup();
```

packages/osc-ui/src/components/ImageGallery/ImageGallery.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+        await setup({ images: singleImageData });
```
stevenbister, line 36: Don't think you need this await here

packages/osc-ui/src/components/ImageGallery/imageGallery.scss
```diff
@@ -0,0 +1,995 @@
+/* stylelint-disable */
+.c-image-gallery {
```
stevenbister, line 2: I would probably swap our styles around with the core styles so we are lower in the cascade, will make overwriting things a little easier

stevenbister (changes requested):
Looks good man, a couple of little suggestions but I guess there's not much else to do on this one.

Re the styles: 

Not using css grid? 😮 😉 
No need to change it now but when the work me and Si have done is merged there will be a grid system and media query variables in there so we'll be able to swap them out or replace the grid entirely.

I think removing the stylesheet from the node_modules is a good shout though if we're including the core styles in here

packages/osc-ui/src/components/ImageGallery/ImageGallery.spec.tsx
mattwendzina: Done

packages/osc-ui/src/components/ImageGallery/imageGallery.scss
mattwendzina: Good shout. Updated 👍

stevenbister:
> Haha. I have used grid a lot but not tonnes so not totally confident on the right approach with it. Would be good to learn/see it in action though..!

Haha yeah it is a bit funky until you get used to it!

> I wasn't really thinking the node_modules thing through. Don't think there's really an easy way to remove these as they just get installed with the package. Had a chat with Al and we couldn't see any simple solutions to this. 🤷

Ahh yeah fair enough, I thought it came as a separate package. I think as long as we're doing what you've done with the styles (taking ownership of them) and not importing it then it'll be fine

packages/osc-ui/src/components/ImageGallery/ImageGallery.spec.tsx
```diff
@@ -0,0 +1,47 @@
+/**
+ * @vitest-environment jsdom
+ */
```
stevenbister, line 3: Sorry man, just noticed this. You shouldn't need this now 👍 A fix went in a version or so ago 🎉 
https://github.com/Open-Study-College/osc/releases/tag/v0.15.1

stevenbister (changes requested):
Sorry mate I noticed one more thing, other than that I'm happy 👍

packages/osc-ui/src/components/ImageGallery/ImageGallery.spec.tsx
mattwendzina: Cool. Done 👍

stevenbister (approved):
LGTM 🎉

PR #635, opened 2022-12-09
WIP: feat: Forms component 

mattwendzina:
## 📝 Description

Initial approach is to create a form component (which can be standard or remix Form), and pass in the Input props (a separate component). 

This approach is advocated by Kent Dodds (https://epicreact.dev/improve-the-performance-of-your-react-forms/) as a way of improving form performance and ensuring only the input being typed into gets rendered rather than all. There are some caveats we'd probably have to deal with when one input element needs to know the value of another (e.g. for confirming password creation), but there are some reasonable approaches laid out for this. 

In terms of the Input component itself. I've got a switch case to switch between different types of inputs which seems like a reasonable approach to me, but if anyone thinks differently then let me know.

I think there are quite a few Radix components we are still going to want to utilise - e.g. Slider, Label, Toggle. I've just added these to the Input switch cases for now, but thinking maybe these would be their own inputs which get passed into the Form component to make it a bit more composable? Anyone have thoughts here?

The datepicker - I guess we'll want to pull in a component library for this?. I've not done any research on this yet tbh. I can see that there is a default HTML date picker, but not sure whether this will do and how easy it is to style. If anyone has any thoughts on this then let me know.

In terms of the submitting data logic. I'm going with the assumption that this will sit in e-commerce and that props will be passed to the form component to handle this. I've got handlesubmit stuff in there atm, but just to be able to test it atm.

## 🚀 New behavior


stevenbister:
Some initial thoughts, gonna give it a proper look over on Monday but this was open and I can't help myself 😂 

- Re the Input component, I would definitely separate that out into separate components. I would probably expect to see a component for each type of input. Presumably we would also have a `fieldset` and [error components](https://chakra-ui.com/docs/components/form-control/usage#error-message) a la Chakra?

- Re datepicker; I found this library, [React Aria](https://react-spectrum.adobe.com/react-aria/getting-started.html) (literally the day after the Radix work finished and I like it more lol). It has a pretty good looking [datepicker](https://react-spectrum.adobe.com/react-aria/useDatePicker.html) and a [date-range picker](https://react-spectrum.adobe.com/react-aria/useDateRangePicker.html) we can lift out and it should play fairly nicely with Radix.

- Form logic; yes this should sit in the remix apps so we can make use of their form handling and hooks such as [useFetcher](https://remix.run/docs/en/v1/api/remix#usefetcher)

stevenbister:
A couple more thoughts on this:

- Might be worth wrapping the components in a `forwardRef`, there may be times where we'd want to reference the input externally
- Would also be good to type them using the `HTMLInputElement` (or whichever element it is) type so we can pass the native attributes through, give me a shout if you don't know how / want showing how to do this :) Would make adding attributes a bit more flexible

mattwendzina:
> A couple more thoughts on this:
> 
> * Might be worth wrapping the components in a `forwardRef`, there may be times where we'd want to reference the input externally
> * Would also be good to type them using the `HTMLInputElement` (or whichever element it is) type so we can pass the native attributes through, give me a shout if you don't know how / want showing how to do this :) Would make adding attributes a bit more flexible

Thanks, yeah I'd thought about the forward ref but not got around to that yet. Will add it to the todo list though. And yeah, want to try and get the TS types in too. Gonna have a go, and I think your buttons is a good example. Will give you a shout if I'm having probs = )

stevenbister:
> In terms of the datepicker, I saw a comment on a Radix thread (thought I'd check when/if they'd bring one out 😄 ... apparently one day...), but someone left a comment 3 days ago saying lots of problems with React 18/SSR with the React-Aria datepicker sadly! - [radix-ui/primitives#969](https://github.com/radix-ui/primitives/discussions/969)

Ahh that's a shame :/ I'll keep my open for any other libraries we could use, I think I remember the default html datepicker having issues, but it's been a little while since I've looked into it.

aaacafe786:
hey all, keen slider comes with a built in datetimepicker, alternatively we can use the one from the request form

PR #639, opened 2022-12-13
Feature/css reset

hi-im-si:
## 📝 Description

- Adds a Modern CSS Reset
- Moves CSS Reset and Normalize to generic folder as specified in ITCSS methodology

No CSS base rules applied

## 🚀 New behavior

Adds in a simple CSS reset to normalize the browsers base interpretation

## 📝 Additional Information
N/A


packages/osc-ui/src/styles/generic/_reset.scss
```diff
@@ -0,0 +1,129 @@
...
+a,
+button {
+    cursor: revert;
+}
+
+/* Remove list styles (bullets/numbers) */
+ol,
+ul,
+menu {
+    list-style: none;
```
stevenbister, line 98: Do we want to do this for every list? I imagine there would be times where we want the bullets. I.e. content out of a wysiwyg, which probably wouldn't have a class applied.

Perhaps doing something like `ul[class]`? So we're only removing the list styles when we're specifically doing something with it?

stevenbister (commented):
Fancy! Using the `:where()` function :P Looks good, just one question below

packages/osc-ui/src/styles/generic/_reset.scss
hi-im-si: As we discussed, the reset is purely designed to bring all elements back to a default and therefore utilise overrides using classes for building on top of these

PR #647, opened 2022-12-29
Feature/select component

### LINKED ISSUE #613: Forms label
### Link task to story

### Description

n/a

mattwendzina:
Closes #613 
Closes - Forms Select component (Ticket needs creating...)

## 📝 Description

Adds Radix Select component and Label component. I've added some basic error handling in but this will need adapting as we work out our error handling approach. This approach assumes receiving a prop from the form - `wasSubmitted`, which handles error checking upon submission rather than during the form being completed. See video below:

https://user-images.githubusercontent.com/36510083/210082040-9a04d8bb-8bf0-4ead-8428-8dccad20af9f.mov

There's also an update to the modifier function to allow use of multiple modifiers and an update to the button.scss with a z-index of `-1` as this was sitting on top of the Select dropdown.

Also, some of the colours like the greys and errors will need updating.

N/A

## 🚀 New behavior

Adds Select component


packages/osc-ui/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,117 @@
...
+            <div className={displayError ? `${selectClasses} c-select--error` : selectClasses}>
+                {setDescription(description)}
+                <SelectPrimitive.Root
+                    {...props}
+                    disabled={attributes?.disabled}
+                    name={selectName}
+                    onValueChange={(value) => setValue(value)}
+                    required={attributes?.required}
+                >
+                    <SelectPrimitive.Trigger
```
stevenbister, line 68: Unfortunately we're getting an a11y error in Storybook: `Element does not have inner text that is visible to screen readers`. Having a bit of a look under the hood this custom select is built on a `button` so the label isn't labelling it as you would expect.

It looks like we need to add an `aria-label` or perhaps `aria-lablledby` to this trigger element.

Alternatively it looks like Radix recommends using their `Label` component: https://www.radix-ui.com/docs/primitives/components/select#labelling 

Is there a reason we're not currently using that or is it planned for a separate PR?

packages/osc-ui/src/components/Select/select.scss
```diff
@@ -0,0 +1,166 @@
...
+            box-shadow: 0 0 0 2px var(--color-primary);
+        }
+
+        &[data-placeholder] {
+            color: var(--color-secondary);
+        }
+
+        &[data-disabled] {
+            background-color: #fafafa;
+            color: hsl(255deg 3.7% 78.8%);
```
stevenbister, line 36: Assuming these two values are to be updated to the tokens once we add the other greys in?

packages/osc-ui/src/components/Select/select.scss
```diff
@@ -0,0 +1,166 @@
...
+                width: 20px;
+                height: 20px;
+            }
+        }
+    }
+
+    &__content {
+        overflow: hidden;
+        background-color: var(--color-tertiary);
+        box-shadow: 0 10px 38px -10px rgb(22 23 24 / 35%), 0 10px 20px -15px rgb(22 23 24 / 20%);
```
stevenbister, line 51: Can we update these rgb values to a token at all?

packages/osc-ui/src/components/Select/select.scss
```diff
@@ -0,0 +1,166 @@
...
+            outline: none;
+        }
+
+        .c-select__item-indicator {
+            position: absolute;
+            left: 0;
+            display: inline-flex;
+            align-items: center;
+            justify-content: center;
+            width: 25px;
```
stevenbister, line 88: I would use an `em` value here so the icon scales with the font size

packages/osc-ui/src/components/Select/select.scss
```diff
@@ -0,0 +1,166 @@
...
+
+    &--bold {
+        .c-label {
+            font-weight: $weight-bold;
+        }
+    }
+
+    &--secondary {
+        .c-select__trigger {
+            background-color: var(--color-quinary);
```
stevenbister, line 106: Looks like these colour values get reused a few times here. Could we abstract them into a variable like I suggested with the label maybe?

stevenbister (changes requested):
Hey man looking good! Left a few comments for you, appreciate these are fiddly components.

As you've mentioned the colours for greys/errors need adding to the tokens so I've not added comments around them. Just something for you to maybe note; we've decided to use `hsl` as the colour values rather than `hex` or `rgb` so if you can try using that instead that would be great.

Only other thing is you've not committed the package.json with the radix select icon installed :P

packages/osc-ui/src/components/Button/button.scss
```diff
@@ -5,6 +5,7 @@
     --button-foreground-color: var(--colour-white);
 
     position: relative;
+    z-index: -1;
```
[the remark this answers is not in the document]
stevenbister: Or better yet I think there is a file in `styles/settings/_z-index` that you might be able to use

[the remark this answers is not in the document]
mattwendzina: Updated

packages/osc-ui/src/components/Label/label.scss
```diff
@@ -0,0 +1,18 @@
...
+    font-weight: $weight-reg;
+    cursor: pointer;
+
+    &--bold {
+        font-weight: $weight-bold;
+    }
+}
+
+.c-label__required {
+    padding-left: 1px;
```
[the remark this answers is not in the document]
mattwendzina: Done

packages/osc-ui/src/components/Label/label.scss
```diff
@@ -0,0 +1,18 @@
+/* stylelint-disable selector-class-pattern */
+@use "../../styles/settings" as *;
+/* stylelint-disable plugin/selector-bem-pattern */
+
+.c-label {
+    font-size: var(--font-scale-centi);
```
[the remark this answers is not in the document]
mattwendzina: Done

packages/osc-ui/src/components/Label/label.scss
```diff
@@ -0,0 +1,18 @@
...
+    font-weight: $weight-reg;
+    cursor: pointer;
+
+    &--bold {
+        font-weight: $weight-bold;
+    }
+}
+
+.c-label__required {
```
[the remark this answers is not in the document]
mattwendzina: Done

packages/osc-ui/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,317 @@
...
+import React, { useEffect, useRef } from 'react';
+
+import { ListBulletIcon } from '@radix-ui/react-icons';
+import { Select, SelectItem } from './Select';
+
+export default {
+    title: 'osc-ui/Select',
+    component: Select,
+    subcomponents: {},
+    argTypes: {},
```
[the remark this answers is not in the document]
mattwendzina: Yeah my bad, meant to add this and forgot! Added now. Added the JS Docs too, this is a nice touch for the descriptions 👌

packages/osc-ui/src/components/Select/Select.tsx
mattwendzina: This is a bit tricky. I added the Radix label but this didn't solve the issue. I've added `aria-label` to `SelectPrimitive.Trigger` - which is the the button, and this has worked. I've just passed on the `name` - so when screen reading it starts with the name of the Select, e.g. 'Courses, required box pop-up etc...'

packages/osc-ui/src/components/Select/select.scss
mattwendzina: Yep

packages/osc-ui/src/components/Select/select.scss
mattwendzina: Yep, updated to the `shadow` color token.

packages/osc-ui/src/components/Select/select.scss
mattwendzina: Done

packages/osc-ui/src/components/Select/select.scss
mattwendzina: Updated.

packages/osc-ui/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,118 @@
...
+        } = props;
+        const [value, setValue] = useState('');
+
+        const errorMessage = getFieldError(value, required);
+        const displayError = wasSubmitted && errorMessage;
+
+        const modifiers = useModifier('c-select', groupVariants);
+        const selectClasses = classNames('c-select', modifiers);
+
+        const setDescription = (desc: Props['description']) => {
```
stevenbister, line 56: Could this just be?
```suggestion
        const setDescription = (desc: Description) => {
```
As you've got a Description type set above

packages/osc-ui/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { useModifier } from '../../hooks/useModifier';
+import { classNames } from '../../utils/classNames';
+import { getFieldError } from '../../utils/getFieldError';
+import { Icon } from '../Icon/Icon';
+import { Label } from '../Label/Label';
+import './select.scss';
+
+type Description = {
+    label?: string;
+    icon?: React.ReactNode;
```
stevenbister, line 14: Can you go through and remove this `React.`s? Just so it matches our other components

packages/osc-ui/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,118 @@
...
+                    <div className="c-select__error-message">{errorMessage}</div>
+                ) : null}
+            </div>
+        );
+    }
+);
+
+export interface ItemProps extends ComponentPropsWithRef<typeof SelectPrimitive.Item> {}
+
+export const SelectItem: FC<ItemProps> = React.forwardRef<
```
stevenbister, line 103: Might as well start dropping these `FC` types now. There's a PR waiting to go in https://github.com/Open-Study-College/osc/pull/649 to clean up the others so if we can start omitting them in new PRs that would be great :)

stevenbister (changes requested):
Sorry, couple of other things I've noticed -- only minor

packages/osc-ui/src/components/Select/Select.spec.tsx
```diff
@@ -0,0 +1,33 @@
...
+    );
+
+    expect(screen.getByRole('combobox')).toBeInTheDocument();
+});
+test('should render a disabled select component', async () => {
+    const { container } = render(
+        <Select attributes={{ disabled: true }} selectName="courses-1" wasSubmitted>
+            <SelectItem value="a-level-psychology"> A Level Psychology </SelectItem>
+        </Select>
+    ) as any;
```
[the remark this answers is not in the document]
mattwendzina: I was having some issues  but those are sorted now so removed the type casting 👍

packages/osc-ui/src/components/Select/Select.spec.tsx
```diff
@@ -0,0 +1,33 @@
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+import { Select, SelectItem } from './Select';
+
+test('should render a select component', async () => {
```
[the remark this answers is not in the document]
mattwendzina: Good point, nope! Removed.

packages/osc-ui/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,117 @@
...
+import './select.scss';
+
+type SelectAttributes = {
+    disabled?: boolean;
+    required?: boolean;
+};
+
+type Description = {
+    label: string;
+    icon: React.ReactNode;
```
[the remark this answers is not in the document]
mattwendzina: Updated. I've made both optional as I think it's okay to have the Icon instead of the Label. It's an accessible Icon and it doesn't seem to cause any issues accessibility wise - still lists the name of the Select. Only thing is pressing the label doesn't open the Select... as far as I can see I've implemented this as per docs, so perhaps it doesn't operate like this for a Select. I personally don't think this is a problem given that a dropdown is a pretty big space to click anyway.

packages/osc-ui/src/components/Select/Select.tsx
mattwendzina: Yeah good shout. Updated.

packages/osc-ui/src/components/Select/Select.tsx
mattwendzina: Done.

packages/osc-ui/src/components/Select/Select.tsx
mattwendzina: Updated 👍

packages/osc-ui/src/components/Select/Select.tsx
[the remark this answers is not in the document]
stevenbister: Cool! Yeah I think that's default for selects not to be opened when you click the label so that's all good

stevenbister (approved):
LGTM!

hi-im-si (changes requested):
When you click on the label, it should focus the select to enable easy keyboard access for the user - can we apply this update

PR #652, opened 2023-01-10
Feature/avatar notification dot

mattwendzina:
Closes # 

## 📝 Description
Add notification dot and count into Avatar

## 🚀 New behavior

Allows the use of a notification dot and counter in conjunction with the Avatar

### DIFF
--- a/packages/osc-ui/src/components/Avatar/Avatar.tsx
+++ b/packages/osc-ui/src/components/Avatar/Avatar.tsx
@@ -6,12 +6,14 @@ import './avatar.scss';
 
 export interface AvatarProps {
     className?: string;
-    src?: string;
+    count?: number;
     name: string;
+    notification: boolean;
+    src?: string;
 }
 
 export const Avatar: FC<AvatarProps> = (props: AvatarProps) => {
-    const { className, src, name } = props;
+    const { className, count, name, notification, src } = props;
     const classes = classNames('c-avatar', className);
     const initials = name
         ? name
@@ -23,6 +25,11 @@ export const Avatar: FC<AvatarProps> = (props: AvatarProps) => {
     return (
         <AvatarPrimitive.Root className={classes}>
             <AvatarPrimitive.Image className="c-avatar__image" src={src} alt={name} />
+            {notification ? (
+                <div className="c-avatar__notification-badge">
+                    <span className="c-avatar__notification-badge--count">{count}</span>
+                </div>
+            ) : null}
             <AvatarPrimitive.Fallback className="c-avatar__fallback">
                 {initials}
             </AvatarPrimitive.Fallback>
--- a/packages/osc-ui/src/components/Avatar/Avatar.tsx
+++ b/packages/osc-ui/src/components/Avatar/Avatar.tsx
@@ -8,12 +8,12 @@ export interface AvatarProps {
     className?: string;
     count?: number;
     name: string;
-    notification: boolean;
+    notification?: { show: boolean; count?: number };
     src?: string;
 }
 
 export const Avatar: FC<AvatarProps> = (props: AvatarProps) => {
-    const { className, count, name, notification, src } = props;
+    const { className, name, notification = { show: false }, src } = props;);
     const classes = classNames('c-avatar', className);
     const initials = name
         ? name
@@ -25,9 +25,11 @@ export const Avatar: FC<AvatarProps> = (props: AvatarProps) => {
     return (
         <AvatarPrimitive.Root className={classes}>
             <AvatarPrimitive.Image className="c-avatar__image" src={src} alt={name} />
-            {notification ? (
+            {notification.show ? (
                 <div className="c-avatar__notification-badge">
-                    <span className="c-avatar__notification-badge--count">{count}</span>
+                    <span className="c-avatar__notification-badge--count">
+                        {notification.count}
+                    </span>
                 </div>
             ) : null}
             <AvatarPrimitive.Fallback className="c-avatar__fallback">
--- a/packages/osc-ui/src/components/Avatar/avatar.scss
+++ b/packages/osc-ui/src/components/Avatar/avatar.scss
@@ -1,4 +1,7 @@
+@use "../../styles/settings" as *;
+
 .c-avatar {
+    position: relative;
     display: inline-flex;
     align-items: center;
     justify-content: center;
@@ -10,10 +13,11 @@
     user-select: none;
 
     &__image {
+        position: relative;
         width: 100%;
         height: 100%;
-        object-fit: cover;
         border-radius: inherit;
+        object-fit: cover;
     }
 
     &__fallback {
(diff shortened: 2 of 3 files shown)


packages/osc-ui/src/components/Avatar/avatar.scss
```diff
@@ -1,31 +1,57 @@
...
-        font-size: 15px;
-        font-weight: 500;
-        line-height: 1;
+        width: 0.875em;
+        height: 0.875em;
+        background: var(--color-primary);
+        border-radius: 50%;
+        box-shadow: 0 0 0 2px var(--color-tertiary);
+
+        &--count {
```
stevenbister, line 52: Might be worth adding `font-variant-numeric: tabular-nums;`? Not a super obvious change but it'll keep the number values a consistent height/width

stevenbister (changes requested):
Looks good! My only thought is we'll need to handle large notification counts, for example if someone has something like 200 notifications the value looks a bit tight and if it goes into 4 figures is overflows completely.
image
image

Perhaps we could add some logic that will set a max count and when the passed value goes over that we add a "+" or something so it would look something like this
image

Doesn't have to be 99 but I feel like that's a good point to set it as three figure numbers are where it starts to get funky

hi-im-si:
It would be great to look at including an Avatar fallback background if there is no image.

Some examples:
https://github.com/bukinoshita/hash-avatar

```
import * as React from 'react';
import hash from 'string-hash';
import color from 'tinycolor2';
function Fallback({ name }) {
  const hashed = hash(name);
  const c = color({ h: hashed % 360, s: 0.95, l: 0.5 });
  const c1 = c.toHexString();
  const c2 = c.triad()[1].toHexString();
  return (
    <svg
      role="img"
      aria-label={name}
      width="40"
      height="40"
      viewBox="0 0 80 80"
    >
      <defs>
        <linearGradient
          x1="0%"
          y1="0%"
          x2="100%"
          y2="100%"
          id={name}
        >
          <stop stopColor={c1} offset="0%" />
          <stop stopColor={c2} offset="100%" />
        </linearGradient>
      </defs>
      <g stroke="none" strokeWidth="1" fill="none">
        <rect
          fill={`url(#${name})`}
          x="0"
          y="0"
          width="100%"
          height="100%"
        />
      </g>
    </svg>
  );
}
```
This code converts the name into a hex string:
image

packages/osc-ui/src/components/Avatar/Avatar.tsx
```diff
@@ -21,9 +23,16 @@ export const Avatar: FC<AvatarProps> = (props: AvatarProps) => {
         : null;
 
     return (
-        <AvatarPrimitive.Root className={classes}>
+        <AvatarPrimitive.Root className={src ? classes : `${classes} c-avatar__fallback`}>
             <AvatarPrimitive.Image className="c-avatar__image" src={src} alt={name} />
-            <AvatarPrimitive.Fallback className="c-avatar__fallback">
+            {notification.show ? (
+                <div className="c-avatar__notification-badge">
```
hi-im-si, line 29: try to keep class names as short and concise as possible (i can appreciate that this is not always the case) so for this example i would suggest:

```suggestion
                <div className="c-avatar__badge">
```

To coincide with this, i would image we would utilise the badge component here?
`<Badge />`

packages/osc-ui/src/components/Avatar/avatar.scss
```diff
@@ -1,31 +1,57 @@
...
+            align-items: center;
+            justify-content: center;
+            width: 100%;
+            height: 100%;
+            background-color: var(--color-nonary);
+            color: var(--color-tertiary);
+            font-size: 15px;
+            font-weight: 500;
+            line-height: 1;
+        }
```
hi-im-si, line 36: is there any need for this to be a modifier? The initials should be the default fallback so would probably expect it to be just part of `c-avatar__fallback`

packages/osc-ui/src/components/Avatar/avatar.scss
```diff
@@ -1,31 +1,57 @@
...
+        &--initials {
+            display: flex;
+            align-items: center;
+            justify-content: center;
+            width: 100%;
+            height: 100%;
+            background-color: var(--color-nonary);
+            color: var(--color-tertiary);
+            font-size: 15px;
+            font-weight: 500;
```
hi-im-si, line 34: i believe there are some variables we can use for font-size, font-weight and line-height?

hi-im-si (changes requested):
also....see separate comment about the fallback background colour

mattwendzina:
> Looks good! My only thought is we'll need to handle large notification counts, for example if someone has something like 200 notifications the value looks a bit tight and if it goes into 4 figures is overflows completely. image image
> 
> Perhaps we could add some logic that will set a max count and when the passed value goes over that we add a "+" or something so it would look something like this image
> 
> Doesn't have to be 99 but I feel like that's a good point to set it as three figure numbers are where it starts to get funky

Updated 👍

packages/osc-ui/src/components/Avatar/Avatar.tsx
mattwendzina: Updated classnames and utilised Badge component 👍

packages/osc-ui/src/components/Avatar/avatar.scss
mattwendzina: Done

packages/osc-ui/src/components/Avatar/avatar.scss
mattwendzina: Added

packages/osc-ui/src/components/Avatar/Avatar.tsx
```diff
@@ -21,10 +46,22 @@ export const Avatar: FC<AvatarProps> = (props: AvatarProps) => {
...
-            <AvatarPrimitive.Fallback className="c-avatar__fallback">
-                {initials}
+            {notification.show ? (
+                <Badge
+                    className={
+                        notification.count
+                            ? 'c-avatar__badge c-avatar__badge--count'
+                            : 'c-avatar__badge'
+                    }
+                    badgeName={+notification.count > 99 ? '99+' : notification.count}
```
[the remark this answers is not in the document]
mattwendzina: Yeah, good idea. I've added one. I've made the suffix optional which perhaps it doesn't need to be, can't think of a use case but left it that way incase there were.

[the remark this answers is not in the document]
stevenbister: Noice!

stevenbister (approved):
LGTM!

packages/osc-ui/src/components/Avatar/Avatar.tsx
```diff
@@ -21,10 +47,26 @@ export const Avatar: FC<AvatarProps> = (props: AvatarProps) => {
...
+        <AvatarPrimitive.Root className={src ? classes : `${classes} c-avatar__fallback`}>
             <AvatarPrimitive.Image className="c-avatar__image" src={src} alt={name} />
-            <AvatarPrimitive.Fallback className="c-avatar__fallback">
-                {initials}
+            {notification.show ? (
+                <Badge
+                    className={
+                        notification.count
+                            ? 'c-avatar__badge c-avatar__badge--count'
+                            : 'c-avatar__badge'
```
hi-im-si, line 57: The Badge component will need completing and then rolling out on to this

hi-im-si: This can be done afterwards!

PR #653, opened 2023-01-10
feat: add TextInput component, styles and tests

### LINKED ISSUE #614: Forms text input
### Link task to story

### Description

n/a

mattwendzina:
Closes #614 

## 📝 Description

Adds TextInput component. 

N/A

## 🚀 New behavior

Adds TextInput component and covers all required styles

## 📝 Additional Information
There is one accessibility issue on the error message text. It says there is insufficient colour contrast of 4.05 and the expected ratio is 4:5:1. So it's not far off but maybe we need to make the error red colour a touch darker?

### DIFF
--- a/packages/osc-ui/src/components/Label/label.scss
+++ b/packages/osc-ui/src/components/Label/label.scss
@@ -7,12 +7,12 @@
     font-weight: $weight-reg;
     cursor: pointer;
 
+    &__required {
+        padding-left: 0.063em;
+        font-size: var(--font-scale-centi);
+    }
+
     &--bold {
         font-weight: $weight-bold;
     }
-}
-
-.c-label__required {
-    padding-left: 1px;
-    font-size: var(--font-scale-centi);
 }
--- a/packages/osc-ui/src/components/Label/label.scss
+++ b/packages/osc-ui/src/components/Label/label.scss
@@ -1,0 +1,18 @@
+/* stylelint-disable selector-class-pattern */
+@use "../../styles/settings" as *;
+/* stylelint-disable plugin/selector-bem-pattern */
+
+.c-label {
+    font-size: var(--font-scale-centi);
+    font-weight: $weight-reg;
+    cursor: pointer;
+
+    &--bold {
+        font-weight: $weight-bold;
+    }
+}
+
+.c-label__required {
+    padding-left: 1px;
+    font-size: var(--font-scale-centi);
+}
--- a/packages/osc-ui/src/components/Label/label.scss
+++ b/packages/osc-ui/src/components/Label/label.scss
@@ -4,7 +4,6 @@
 /* stylelint-disable plugin/selector-bem-pattern */
 
 .c-label {
-    font-size: var(--font-scale-base);
     font-weight: $fw-reg;
     cursor: pointer;
 
--- a/packages/osc-ui/src/components/Label/label.scss
+++ b/packages/osc-ui/src/components/Label/label.scss
@@ -1,15 +1,15 @@
 /* stylelint-disable selector-class-pattern */
+@use "sass:math";
 @use "../../styles/settings" as *;
 /* stylelint-disable plugin/selector-bem-pattern */
 
 .c-label {
-    font-size: var(--font-scale-centi);
+    font-size: var(--font-scale-base);
     font-weight: $weight-reg;
     cursor: pointer;
 
     &__required {
         padding-left: 0.063em;
-        font-size: var(--font-scale-centi);
     }
 
     &--bold {
--- a/packages/osc-ui/src/components/Label/label.scss
+++ b/packages/osc-ui/src/components/Label/label.scss
@@ -5,14 +5,14 @@
 
 .c-label {
     font-size: var(--font-scale-base);
-    font-weight: $weight-reg;
+    font-weight: $fw-reg;
     cursor: pointer;
 
     &__required {
-        padding-inline-start: #{math.div(1, $base-font-size)}em;
+        padding-inline-start: #{math.div(1, $base-fs)}em;
     }
 
     &--bold {
-        font-weight: $weight-bold;
+        font-weight: $fw-bold;
     }
 }
--- a/packages/osc-ui/src/components/Label/label.scss
+++ b/packages/osc-ui/src/components/Label/label.scss
@@ -9,7 +9,7 @@
     cursor: pointer;
 
     &__required {
-        padding-left: 0.063em;
+        padding-inline-start: #{math.div(1, $base-font-size)}em;
     }
 
     &--bold {
--- a/packages/osc-ui/src/components/TextInput/TextInput.tsx
+++ b/packages/osc-ui/src/components/TextInput/TextInput.tsx
@@ -57,7 +57,9 @@ export const TextInput = forwardRef<HTMLInputElement, Props>((props: Props, forw
         ...rest
     } = props;
 
-    const [value, setValue] = useState<string | number | readonly string[]>('');
+    const [value, setValue] = useState<string | number | readonly string[]>(
+        defaultValue ? defaultValue : ''
+    );
 
     const errorMessage = getFieldError(value, required);
     const displayError = wasSubmitted && errorMessage;
@@ -71,13 +73,6 @@ export const TextInput = forwardRef<HTMLInputElement, Props>((props: Props, forw
 
     const iconModifier = useModifier('c-input__icon', variants);
     const iconClasses = classNames('c-input__icon', iconModifier);
-
-    useEffect(() => {
-        if (defaultValue) {
-            setValue(defaultValue);
-        }
-        // eslint-disable-next-line react-hooks/exhaustive-deps -- only needs to run on mount
-    }, []);
 
     return (
         <div className="c-input__outer-container">
--- a/packages/osc-ui/src/components/TextInput/TextInput.tsx
+++ b/packages/osc-ui/src/components/TextInput/TextInput.tsx
@@ -18,6 +18,8 @@ type IconType = {
 type Action = {
     icon?: IconType;
     type: string;
+    variant?: Variants;
+    size?: 'sm' | 'md' | 'lg';
 };
 
 type Variants = 'secondary' | 'tertiary' | 'quaternary';
@@ -86,7 +88,11 @@ export const TextInput = forwardRef<HTMLInputElement, Props>((props: Props, forw
         </div>
     );
 
-    const InputButton = ({ action: { icon, size, variant } }) => {
+    const InputButton = (props: { action: Action }) => {
+        const {
+            action: { icon, size, variant },
+        } = props;
+
         return (
             <Button className="c-input__button" variant={variant} size={size}>
                 {icon ? <AccessibleIcon label={icon.label}>{icon.content}</AccessibleIcon> : null}
--- a/packages/osc-ui/src/components/TextInput/TextInput.tsx
+++ b/packages/osc-ui/src/components/TextInput/TextInput.tsx
@@ -1,0 +1,109 @@
+import type { HTMLInputTypeAttribute, InputHTMLAttributes, ReactNode } from 'react';
+import React, { forwardRef, useEffect, useState } from 'react';
+import { useElement } from '../../hooks/useElement';
+import { useModifier } from '../../hooks/useModifier';
+import { classNames } from '../../utils/classNames';
+import { getFieldError } from '../../utils/getFieldError';
+import { Label } from '../Label/Label';
+import './text-input.scss';
+import { Button } from '../Button/Button';
+import { Icon } from '../Icon/Icon';
+
+type Action = {
+    label?: string;
+    icon?: ReactNode;
+    type: string;
+};
+export interface Props extends InputHTMLAttributes<HTMLInputElement> {
+    action?: Action;
+    icon?: { content: ReactNode; label: string };
+    type: HTMLInputTypeAttribute;
+    variants?: string[];
+    wasSubmitted?: boolean;
+}
+
+export const TextInput = forwardRef<HTMLInputElement, Props>((props: Props, forwardedRef) => {
+    const {
+        action,
+        defaultValue,
+        id,
+        icon,
+        name,
+        placeholder,
+        required,
+        type,
+        variants,
+        wasSubmitted = false,
+        ...rest
+    } = props;
+
+    const [value, setValue] = useState<string | number | readonly string[]>('');
+
+    const errorMessage = getFieldError(value, required);
+    const displayError = wasSubmitted && errorMessage;
+
+    const element = useElement('c-input', type);
+    const variantsModifier = useModifier(element, variants);
+    const inputClasses = classNames('c-input', element, variantsModifier);
+
+    const containerModifier = useModifier('c-input__container', variants);
+    const containerClasses = classNames('c-input__container', containerModifier);
+
+    const iconModifier = useModifier('c-input__icon', variants);
+    const iconClasses = classNames('c-input__icon', iconModifier);
+
+    useEffect(() => {
+        if (defaultValue) {
+            setValue(defaultValue);
+        }
+    }, []);
+
+    return (
+        <div className="c-input__outer-container">
+            <div
+                className={
+                    displayError
+                        ? `${containerClasses} c-input__container--error`
+                        : `${containerClasses}`
+                }
+            >
+                <input
+                    aria-invalid={displayError ? true : false}
+                    aria-describedby={displayError ? `${id}-error` : undefined}
+                    className={inputClasses}
+                    defaultValue={defaultValue}
+                    id={id}
+                    name={id}
+                    onChange={(event) => setValue(event.currentTarget.value)}
+                    placeholder={placeholder}
+                    ref={forwardedRef}
+                    type={type}
+                    {...rest}
+                />
+                <Label
+                    htmlFor={id}
+                    name={name}
+                    required={required}
+                    variants={value ? ['filled'] : null}
+                />
+                {icon ? (
+                    <div className={iconClasses}>
+                        <Icon label={icon.label}>{icon.content}</Icon>
+                    </div>
+                ) : null}
+                {displayError ? (
+                    <span className="c-input__error-message" role="alert" id={`${id}-error`}>
+                        {errorMessage}
+                    </span>
+                ) : null}
+            </div>
+            {action?.type === 'submit' ? (
+                <Button className="c-input__button">
+                    <Icon label={action?.label}>{action?.icon}</Icon>
+                </Button>
+            ) : null}
+        </div>
+    );
+});
+
+TextInput.displayName = 'TextInput';
--- a/packages/osc-ui/src/components/TextInput/TextInput.tsx
+++ b/packages/osc-ui/src/components/TextInput/TextInput.tsx
@@ -73,6 +73,29 @@ export const TextInput = forwardRef<HTMLInputElement, Props>((props: Props, forw
     const iconModifier = useModifier('c-input__icon', variants);
     const iconClasses = classNames('c-input__icon', iconModifier);
 
+    const InputError = () => (
+        <>
+            <div className="c-input__icon c-input__icon--error">
+                <Icon label={icon?.label}>{icon?.content}</Icon>
+            </div>
+            <span className="c-input__error-message" role="alert" id={`${id}-error`}>
+                {errorMessage}
+            </span>
+        </>
+    );
+
+    const InputIcon = () => (
+        <div className={iconClasses}>
+            <Icon label={icon.label}>{icon.content}</Icon>
+        </div>
+    );
+
+    const InputButton = () => (
+        <Button className="c-input__button">
+            <Icon label={action?.icon?.label}>{action?.icon?.content}</Icon>
+        </Button>
+    );
+
     return (
         <div className="c-input__outer-container">
             <div
@@ -100,27 +123,10 @@ export const TextInput = forwardRef<HTMLInputElement, Props>((props: Props, forw
                     required={required}
                     variants={value ? ['filled'] : null}
                 />
-                {icon && icon.type !== 'error' ? (
-                    <div className={iconClasses}>
-                        <Icon label={icon.label}>{icon.content}</Icon>
-                    </div>
-                ) : null}
-                {displayError ? (
-                    <>
-                        <div className="c-input__icon c-input__icon--error">
-                            <Icon label={icon?.label}>{icon?.content}</Icon>
-                        </div>
-                        <span className="c-input__error-message" role="alert" id={`${id}-error`}>
-                            {errorMessage}
-                        </span>
-                    </>
-                ) : null}
+                {icon && icon.type !== 'error' ? <InputIcon /> : null}
+                {displayError ? <InputError /> : null}
             </div>
-            {action?.type === 'submit' ? (
-                <Button className="c-input__button">
-                    <Icon label={action?.icon?.label}>{action?.icon?.content}</Icon>
-                </Button>
-            ) : null}
+            {action?.type === 'submit' ? <InputButton /> : null}
         </div>
     );
 });
--- a/packages/osc-ui/src/components/TextInput/TextInput.tsx
+++ b/packages/osc-ui/src/components/TextInput/TextInput.tsx
@@ -1,5 +1,5 @@
 import type { InputHTMLAttributes, ReactNode } from 'react';
-import React, { forwardRef, useEffect, useState } from 'react';
+import React, { forwardRef, useState } from 'react';
 import { useElement } from '../../hooks/useElement';
 import { useModifier } from '../../hooks/useModifier';
 import { classNames } from '../../utils/classNames';
--- a/packages/osc-ui/src/components/TextInput/TextInput.tsx
+++ b/packages/osc-ui/src/components/TextInput/TextInput.tsx
@@ -7,7 +7,7 @@ import { getFieldError } from '../../utils/getFieldError';
 import { Label } from '../Label/Label';
 import './text-input.scss';
 import { Button } from '../Button/Button';
-import { Icon } from '../Icon/Icon';
+import { AccessibleIcon } from '../Icon/Icon';
 
 type IconType = {
     content: ReactNode;
@@ -56,9 +56,7 @@ export const TextInput = forwardRef<HTMLInputElement, Props>((props: Props, forw
         ...rest
     } = props;
 
-    const [value, setValue] = useState<string | number | readonly string[]>(
-        defaultValue ? defaultValue : ''
-    );
+    const [value, setValue] = useState(defaultValue ? defaultValue : '');
 
     const errorMessage = getFieldError(value, required);
     const displayError = wasSubmitted && errorMessage;
@@ -75,9 +73,7 @@ export const TextInput = forwardRef<HTMLInputElement, Props>((props: Props, forw
 
     const InputError = () => (
         <>
-            <div className="c-input__icon c-input__icon--error">
-                <Icon label={icon?.label}>{icon?.content}</Icon>
-            </div>
+            <div className="c-input__icon c-input__icon--error">{icon?.content}</div>
             <span className="c-input__error-message" role="alert" id={`${id}-error`}>
                 {errorMessage}
             </span>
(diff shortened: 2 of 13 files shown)


packages/osc-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,125 @@
...
+     */
+    action?: Action;
+    /**
+     * An object that contains an Icon and a label for accessibility
+     */
+    icon?: IconType;
+    /**
+     * Sets the custom styles, e.g. "Secondary", "Tertiary"
+     */
+    variants?: string[];
```
stevenbister, line 33: I would maybe try and type the values that can be passed into this array (might need a generic?) but typing that we can only pass "primary", "secondary", "tertiary" etc. would be useful. This is probably the same in the Select component actually?

Another thought might be to remove the "error" variant from here. So when the `displayError` const is true then a `has-error` classname gets added. Might make the dx a little nicer?

packages/osc-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const containerClasses = classNames('c-input__container', containerModifier);
+
+    const iconModifier = useModifier('c-input__icon', variants);
+    const iconClasses = classNames('c-input__icon', iconModifier);
+
+    useEffect(() => {
+        if (defaultValue) {
+            setValue(defaultValue);
+        }
+    }, []);
```
stevenbister, line 75: Does this need a dependency? If not might be worth adding
```ts
// eslint-disable-next-line react-hooks/exhaustive-deps
```
With a note as to why its disabled

stevenbister (changes requested):
Nice man! Couple of little comments but otherwise all good 😎

packages/osc-ui/src/components/TextInput/TextInput.tsx
mattwendzina: Added a union type. I had a look into generics but not sure how you'd use on in this case. If you can think of a better way of doing it I'd be keen to know = )

packages/osc-ui/src/components/TextInput/TextInput.tsx
mattwendzina: Done 👍

mattwendzina:
This is now rebased and using the correct tokens.

packages/osc-ui/src/components/TextInput/TextInput.tsx
stevenbister: Ahaa no sorry that's my bad I meant union not generic 😅 What you've done is what I meant

packages/osc-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    const containerModifier = useModifier('c-input__container', variants);
+    const containerClasses = classNames('c-input__container', containerModifier);
+
+    const iconModifier = useModifier('c-input__icon', variants);
+    const iconClasses = classNames('c-input__icon', iconModifier);
+
+    useEffect(() => {
+        if (defaultValue) {
+            setValue(defaultValue);
```
stevenbister, line 77: Could we set this directly in the state and lose the useEffect?
```ts
const [value, setValue] = useState<string | number | readonly string[]>(defaultValue);
```

stevenbister (changes requested):

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,165 @@
...
+.c-input__outer-container {
+    display: flex;
+}
+
+.c-input__container {
+    position: relative;
+    display: flex;
+    flex-direction: column;
+
+    .c-input {
```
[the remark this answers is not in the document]
mattwendzina: Done

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,165 @@
...
+                background: none;
+                border: none;
+                border-bottom: solid 0.125em var(--color-error);
+                box-shadow: none;
+            }
+        }
+    }
+}
+
+::placeholder {
```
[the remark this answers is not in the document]
mattwendzina: Updated

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    &--error {
+        .c-input {
+            background-color: var(--color-error-20);
+            border: 1px solid var(--color-error);
+            box-shadow: 0 1px 2px var(--color-neutral-500), 0 0 5px 3px var(--color-septenary);
+
+            &__error-message {
+                order: 2;
```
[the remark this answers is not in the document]
mattwendzina: This is still required and does work despite what the inspector says lol. I've added `display:flex` though to keep it happy.

packages/osc-ui/src/components/Label/label.scss
```diff
@@ -0,0 +1,18 @@
...
+    font-size: var(--font-scale-centi);
+    font-weight: $weight-reg;
+    cursor: pointer;
+
+    &--bold {
+        font-weight: $weight-bold;
+    }
+}
+
+.c-label__required {
```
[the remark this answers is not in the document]
mattwendzina: Done 👍

packages/osc-ui/src/components/TextInput/TextInput.tsx
mattwendzina: Good shout. I seemed to remember reading a long time ago that this wasn't the best way to do it, but just read another article that advocates it and explains the differences, main one being that it will be available on first render, plus it's a little cleaner!
https://stackoverflow.com/questions/66993812/usestate-vs-useeffect-setting-initial-value
I'll update this for the others too.

mattwendzina:
@hi-im-si - I should have done the Label as a separate PR, but didn't so I've made a change here, but seems pointless doing it in all the others - as it's such a small change. Letting you know so that the change here is the one that gets used with dealing with conflicts. Is this okay, or would you rather I make the change across all of them?

packages/osc-ui/src/components/TextInput/text-input.scss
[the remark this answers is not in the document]
stevenbister: Haha fair enough 🤷

packages/osc-ui/src/components/TextInput/TextInput.tsx
stevenbister: Sweet! Yeah any excuse to get rid of a useEffect lol! Eslint is complaining that the import is still there though

stevenbister (changes requested):
Looking really good mate! Just got that eslint warning to clear up and I think we're good to go 🚀

mattwendzina:
Thanks @stevenbister! - removed useEffect! Also just moved passing placeholder in explicitly - this can just go in with the 'rest'.

stevenbister (approved):
Nice! All good to go 🚀

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,162 @@
...
+
+        &--tertiary {
+            .c-input__icon {
+                top: 35%;
+            }
+        }
+
+        &--quaternary {
+            .c-input {
+                padding: 0.375em 2.563em 0.375em 0;
```
hi-im-si, line 110: see padding comments previous

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,162 @@
...
+                color: var(--color-neutral-600);
+            }
+
+            .c-label {
+                position: absolute;
+                top: -9999px;
+                left: -9999px;
```
hi-im-si, line 117: we should be using the `sr-only` class or method

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+
+        &--quaternary + .c-input__button {
+            svg {
+                width: 1.875em;
+                height: 1.875em;
```
hi-im-si, line 124: this is actually a good example of where `em` probably is better to use

packages/osc-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,128 @@
...
+                    ref={forwardedRef}
+                    type={type}
+                    {...rest}
+                />
+                <Label
+                    htmlFor={id}
+                    name={name}
+                    required={required}
+                    variants={value ? ['filled'] : null}
+                />
```
hi-im-si, line 102: is there a specific reason for why the label sits below the input and we are using flex order to readjust?

hi-im-si (commented):
need to discuss whether these should be actioned or we close and refernce in a separate fix branch. any thoughts @stevenbister

packages/osc-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,128 @@
...
+                    aria-describedby={displayError ? `${id}-error` : undefined}
+                    className={inputClasses}
+                    defaultValue={defaultValue}
+                    id={id}
+                    name={id}
+                    onChange={(event) => setValue(event.currentTarget.value)}
+                    ref={forwardedRef}
+                    type={type}
+                    {...rest}
+                />
```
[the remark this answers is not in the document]
stevenbister: Rather than add a wrapper we could use grid instead? And we'd benefit from the same (if not more?) control than if we used flex. 
I'd imagine something along these lines:

```scss
__container {
    display: grid;
    grid-template-rows: repeat(3, minmax(0, auto));  // could omit this and let the rows be set implicitly
}

__label,
__input,
__icon,
__error {
    // explicitly set the column position as we're going to be layering some items into the same row
    // and we don't want the implicit columns to be added.
    grid-column: 1; 
}

__label {
    grid-row: 1;
}

__input,
__icon {
    // layer the icon and input into the same row
    // this means we can remove the position: absolute / transform: translate stuff
    grid-row: 2;
}

__icon {
    justify-self: end;
    align-self: center;
}

__error {
  grid-row: 3;
}
```

packages/osc-ui/src/components/TextInput/TextInput.tsx
stevenbister: Aha I missed this! Yeah, unless there's a good reason it should sit before the input to keep the reading/tabbing order

packages/osc-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,128 @@
...
+                            <Icon label={icon?.label}>{icon?.content}</Icon>
+                        </div>
+                        <span className="c-input__error-message" role="alert" id={`${id}-error`}>
+                            {errorMessage}
+                        </span>
+                    </>
+                ) : null}
+            </div>
+            {action?.type === 'submit' ? (
+                <Button className="c-input__button">
```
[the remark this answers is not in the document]
stevenbister: I disagree on this one, buttons will assume the `type="submit"` when it's associated with a form by default

> The default behavior of the button. Possible values are:
> `submit`: The button submits the form data to the server. This is the default if the attribute is not specified for buttons associated with a `<form>`, or if the attribute is an empty or invalid value.

I think we discussed it a while back with Chakra because they add `type="button"` so we would have to explicitly change it, but our buttons don't do that

packages/osc-ui/src/components/TextInput/text-input.scss
stevenbister: We have a `<VisuallyHidden>` component we could wrap around this which uses that `sr-only` class

```tsx
<VisuallyHidden asChild>
    <Label ... />
</VisuallyHidden>
```

stevenbister:
> need to discuss whether these should be actioned or we close and refernce in a separate fix branch. any thoughts @stevenbister

I'm guessing most of these comments will be in the other form fields as well, so perhaps we merge these in and open a couple of issues to tackle them all in one PR rather than keeping track across several?

packages/osc-ui/src/components/TextInput/TextInput.tsx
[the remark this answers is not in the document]
hi-im-si: You are correct @stevenbister - happy to retract :)

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,162 @@
...
+/* stylelint-disable selector-class-pattern */
+@use "../../styles/settings" as *;
+/* stylelint-disable plugin/selector-bem-pattern */
+
+.c-input {
+    display: flex;
+    order: 1;
+    min-width: 20ch;
+    max-width: 30ch;
+    padding: 0.375em 2.563em 0.375em 1.25em;
```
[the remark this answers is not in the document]
stevenbister: Hmmm, yeah looking at this again I think these should be `px` as I guess we wouldn't want the space around the text to scale especially the x-axis as it could cause the actual "text area" to become quite narrow

packages/osc-ui/src/components/TextInput/TextInput.tsx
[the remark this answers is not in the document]
stevenbister: Here's a basic example:
https://codesandbox.io/s/white-platform-xqdneu?file=/src/styles.css

packages/osc-ui/src/components/TextInput/text-input.scss
hi-im-si: even better :)

packages/osc-ui/src/components/TextInput/TextInput.tsx
hi-im-si: > > need to discuss whether these should be actioned or we close and refernce in a separate fix branch. any thoughts @stevenbister
> 
> I'm guessing most of these comments will be in the other form fields as well, so perhaps we merge these in and open a couple of issues to tackle them all in one PR rather than keeping track across several?
> 
> Although having said that maybe worth waiting for @mattwendzina to answer a couple of your questions before closing?

Yeah unfortunately i think theres a couple in there that @mattwendzina will need to answer just to be 100% sure. i think the padding/units need to be addressed as a team - not sure if they have filtered across other work.

hi-im-si (changes requested):
see previous comments

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,162 @@
...
+                pointer-events: none;
+
+                &.c-label--filled {
+                    transform: translate(0, -1.25em) scale(0.8);
+                }
+            }
+
+            .c-label,
+            .c-input {
+                height: 1.875em;
```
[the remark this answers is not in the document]
hi-im-si: https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/#widths-and-heights-9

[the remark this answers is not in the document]
hi-im-si: we could even remove the `height` and then add in some top position on the label to bump it down inline with the design

packages/osc-ui/src/components/TextInput/TextInput.tsx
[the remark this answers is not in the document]
mattwendzina: Updated 👍

packages/osc-ui/src/components/TextInput/text-input.scss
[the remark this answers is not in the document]
mattwendzina: Done

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,162 @@
...
+
+    &__number {
+        padding: 0.375em 0.375em 0.375em 1.25em;
+    }
+
+    &__icon {
+        position: absolute;
+        top: 50%;
+        right: 5%;
+        transform: translateY(-50%);
```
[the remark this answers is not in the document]
mattwendzina: Replaced with the css grid approach 👍

packages/osc-ui/src/components/TextInput/text-input.scss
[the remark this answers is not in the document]
mattwendzina: Removed

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,162 @@
...
+            }
+
+            .c-input:focus + .c-label {
+                transform: translate(0, -1.25em) scale(0.8);
+            }
+        }
+
+        &--tertiary {
+            .c-input__icon {
+                top: 35%;
```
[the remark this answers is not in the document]
mattwendzina: Removed

packages/osc-ui/src/components/TextInput/text-input.scss
mattwendzina: Sorted

packages/osc-ui/src/components/TextInput/text-input.scss
mattwendzina: Added VisuallyHidden component into the Label component with an optional `hidden` prop for setting it.

packages/osc-ui/src/components/TextInput/TextInput.tsx
mattwendzina: I've left below as per our conversation.

mattwendzina:
I've also added reduced the size of the svg that sits in the button on the quaternary style since this has updated in the rebase to have correct button styling. This is obviously a bit different from the styling now anyway... so need to decided whether we want to keep like this or change back?

packages/osc-ui/src/components/TextInput/TextInput.tsx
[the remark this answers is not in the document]
stevenbister: Sweeeeeeet

packages/osc-ui/src/components/TextInput/text-input.scss
stevenbister: Noice! Nothing you need to change, but if you add the `asChild` prop it'll pass the props through to the child component and remove the extra div that gets added as it's using https://www.radix-ui.com/docs/primitives/utilities/slot 😎 

Like I said though, not need to change what you've done, just a fun fact :P

stevenbister:
> I've also added reduced the size of the svg that sits in the button on the quaternary style since this has updated in the rebase to have correct button styling. This is obviously a bit different from the styling now anyway... so need to decided whether we want to keep like this or change back?

You might have to adjust the icon size but if you use `variant="quaternary"` to give you the correct styling

packages/osc-ui/src/components/TextInput/text-input.scss
mattwendzina: Noice.

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,160 @@
...
+
+$input-py: calc(var(--space-scale-micro) - 3px);
+$input-pr: var(--space-scale-delta);
+$input-pl: var(--space-scale-epsilon);
+
+.c-input {
+    display: flex;
+    order: 1;
+    min-width: 20ch;
+    max-width: 30ch;
```
stevenbister, line 14: I don't think we should be setting a max width on this, looking at Figma the inputs have potential to be wider than this. I would probably control the size of the inputs with the container. So take this off here, then on `__container` class add a width of 100%, that way the layout of the form will be able to control the width. 

Does that make sense? Sorry didn't occur to me before :S

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,160 @@
...
+        pointer-events: none;
+    }
+
+    &__outer-container {
+        display: flex;
+    }
+
+    &__container {
+        display: grid;
+        flex-direction: column;
```
stevenbister, line 47: This `flex-direction` won't have an effect anymore now we're using `grid`

packages/osc-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,132 @@
...
+    );
+
+    const InputIcon = () => (
+        <div className={iconClasses}>
+            <AccessibleIcon label={icon.label}>{icon.content}</AccessibleIcon>
+        </div>
+    );
+
+    const InputButton = () => (
+        <Button className="c-input__button">
```
stevenbister, line 90: Not sure if you saw my comment yesterday about the quaternary style?

https://github.com/Open-Study-College/osc/pull/653#issuecomment-1412237632

stevenbister (changes requested):
Sorry Matt bet this feels endless 😅 Hopefully last couple of comments and fairly minor, I think

packages/osc-ui/src/components/TextInput/text-input.scss
mattwendzina: Updated 👍

packages/osc-ui/src/components/TextInput/text-input.scss
mattwendzina: Removed

packages/osc-ui/src/components/TextInput/TextInput.tsx
mattwendzina: Added now 👍

stevenbister (approved):
LGTM

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,160 @@
...
+    cursor: text;
+    grid-column: 1;
+    grid-row: 2;
+
+    &__icon {
+        justify-self: end;
+        align-self: center;
+        grid-row: 2;
+        grid-column: 1;
+        padding-right: 5%;
```
hi-im-si, line 24: use `$space-` variable

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,160 @@
+/* stylelint-disable no-descending-specificity */
+/* stylelint-disable selector-class-pattern */
+@use "../../styles/settings" as *;
+/* stylelint-disable plugin/selector-bem-pattern */
+
+$input-py: calc(var(--space-scale-micro) - 3px);
+$input-pr: var(--space-scale-delta);
+$input-pl: var(--space-scale-epsilon);
```
hi-im-si: switch to sass variables

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,160 @@
...
+
+.c-input {
+    display: flex;
+    order: 1;
+    padding: $input-py $input-pr $input-py $input-pl;
+    border: solid 1px var(--color-quinary);
+    font-size: var(--font-scale-base);
```
hi-im-si: should not need to define the font size as this should inherit from the `<body>`

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,160 @@
...
+        justify-self: end;
+        align-self: center;
+        grid-row: 2;
+        grid-column: 1;
+        padding-right: 5%;
+    }
+
+    &:focus {
+        outline: 3px solid transparent;
+        box-shadow: 0 0 0 2px var(--color-primary);
```
hi-im-si, line 29: use the variables defined in the new `settings/shadows.scss`

hi-im-si: the default focus styles should work here instead of defining it. these can be found in: `packages/osc-ui/src/styles/elements/_type.scss`

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,160 @@
...
+        width: 100%;
+
+        .c-label {
+            grid-row: 1;
+            grid-column: 1;
+        }
+
+        &--secondary {
+            .c-input {
+                border-radius: 6px;
```
hi-im-si: use a `$rad-` variable defined in `settings/radius.scss`

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,160 @@
...
+        &--tertiary,
+        &--quaternary {
+            position: relative;
+
+            .c-label {
+                display: flex;
+                align-items: center;
+                grid-row: 2;
+                transform-origin: top left;
+                transition: 100ms ease-in-out;
```
hi-im-si, line 66: use variables for animation timing:
```suggestion
                transition: $timing-s ease-in-out;
```

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,160 @@
...
+                pointer-events: none;
+
+                &.c-label--filled {
+                    top: 10px;
+                    transform: translate(0, -1.25em) scale(0.8);
```
hi-im-si, line 71: I would tend to avoid using `em` for transforms. Maybe @stevenbister could share his views on this?

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,160 @@
...
+        &--quaternary {
+            .c-input {
+                padding: $input-py $input-pr $input-py 0;
+                color: var(--color-neutral-600);
+            }
+
+            .c-label {
+                position: absolute;
+                top: -9999px;
+                left: -9999px;
```
hi-im-si, line 110: i guess because you have used the `VisuallyHidden` component, then you can remove these redundant styles?

packages/osc-ui/src/components/TextInput/text-input.scss
```diff
@@ -0,0 +1,160 @@
...
+            .c-input {
+                background-color: var(--color-error-20);
+                border: 1px solid var(--color-error);
+                box-shadow: 0 1px 2px var(--color-neutral-500), 0 0 5px 3px var(--color-septenary);
+
+                &__error-message {
+                    display: flex;
+                    order: 2;
+                    color: var(--color-error);
+                    font-size: var(--font-scale-base);
```
hi-im-si: is this needed - can it just be inherited by the `<body>`

hi-im-si (changes requested):

packages/osc-ui/src/components/TextInput/text-input.scss
mattwendzina: Done

packages/osc-ui/src/components/TextInput/text-input.scss
mattwendzina: Done

packages/osc-ui/src/components/TextInput/text-input.scss
mattwendzina: Removed

packages/osc-ui/src/components/TextInput/text-input.scss
mattwendzina: Done

packages/osc-ui/src/components/TextInput/text-input.scss
mattwendzina: Yep, good spot 👍

packages/osc-ui/src/components/TextInput/text-input.scss
mattwendzina: Yep, not needed. Removed.

packages/osc-ui/src/components/TextInput/text-input.scss
stevenbister: Hmmm, tbh I don't really have a strong opinion. I guess I generally would use `px` or `rem` but in this case as you're wanting to move the label to make space for some text I reckon `em` might be better? As you'd want it to move relative to the text size 🤔 

@hi-im-si Is there any reason you'd avoid using em? 

As an aside @mattwendzina This transform property is used in a couple of places (also on line 98 `.c-input:focus + .c-label`) so might be worth either consolidating the selectors or putting it in a variable

mattwendzina: Left as is, and moved label transform into a variable 👍

hi-im-si (approved):

PR #654, opened 2023-01-11
feat: add RadioGroup component, styles and tests

### LINKED ISSUE #616: Forms radio input
### Link task to story

### Description

n/a

mattwendzina:
Closes #616 

## 📝 Description

> Adds RadioGroup component

N/A

## 🚀 New behavior

This adds the Radix RadioGroup component and styling

## Additional Details

Some of the colours will need updating update colour tokens PR goes in.

### DIFF
--- a/packages/osc-ui/src/components/RadioGroup/RadioGroup.tsx
+++ b/packages/osc-ui/src/components/RadioGroup/RadioGroup.tsx
@@ -110,7 +110,7 @@ export const RadioItem = forwardRef<ElementRef<typeof RadioGroupPrimitive.Item>,
                 >
                     <RadioGroupPrimitive.Indicator className="c-radio-group__indicator" />
                 </RadioGroupPrimitive.Item>
-                <Label htmlFor={id}>{name}</Label>
+                <Label htmlFor={id} name={name} />
             </div>
         );
     }
--- a/packages/osc-ui/src/components/RadioGroup/RadioGroup.tsx
+++ b/packages/osc-ui/src/components/RadioGroup/RadioGroup.tsx
@@ -16,9 +16,9 @@ export interface RadioGroupProps extends ComponentPropsWithoutRef<typeof RadioGr
      */
     children: ReactNode;
     /**
-     * An optional description for the radio group
+     * A description for the radio group
      */
-    description?: { id: string; value: string };
+    description: { id: string; value: string };
     /**
      * The name of the group. Submitted with its owning form as part of a name/value pair.
      */
@@ -55,15 +55,12 @@ export const RadioGroup = (props: RadioGroupProps) => {
     const displayError = wasSubmitted && errorMessage;
 
     return (
-        <div>
-            {description ? (
-                <p className="c-radio-group__description" id={description?.id}>
-                    {description?.value}
-                </p>
-            ) : null}
+        <fieldset>
+            <legend className="c-radio-group__description" id={description.id}>
+                {description.value}
+            </legend>
             <RadioGroupPrimitive.Root
-                aria-label={name}
-                aria-labelledby={description?.id}
+                aria-labelledby={description.id}
                 className={
                     displayError ? `${radioGroupClasses} c-radio-group--error` : radioGroupClasses
                 }
@@ -80,7 +77,7 @@ export const RadioGroup = (props: RadioGroupProps) => {
                     </div>
                 ) : null}
             </RadioGroupPrimitive.Root>
-        </div>
+        </fieldset>
     );
 };
 
--- a/packages/osc-ui/src/components/RadioGroup/RadioGroup.tsx
+++ b/packages/osc-ui/src/components/RadioGroup/RadioGroup.tsx
@@ -1,0 +1,121 @@
+import * as RadioGroupPrimitive from '@radix-ui/react-radio-group';
+import type { ComponentPropsWithoutRef, ComponentPropsWithRef, ElementRef, ReactNode } from 'react';
+import React, { forwardRef, useState } from 'react';
+import { Label } from '../Label/Label';
+
+import { useModifier } from '../../hooks/useModifier';
+import { classNames } from '../../utils/classNames';
+import { getFieldError } from '../../utils/getFieldError';
+import './radio-group.scss';
+
+type Variants = 'secondary' | 'tertiary';
+
+export interface RadioGroupProps extends ComponentPropsWithoutRef<typeof RadioGroupPrimitive.Root> {
+    /**
+     * The Radio Items
+     */
+    children: ReactNode;
+    /**
+     * An optional description for the radio group
+     */
+    description?: { id: string; value: string };
+    /**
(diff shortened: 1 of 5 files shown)


packages/osc-ui/src/components/RadioGroup/radio-group.scss
```diff
@@ -0,0 +1,101 @@
...
+    display: flex;
+    flex-direction: column;
+    gap: 0.625em;
+
+    &__item-container {
+        display: flex;
+        align-items: center;
+    }
+
+    &__item {
```
stevenbister, line 15: I think we need to add `cursor: pointer;` to this property. At the moment it's only on the label so it looks a bit like you can't click on the "dot"

packages/osc-ui/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    const modifiers = useModifier('c-radio-group', variants);
+    const radioGroupClasses = classNames('c-radio-group', modifiers);
+
+    const errorMessage = getFieldError(value, required);
+    const displayError = wasSubmitted && errorMessage;
+
+    return (
+        <div>
+            <p className="c-radio-group__description" id={description?.id}>
```
stevenbister: I think a `fieldset` and a `legend` would be better here as opposed to a `div` and a `p` it's better semantically.

I would also maybe consider how to handle what to do if a description isn't added. You've set the type as optional but there's no conditions here so if it's not added we'll get some empty markup.
Personally I would make it required as you'd always want to include a `legend` / description for your group.

packages/osc-ui/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,121 @@
...
+                {description?.value}
+            </p>
+            <RadioGroupPrimitive.Root
+                aria-label={name}
+                aria-labelledby={description?.id}
```
stevenbister, line 66: You don't need both `aria-label` and `lablledby`. Labelled by takes precedence over `aria-label` so you can probably just use that, that way it'll also match your legend above.

packages/osc-ui/src/components/RadioGroup/radio-group.scss
```diff
@@ -0,0 +1,101 @@
...
+            }
+        }
+    }
+
+    .c-label {
+        padding-left: 0.625em;
+    }
+
+    &[data-disabled] {
+        color: #b9b9b9;
```
stevenbister, line 98: Can this be a variable or hsl value?

packages/osc-ui/src/components/RadioGroup/radio-group.scss
```diff
@@ -0,0 +1,101 @@
...
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        width: 100%;
+        height: 100%;
+
+        &::after {
+            display: block;
+            width: 0.875em;
+            height: 0.875em;
```
stevenbister, line 42: I think these values should be calculated from the `__item` width and height. That way if anything changes we don't have to workout the size again

stevenbister (changes requested):
Nice man! Few comments for ya

packages/osc-ui/src/components/RadioGroup/RadioGroup.tsx
mattwendzina: Thanks mate, I've had a think and play with this. Basically I see `aria-label` working as a fallback. `Lablledby` is pointing to the description, but I've made this optional, perhaps wrongly so, but the actual designs don't have description text, I just added it so though perhaps there may be use cases where a written description isn't needed. In these cases it would fall back to the name given to the Label.

packages/osc-ui/src/components/RadioGroup/radio-group.scss
mattwendzina: Done

packages/osc-ui/src/components/RadioGroup/RadioGroup.tsx
mattwendzina: Good shout. I've updated this and agree it should be required. I've taken the optional chaining off, but I don't know whether it should be kept really. The TS will ensure correct typing but won't stop runtime errors so without setting a default I wonder whether we should always be using optional chaining?

packages/osc-ui/src/components/RadioGroup/RadioGroup.tsx
mattwendzina: So I wrote this comment before the above one lol. Given that I'm now making it required, I've removed `aria-label`.

packages/osc-ui/src/components/RadioGroup/radio-group.scss
mattwendzina: Yep, updated 👍

packages/osc-ui/src/components/RadioGroup/radio-group.scss
mattwendzina: I'm not sure exactly what you mean, how would I do this?

packages/osc-ui/src/components/RadioGroup/RadioGroup.tsx
stevenbister: Haha no worries

packages/osc-ui/src/components/RadioGroup/radio-group.scss
stevenbister: So you've got a width and height set to 1.25em right and the after has a width & height of 0.875. So the difference is 0.7 (0.875 / 1.25 = 0.7).

Now with that info you just need to do a calc, i'd probably set it up something like this

```scss
&__item {
    --radio-item-size: 2em;
    --radio-indicator-diff: 0.7;

    width: var(--radio-item-size);
    height: var(--radio-item-size);

    &:after {
        width: calc(var(--radio-item-size) * var(--radio-indicator-diff));
        height: calc(var(--radio-item-size) * var(--radio-indicator-diff));
    }
```

Now if we change the --radio-item-size at any point then the `after` indicator should scale with it :)

packages/osc-ui/src/components/RadioGroup/RadioGroup.tsx
mattwendzina: Yeah, good point, didn't think of that 👍

packages/osc-ui/src/components/RadioGroup/radio-group.scss
mattwendzina: Ahh, nice, I hadn't quite made the connection, but makes sense now! Thanks = ) Updated 👍

stevenbister: Yeah, in the grand scheme of things it's probably not that big a deal but I feel like it makes things a little more maintainable & scalable

stevenbister (approved):
LGTM

PR #655, opened 2023-01-12
Checkbox component

### LINKED ISSUE #615: Forms checkbox input
### Link task to story

### Description

n/a

mattwendzina:
Closes #615 

## 📝 Description

Adds Checkbox component

N/A

## 🚀 New behavior

This adds the Radix Checkbox component and styling

## 📝 Additional Information

For the secondary styling I've changed the Focus colour to `secondary` when selected because when the checkbox is selected it's the same colour as the Focus ring and there is no border colour so you can't differentiate it. See pic below:
Screenshot 2023-01-12 at 09 25 56

I'm using `aria-label` as the accessibility tab in storybook is complaining that it's a <button>, which it is underneath the hood because we're using Radix. It's still linked up to the Radix <Label /> component in same way as their docs which keeps them linked.


mattwendzina:
This is now rebased and using the correct colour tokens.

packages/osc-ui/src/components/Label/label.scss
```diff
@@ -0,0 +1,18 @@
...
+    font-size: var(--font-scale-centi);
+    font-weight: $weight-reg;
+    cursor: pointer;
+
+    &--bold {
+        font-weight: $weight-bold;
+    }
+}
+
+.c-label__required {
```
stevenbister, line 15: Think I mentioned this on a previous pr but can this be nested?

packages/osc-ui/src/components/Checkbox/checkbox.scss
```diff
@@ -0,0 +1,90 @@
...
+    display: flex;
+    flex-wrap: wrap;
+    align-items: center;
+    margin: 0.5em;
+
+    .c-label {
+        padding-left: 0.625em;
+    }
+
+    .c-checkbox {
```
stevenbister, line 16: Should this be nested within the container? I think ideally we'd want it to have a lower specificity if possible?

stevenbister (changes requested):

packages/osc-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { Label } from '../Label/Label';
+import './checkbox.scss';
+
+type Variants = 'secondary' | 'tertiary';
+
+export interface CheckboxProps extends ComponentPropsWithRef<typeof CheckboxPrimitive.Root> {
+    /**
+     * An optional description for the checkbox
+     */
+    description?: { id: string; value: string };
```
stevenbister, line 17: I think similar to #654 this should be required

packages/osc-ui/src/components/Label/label.scss
mattwendzina: Just updating this on the TextInput rather than doing it across the files - will add a comment there and let @hi-im-si know for resolving conflicts = )

packages/osc-ui/src/components/Checkbox/checkbox.scss
mattwendzina: Yep, good shout. Done 👍

packages/osc-ui/src/components/Checkbox/Checkbox.tsx
mattwendzina: Don't think it will always be required as you can have single checkboxes, and as per as discussion, looks like you can have a `<fieldset/>` without a `<legend/>`.

packages/osc-ui/src/components/Label/label.scss
stevenbister: Sounds good

packages/osc-ui/src/components/Checkbox/Checkbox.tsx
stevenbister: Yep! Definitely an interesting one! Just dropping this here for future reference
https://html.spec.whatwg.org/multipage/form-elements.html#the-fieldset-element

stevenbister (approved):
LGTM

packages/osc-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,107 @@
...
+            icon,
+            id,
+            name,
+            required,
+            value,
+            variants,
+            wasSubmitted = false,
+        } = props;
+
+        const [checked, setChecked] = useState<string | boolean>(false);
```
aaacafe786, line 61: gonna move this out of the checkbox component in a different pr

packages/osc-ui/src/components/Checkbox/checkbox.scss
```diff
@@ -0,0 +1,95 @@
...
+        pointer-events: none;
+
+        + .c-label {
+            color: var(--color-neutral-500);
+            pointer-events: none;
+        }
+    }
+
+    .o-icon {
+        &.check {
```
stevenbister, line 90: Can we change this class to `is-checked`?

packages/osc-ui/src/components/Checkbox/checkbox.scss
```diff
@@ -0,0 +1,95 @@
+/* stylelint-disable selector-class-pattern */
+@use "../../styles/settings" as *;
+/* stylelint-disable plugin/selector-bem-pattern */
+
+.c-checkbox {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 1.25em;
```
stevenbister, line 9: Perhaps these values could be variables?

packages/osc-ui/src/components/Checkbox/checkbox.scss
```diff
@@ -0,0 +1,95 @@
...
+    &__description {
+        flex: 1 100%;
+        font-size: $fs-s;
+    }
+
+    &__indicator {
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        width: 0.875em;
```
stevenbister, line 24: Same as before could these be variables? Or even better calculated from the checkbox size? Think we did something similar for the radios if I remember right?

packages/osc-ui/src/components/Checkbox/checkbox.scss
```diff
@@ -0,0 +1,95 @@
...
+            pointer-events: none;
+        }
+    }
+
+    .o-icon {
+        &.check {
+            width: 0.625em;
+            height: 0.625em;
+        }
+    }
```
hi-im-si, line 94: i would sub-class the icon rather than take this approach e.g. `.c-checkbox__icon` or `c-checkbox__checkmark`

packages/osc-ui/src/components/Checkbox/checkbox.scss
```diff
@@ -0,0 +1,95 @@
...
+
+        + .c-label {
+            color: var(--color-neutral-500);
+            pointer-events: none;
+        }
+    }
+
+    .o-icon {
+        &.check {
+            width: 0.625em;
```
stevenbister, line 91: Icon has a css variable `--icon-size` which sets the width and height, so you might be able to use that here instead?

stevenbister (changes requested):
Looks good, couple of minors/nitpicks

packages/osc-ui/src/components/Checkbox/checkbox.scss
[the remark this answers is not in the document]
stevenbister: `rem` would probably be best so we can scale it still, I think 🤔

packages/osc-ui/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,151 @@
+import type { Meta, Story } from '@storybook/react';
+import React, { useEffect, useRef } from 'react';
+
+import { CheckIcon } from '@radix-ui/react-icons';
```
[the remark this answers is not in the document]
mattwendzina: Deleted

packages/osc-ui/src/components/Checkbox/checkbox.scss
mattwendzina: Updated

packages/osc-ui/src/components/Checkbox/checkbox.scss
mattwendzina: Done 👍

packages/osc-ui/src/components/Checkbox/checkbox.scss
mattwendzina: Sorted

packages/osc-ui/src/components/Checkbox/checkbox.scss
mattwendzina: Done

packages/osc-ui/src/components/Checkbox/checkbox.scss
mattwendzina: Cool, sorted.

packages/osc-ui/src/components/Checkbox/checkbox.scss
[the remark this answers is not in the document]
mattwendzina: Updated to rem

PR #659, opened 2023-01-17
Feature/navbar

### LINKED ISSUE #179: building header component
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/175

### Description

E2E
Integration
Visibility in the header

### LINKED ISSUE #182: Building Navigation component
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/180

### Description

E2E
Integration
Interactions on how it hovers.

stevenbister:
- Closes #179
- Closes #182 

## 📝 Description

Adds a header, navbar and burger component to `osc-ui`

N/a

## 🚀 New behavior

Adds three new composable components, which can work together to build a header with a navbar.

The biggest chunk of work is in the Navbar component which is is inspired by Radix and consists of `Navbar`, `NavContent`, `NavItem`, `NavLink`, `NavList`, `NavSubMenu` and `NavTrigger`.
The `NavSubMenu` makes use of React's `useContext` hook to allow us to keep the component composable while still being able to pass some props through to it's children, such as an `isOpen` state. To access the context `NavSubMenu` is wrapper in a context provider and I've added a custom `useSubNavContext` hook to the end of the file which allows us to access it.

All styling has been added for each component, some styles are shared across files (ie. header.scss may contain some small styles that influence the navbar when it's nested)

## 📝 Additional Information

To generate the unique Ids for the content & trigger components I've installed lodash's uniqueId package and wrapped it in a `useMemo`. Once we fully upgrade to React 18 this can be dropped in favour of the new `useId` hook.

Once  #648 is merged we can replace the inline svgs with the `Icon` component and make use of the spritesheet.

I haven't added the search bar or any of the functionality to the buttons in the "actionNav" as they need dependencies we don't have on main yet (algolia/hydrogen)


stevenbister:
Storybook links:

- Buger: https://62cd418bce39ee57e0c8cc30-srntyupecx.chromatic.com/?path=/docs/osc-ui-burger--primary
- Navbar: https://62cd418bce39ee57e0c8cc30-srntyupecx.chromatic.com/?path=/docs/osc-ui-navbar--primary
- Header: https://62cd418bce39ee57e0c8cc30-srntyupecx.chromatic.com/?path=/docs/osc-ui-header--primary
(there's some weird overflow on here which has something to do with the iframe I think, generally that should break before it does that 🤔 ... if only container queries had better support...)

stevenbister:
Just thought, I probably need to style a single level nav a little more as it looks a bit funky if there's only one level of links

image

packages/osc-ui/src/components/Burger/Burger.tsx
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Sets the data-state attribute of the button.
+     *
+     * @default false
+     */
+    isOpen: boolean;
+}
+type BurgerRef = ForwardRefExoticComponent<BurgerProps & RefAttributes<HTMLButtonElement>>;
+
+export const Burger = forwardRef<ElementRef<BurgerRef>, ComponentPropsWithoutRef<BurgerRef>>(
```
mattwendzina, line 34: I still find some of this TS stuff pretty confusing lol, but from what I've understood I think it may be that one should use `ComponentPropsWithRef` when using `forwardRef`. I got this from the following article that I'd used a bit when building out some of the form stuff with `forwardRef`:
https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase/

They show the use case for `withoutProps` as a way of extending `button` props but then note that for when using `forwardRef` that you can use `ComponentPropsWithRef`.

packages/osc-ui/src/components/Burger/Burger.tsx
```diff
@@ -0,0 +1,57 @@
...
+     * The accessible name of the button.
+     */
+    label: string;
+
+    /**
+     * Sets the data-state attribute of the button.
+     *
+     * @default false
+     */
+    isOpen: boolean;
```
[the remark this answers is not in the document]
stevenbister: Ahh good catch! I decided to make it required after I'd written the tests. I found it easy to miss off and generally it'll be controlled by some state so should always be there.

I've updated the test file 👍

packages/osc-ui/src/components/Burger/Burger.tsx
stevenbister: Ahhh nice find! I hadn't come across that before I'd just been following the [Radix extending primitive pattern](https://www.radix-ui.com/docs/primitives/overview/styling#extending-a-primitive) but it does make more sense to use withRef! I've updated

packages/osc-ui/src/components/Burger/burger.scss
```diff
@@ -0,0 +1,65 @@
+@use "../../styles/settings" as *;
+@use "../../styles/tools" as *;
+
+$burger-size: 20px;
+$burger-bar-height: 4px;
+$burger-gap: 4px;
```
Tom-Pearce, line 6: Should these be dynamic from some var/settings somewhere else?

packages/osc-ui/src/components/Header/Header.tsx
```diff
@@ -1,19 +1,212 @@
...
+                    fill="#062134"
+                />
+                <path
+                    d="M266.153 17.4272H275.506C275.291 14.6577 273.792 12.7628 270.865 12.7628C268.509 12.6899 266.51 14.512 266.153 17.4272ZM266.153 18.8848C266.439 22.456 268.509 24.2781 271.864 24.2781C273.792 24.2781 275.006 23.695 276.648 22.6747L276.505 24.4238C275.077 25.3713 273.292 25.8086 271.65 25.8086C267.224 25.8086 264.368 22.7476 264.368 18.5204C264.368 14.6577 266.867 11.1594 271.079 11.1594C275.077 11.1594 277.219 14.3662 277.219 17.7916C277.219 18.3018 277.219 18.5933 277.148 18.8848H266.153Z"
+                    fill="#062134"
+                />
+            </svg>
```
Tom-Pearce, line 87: Are we planning to move the SVG out in the future, or should it be done now? Just seems messy to have it in here to me?

packages/osc-ui/src/components/Navbar/Navbar.spec.tsx
```diff
@@ -0,0 +1,352 @@
...
+                                level={0}
+                                value={item.label}
+                            />
+                        ))}
+                    </NavList>
+                </Navbar>
+            </MemoryRouter>
+        );
+
+        const navTrigger = screen.getByRole('button', { name: 'Courses' });
```
Tom-Pearce, line 329: This might be my ignorance of testing, but would it be better to have these dynamically reference the values found in `testNestedSubMenuNav`? Just thinking if someone was to change the values (could be as simple as spotting a spelling error) then it would cause it to fail at this point. 

Although at the same time I can appreciate that may be the point - ensuring that the final result matches the original data since we know the data as it's defined for the purpose of the test. Just a thought.

Tom-Pearce (commented):
Added a few code comments, more thoughts than requests.

Only other thing I noted is the dark mode styling is broke on Nav in storybook - dunno if this is something that should be working or something we're doing later down the line - makes more sense to me to do it as we're going though?

packages/osc-ui/src/components/Burger/burger.scss
stevenbister: Yes they should, I've updated it to be calculated from the base-size of 16px.

Re the dark mode, yeah that's probably something that needs discussing @hi-im-si, @mattwendzina any thoughts?
Probably makes sense to do it as we go, but I'm conscious we probably need to decide how the colours are gonna swap (i.e. what should the `--color-primary` dark version be.

packages/osc-ui/src/components/Header/Header.tsx
stevenbister: Yep! Was waiting for #648 to be merged, now that's done I've moved the logo into the spritesheet and we can use the `<Icon />` component instead 😎

packages/osc-ui/src/components/Navbar/Navbar.spec.tsx
stevenbister: I think in this case I wouldn't reference the data. Generally in FE testing you want to test based on what you'd expect the user to see, for example if I had something on there that was capitalising the text to `Courses` but the data was `courses` it would fail as a user should expect to see the capital `C`.

I guess I could wrap the reference in the test with the same function to capitalise it but I think this makes it clearer as to what I'd expect

Tom-Pearce: Fine by me, thought that might be the case

packages/osc-ui/src/styles/objects/_icon.scss
```diff
@@ -1,4 +1,4 @@
-.o-icon {
+.o-icon:not(.o-logo) {
```
hi-im-si, line 1: would probably switch this out to be a state class, something like `is-logo`. If the logo does have a corresponding file then it should be a component rather than an object

packages/osc-ui/src/components/Header/Header.tsx
```diff
@@ -1,19 +1,143 @@
...
-export const Header = (props: Props<HTMLDivElement>) => {
-    const { className, ...attr } = props;
-    const classes = classNames(className);
+/* -------------------------------------------------------------------------------------------------
+ * Logo
+ * -----------------------------------------------------------------------------------------------*/
+export const Logo = () => (
+    <RemixNavLink to="/" className="c-header__logo">
+        <AccessibleIcon label="Open Study College">
+            <Icon id="logo" size={[278, 33]} className="o-logo" />
```
hi-im-si, line 29: would probably switch this out to be a state class, something like `is-logo`. If the logo does have a corresponding file then it should be a component rather than an object

packages/osc-ui/src/components/Header/header.scss
```diff
@@ -1,9 +1,151 @@
...
+        grid-template-rows: repeat(2, 1fr);
+        gap: var(--space-scale-micro);
+    }
+
+    a,
+    button {
+        &:focus-visible {
+            box-shadow: 0 0 0 0.25rem var(--color-primary);
+        }
+    }
```
hi-im-si, line 27: when the new variables come across, there has been a global `:focus` style setup so you will probably be able to remove this

packages/osc-ui/src/components/Navbar/navbar.scss
```diff
@@ -0,0 +1,351 @@
...
+    @include mq(mq-token("desk")) {
+        font-weight: $weight-reg;
+    }
+
+    button,
+    a {
+        &:focus-visible {
+            box-shadow: 0 0 0 0.2rem var(--color-primary);
+        }
+    }
```
hi-im-si, line 25: when the new variables come across, there has been a global `:focus` style so this can probably be removed

hi-im-si (changes requested):
Mate, such a good job. Left a couple tiny bits and also noticed that the arrow was missing from the larger nav items:
image
Other than that, when the new variables come across that will be a fun job....NOT!

stevenbister:
>  also noticed that the arrow was missing from the larger nav items:

I knew I'd forgotten something 😅 

> Other than that, when the new variables come across that will be a fun job....NOT!

😭

packages/osc-ui/src/components/Header/Header.tsx
stevenbister: Ah yeah I did toss and turn about whether it should be a component or object! So are you thinking something like

```tsx
<Icon id="logo" size={[278, 33]} className="is-logo" />
```

packages/osc-ui/src/styles/objects/_icon.scss
stevenbister: Sorted! I decided to move the logo into it's own component file as it's gonna get used in the footer too. So now it's `c-logo` :)

packages/osc-ui/src/components/Header/header.scss
stevenbister: Done!

packages/osc-ui/src/components/Navbar/navbar.scss
stevenbister: Done!

packages/osc-ui/src/components/Header/Header.tsx
stevenbister: See other comment

packages/osc-ui/src/styles/utilities/_util.scss
```diff
@@ -197,15 +197,21 @@ $osc-config: (
...
     &-until\@tab {
         @include mq($mq-tab, max) {
             display: none;
         }
     }
 
     &-until\@desk {
-        @include mq($mq-desk, max) {
-            display: none;
+        @include mq(mq-token("desk"), max) {
```
hi-im-si, line 213: use sass variable `$mq-desk`

hi-im-si (changes requested):

packages/osc-ui/src/components/Header/header.scss
```diff
@@ -1,9 +1,150 @@
...
+            }
+        }
+
+        #{$self}__nav & {
+            background-color: var(--color-secondary);
+            color: var(--color-tertiary);
+            gap: 0;
+
+            a,
+            button {
```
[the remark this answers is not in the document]
stevenbister: Yeah, I see your point and generally I would agree. However in this case I left them as tag selectors as we haven't fully built the component that are to go in the "action bar" (search button, cart etc.) so I wanted to keep the selectors generic enough that we could drop anything in there and they would take on the required styles... Having said that this did remind me of the `Slot` primitive from Radix, so I've wrapped the children in a `Slot` with the class of `.c-header__action-item` so we can target that instead :D

packages/osc-ui/src/components/Header/header.scss
```diff
@@ -1,9 +1,150 @@
...
+
+        @include mq($mq-desk) {
+            padding-inline-end: $space-2xs;
+            grid-column: 2 / -1;
+            grid-row: 1;
+        }
+
+        /* stylelint-disable-next-line no-descending-specificity */
+        > a,
+        > button {
```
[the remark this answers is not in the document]
stevenbister: As above

packages/osc-ui/src/styles/objects/_container.scss
```diff
@@ -1,15 +1,35 @@
...
     width: 100%;
     max-width: #{$container-m}px;
     margin-right: auto;
     margin-left: auto;
-    padding-right: $padding;
-    padding-left: $padding;
+    padding-right: $padding-alt;
+    padding-left: $padding-alt;
+
+    @include mq(mq-token("desk")) {
```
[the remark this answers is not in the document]
stevenbister: Ahh knew I'd missed some! Done!

packages/osc-ui/src/styles/utilities/_util.scss
stevenbister: Done!

PR #661, opened 2023-01-20
Feature/nav sanity

### LINKED ISSUE #180: [S4.3]: Building Navigation
### Link to Epic

### Description

N/A

### Linked Tasks

- [ ] #181 
- [x] #182

### LINKED ISSUE #181: Make dynamic with sanity
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/180

### Description

n/a

stevenbister:
- Closes #180 
- Closes #181 

## 📝 Description

Adds the Sanity & osc-ecommerce connection to the header/navbar. Fully expecting some things I've missed here but hopefully a pretty solid first pass.

N/a

## 🚀 New behavior

Nav menus can be added in Sanity studio like so:

https://user-images.githubusercontent.com/23461173/213685900-47180905-c7dc-4d49-ab07-6082a16d7596.mov

To get the shape correct users can add either columns, links or feature links. Not sure if this is the most elegant way of doing it for a user of the CMS but I couldn't think of another way 😅 

You can then select which nav you want to use in the header in the settings area like this:

https://user-images.githubusercontent.com/23461173/220112212-0e553c10-4a00-4890-a975-673d6ea6326d.mov

In osc-ecommerce this will check if the id exists and run a query using that id

I discovered a couple of SSR issues with my original component in PR #659. Mainly the generated IDs were causing a hydration mismatch so I've created a hook called `useUniuqeId` which basically looks to see if we're on the client, if not it will return undefined and if we are then it will generate the ID.
This can be replaced with React's new `useID` (https://beta.reactjs.org/reference/react/useId) which is hydration safe once we upgrade to React 18.

## 📝 Additional Information

I may have uncovered an issue with our dynamic stylesheets, if you're viewing an un-cached page then they don't seem to load when navigating using Remix's Link or NavLink component for example:

https://user-images.githubusercontent.com/23461173/213688345-c06fbce6-5f38-4427-8bcf-83455539796f.mov

I tested it by moving the stylesheets into the root and having them load on all pages and that seems to work so it appears to be related to the [dynamic links](https://github.com/sergiodxa/remix-utils#dynamiclinks) whether that's a bug in the package or to do with our implementation I'll need to investigate. See #662


stevenbister:
Putting into draft for the time being until #659 is merged

stevenbister:
Styles look a bit wonky in ecommerce atm. Fix is in place here: https://github.com/Open-Study-College/osc/pull/754

Issue #663, opened 2023-01-27
Add card layout options


stevenbister:
### Link task to story

Story: #247

### Description

Layout:
Carousel
Grid
Grid Alternate (2 Columns w/ 3 Islands eg. https://www.figma.com/file/ZBn4Shpu2Tj5xigprYFBTD?node-id=225:5241#340260390)

Issue #250, opened 2022-10-06
Card: Make dynamic with Sanity


EnaamAlum:
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/247

### Description

n/a

PR #664, opened 2023-01-31
Feature/cards components

### LINKED ISSUE #249: Build Card Component
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/247

### Description

Image
Title
Copy
CTA
Badge

https://chakra-templates.dev/components/cards

Add testing
E2E
Integration

### LINKED ISSUE #595: Card styling/Theming
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/247

### Description

n/a

stevenbister:
- Closes #249 
- Closes #595 

## 📝 Description

Adds the card components to osc-ui.

As there's already quite a lot going on in this PR I've decided to split #250 & #663 into separate PRs. There is some Sanity stuff in this one as I started there at the beginning of this work to get an idea of how the data might be moved out of Sanity. Feel free to review that but it's likely I'll change it in further work now I'm happy with how the cards will work.

I've decided to split the cards into individual components which are just a wrapper around the base `Card` component. This felt like the most straightforward way to limit the modifier classes and allow `variant` controls for certain types (e.g. blog allows `featured`)

The scss feels a bit untidy IMO but this is the happiest I've been with it so... 🤷 

Other thing to note is I've set an inline `max-width` within the stories to constrain the cards, as I envision the grid/container they will sit in will handle the widths normally.

N/a

## 🚀 New behavior

Card components and styles added to osc-ui


packages/osc-ui/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+};
+HasCustomHeadingText.parameters = {
+    docs: {
+        description: {
+            story: 'You can change the size and style of the heading text by adding the `isSmall` prop to the `<CardTitle>`.<br>To change the color add one of the `u-color-<color>` utility classes such as `u-color-primary`.',
+        },
+    },
+};
+
+export const BlockLink = Template.bind({});
```
mattwendzina, line 147: More of a comment on the UI/UX. Looking at this in storybook it seems a bit weird to have a clickable card with a button inside of it. I think that would be pretty confusing for users. Also, in terms of tabbing - only the button itself seems to be tabbable, if the card is clickable then should they be able to tab and click on this too?

packages/osc-ui/src/components/Card/CourseCard.stories.tsx
```diff
@@ -0,0 +1,330 @@
...
+                component:
+                    'The course variant of the card is mainly used for displaying course content from the cms.',
+            },
+        },
+    },
+} as Meta;
+
+// TODO: Tooltip on wishlist button
+// TODO: Badge needs styling
+// TODO: Popover need styling
```
mattwendzina, line 49: There's a few of these TODO comments dotted around. Are these now complete and need deleting?

packages/osc-ui/src/components/Card/Card.spec.tsx
```diff
@@ -0,0 +1,379 @@
...
+            </BlogCard>
+        );
+
+        expect(document.querySelector('.c-card--blog')).toHaveClass('c-card--media-object');
+    });
+});
+
+describe('Collection Card', () => {
+    test('adds the sm class to the blog card', () => {
+        render(
```
mattwendzina, line 252: Not a big one, but could save some lines of code by creating a separate setup for next three tests as they all render the same code.

packages/osc-ui/src/components/Card/CourseCard.stories.tsx
stevenbister: Nah; they need actioning and were out of scope of this PR so they're in there to remind us to update the card once they're done

packages/osc-ui/src/components/Card/Card.stories.tsx
stevenbister: The example here probably isn't the best for it, the featured blog card is better. But I disagree, on touch devices I think most people will just tap on the card and having the option for both doesn't really detract from anything. They will both take the same action and I've added a condition in there to avoid any bubbling / duplicate events.

Re the tabbing, that's intentional. If you make the whole card focusable then the entire card contents gets read out, which might not be the best experience. It's based on this article https://css-tricks.com/block-links-the-search-for-a-perfect-solution/ (and this is also worth a read: https://inclusive-components.design/cards/)

Thinking about it though perhaps I should look at adding an aria-label to the button to improve the context, so it doesn't just read out "read more" 🤔

mattwendzina:
This looks awesome mate! 🙌 
Looks like there are a few accessibility issues in Storybook - some violations and some incomplete. All appear to be related to colour contrast. 
Other than that I've left a few little comments but nothing major.

packages/osc-ui/src/components/Card/Card.stories.tsx
mattwendzina: Ahh, okay yeah this makes a lot more sense when you see the Featured blog card. Those articles are really helpful. From what I can gather in the Featured Blog card the button is working as a CTA which can be helpful to mobile viewers who don't necessarily know the card is clickable, but presumably it can also be used as a separate link, perhaps similar to how the second article discusses the author link. Anyway, I think I just struggled getting the Card one because of the lack of context.
Agree that a bit more context to screen readers than 'read more' would be helpful, like maybe reading out the card title content?

stevenbister: Yeah for sure! I quite like the pattern, but agree that the card I used in the demo makes it a bit unclear. I'll update it to the featured blog one.

And re the screen reader bit, yeah I think so too.

hi-im-si:
Suggestion to move the CTA to the bottom right on the Featured Full Width Blog Card as below:
image

mattwendzina (commented):
Looks good mate. There does appear to be a few accessibility violations left though on colour contrasts. Appears to be the Primary colour against white, e.g. the h2 on 'Had custom heading text' Card variant. It is our primary colour though, so not sure how much we can do about this. Might be worth seeing what @hi-im-si thinks.
Other is usage of neutral-600 against white, Used for example as subtitle in Primary/Has Round Image of Card variant, so guess these may need to become neutral-700?
There is also a couple of incomplete's too. These seem to be partial clashes of colours against our image type backgrounds so may be hard to avoid I guess if it is the design.

mattwendzina:
> > Looks good mate. There does appear to be a few accessibility violations left though on colour contrasts. Appears to be the Primary colour against white, e.g. the h2 on 'Had custom heading text' Card variant. It is our primary colour though, so not sure how much we can do about this. Might be worth seeing what @hi-im-si thinks.
> 
> Ah yeah, I wonder if upping the font-weight would help here 🤔 e.g. https://webaim.org/resources/contrastchecker/?fcolor=E7136B&bcolor=FFFFFF Will see if Si has any thoughts
> 
> Other is usage of neutral-600 against white, Used for example as subtitle in Primary/Has Round Image of Card variant, so guess these may need to become neutral-700?
> 
> Good catch -- will update that
> 
> There is also a couple of incomplete's too. These seem to be partial clashes of colours against our image type backgrounds so may be hard to avoid I guess if it is the design.
> 
> Yeah, I think when it comes to background images/gradients that will be down to the content editor to make a judgement on what background/colour to use there

Cool. Yeah, I did wonder about whether upping the font weight would work too.

Other than the primary colour contrast issue which I guess Si may have thoughts on, I'm happy to give the 👍

stevenbister:
The test passes if it's bold (`$fw-bold`) I don't _hate_ it but it's certainly not as elegant :S

image

packages/osc-ui/src/components/Card/card.scss
```diff
@@ -0,0 +1,644 @@
...
+            #{$self}__subttl {
+                grid-row: auto;
+                color: $card-color;
+            }
+
+            #{$self}__body {
+                h4,
+                h5,
+                h6,
+                li {
```
hi-im-si, line 557: As discussed before, would prefer to subclass elements over targetting html tags. mixing html semantics with styling can be problematic

packages/osc-ui/src/components/Card/card.scss
```diff
@@ -0,0 +1,644 @@
...
+@layer components {
+    .c-card {
+        --card-border: #{$card-border};
+
+        $self: &;
+
+        width: 100%;
+        overflow: hidden;
+        background-color: var(--color-tertiary);
+        box-shadow: 0 4px 20px rgb(0 0 0 / 10%);
```
hi-im-si, line 48: if we can utilise a variable as specified in `packages/osc-ui/src/styles/settings/_shadows.scss`. I think this is another good example of the design not having uniformity in some of the tokens. the closest variable i can find to this is: `$shadow-m`

hi-im-si (changes requested):

packages/osc-studio/schemas/documents/team.tsx
```diff
@@ -0,0 +1,57 @@
...
+    //     {
+    //         default: true,
+    //         name: 'editorial',
+    //         title: 'Editorial',
+    //     },
+    //     {
+    //         name: 'seo',
+    //         title: 'SEO',
+    //     },
+    // ],
```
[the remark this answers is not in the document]
stevenbister: Ooops that shouldn't be in there 😅

packages/osc-ui/src/components/Card/card.scss
stevenbister: I guess my argument here would be that I'm trying to target the elements coming out of the WYSIWYG which don't have classes, so do we need to add classes to those elements?

hi-im-si (approved):

PR #665, opened 2023-02-01
Feature/admin api

### LINKED ISSUE #584: [S3.9]: Admin Service (SS Tasks)
### Link to Epic

### Description

_No response_

### Linked Tasks

#585 

_No response_

### LINKED ISSUE #586: [S3.10]: Admin Endpoint
### Link to Epic

### Description

n/a

### Linked Tasks

#587 
#588 
#589 
#590 
#591 

_No response_

Tom-Pearce:
- Closes #584 
- Closes #586

## 📝 Description

Adds a skeleton for our internal administrative API. Exposes a basic query to get unassigned students for tutor allocation

None

## 🚀 New behavior

Exposes endpoint for getting unassigned students

### DIFF
--- a/packages/osc-api-admin/src/index.ts
+++ b/packages/osc-api-admin/src/index.ts
@@ -6,8 +6,8 @@ dotenv.config();
 async function startServer() {
     const { url } = await startStandaloneServer(server(), {
         context: async ({ req }) => {
-            // @ts-ignore - Type 'string[]' is not assignable to type 'string'. user will never be an array (comes from gateway)
-            const user = req.headers.user ? JSON.parse(req.headers.user) : null;
+            const isUser = req.headers?.user as string;
+            const user = isUser ? JSON.parse(isUser) : null;
 
             return {
                 user,
--- a/packages/osc-api-admin/src/index.ts
+++ b/packages/osc-api-admin/src/index.ts
@@ -1,0 +1,21 @@
+import { startStandaloneServer } from '@apollo/server/standalone';
+import * as dotenv from 'dotenv';
+import { server } from '~/server';
+dotenv.config();
+
+async function startServer() {
+    const { url } = await startStandaloneServer(server(), {
+        context: async ({ req }) => {
+            // @ts-ignore - Type 'string[]' is not assignable to type 'string'. user will never be an array (comes from gateway)
+            const user = req.headers.user ? JSON.parse(req.headers.user) : null;
+
+            return {
+                user,
+            };
+        },
+        listen: { port: 4005 },
+    });
+    console.info(`🚀 Admin Microservice ready at: ${url}`);
+}
+
+void startServer();
--- a/packages/osc-api-admin/src/resolvers/resolver.ts
+++ b/packages/osc-api-admin/src/resolvers/resolver.ts
@@ -1,0 +1,16 @@
+import type { getUnassignedStudentsArgs } from '~/types/arguments';
+import type { AdminContext } from '~/types/interfaces';
+import * as admin from '~/utils/admin';
+
+export const resolvers = {
+    Query: {
+        unassignedStudents: async (
+            _: undefined,
+            args: getUnassignedStudentsArgs,
+            context: AdminContext
+        ) => {
+            return await admin.unassignedStudents(args.limit);
+        },
+    },
+    Mutation: {},
+};
--- a/packages/osc-api-admin/src/resolvers/resolver.ts
+++ b/packages/osc-api-admin/src/resolvers/resolver.ts
@@ -1,14 +1,9 @@
 import type { getUnassignedStudentsArgs } from '~/types/arguments';
-import type { AdminContext } from '~/types/interfaces';
 import * as admin from '~/utils/admin';
 
 export const resolvers = {
     Query: {
-        unassignedStudents: async (
-            _: undefined,
-            args: getUnassignedStudentsArgs,
-            context: AdminContext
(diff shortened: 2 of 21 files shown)


packages/osc-api-admin/src/index.ts
```diff
@@ -0,0 +1,21 @@
+import { startStandaloneServer } from '@apollo/server/standalone';
+import * as dotenv from 'dotenv';
+import { server } from '~/server';
+dotenv.config();
+
+async function startServer() {
+    const { url } = await startStandaloneServer(server(), {
+        context: async ({ req }) => {
+            // @ts-ignore - Type 'string[]' is not assignable to type 'string'. user will never be an array (comes from gateway)
```
mattwendzina, line 9: Not sure if you could use type casting here if you know a user never comes back as an array?
e.g. 
```
const isUser = req.headers?.user as string;
const user = isUser ? JSON.parse(isUser) : null;
```

packages/osc-api-admin/src/resolvers/resolver.ts
```diff
@@ -0,0 +1,16 @@
+import type { getUnassignedStudentsArgs } from '~/types/arguments';
+import type { AdminContext } from '~/types/interfaces';
+import * as admin from '~/utils/admin';
+
+export const resolvers = {
+    Query: {
+        unassignedStudents: async (
+            _: undefined,
+            args: getUnassignedStudentsArgs,
+            context: AdminContext
```
mattwendzina, line 10: Is there a reason you're pulling out the context but not using it?

packages/osc-api-admin/src/schemas/schema.ts
```diff
@@ -0,0 +1,40 @@
+import { gql } from 'apollo-server-core';
+
+export const typeDefs = gql`
+    extend schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key", "@shareable"])
+
+    type Student {
+        studentId: Int
```
mattwendzina, line 7: Should the Ids by non-nullable?

mattwendzina:
Looks good from what I can understand. Couple of things:

- Should we have any tests written for our queries?
- I've tried running the server locally and querying unassigned students - I get back a 'Not authorised' message. Do I need to add any env-vars to run locally, and also is there a test DB that I would need to point to?

packages/osc-api-admin/src/index.ts
Tom-Pearce: Thanks, been trying to fix that for ages. Was trying to tell TS that it will always come as a string, but it never liked it. Guess this still isn't quite doing that but at least it removes the error!

packages/osc-api-admin/src/resolvers/resolver.ts
Tom-Pearce: Nope, probably just left it in there from copying previous setup! Removed now

packages/osc-api-admin/src/schemas/schema.ts
Tom-Pearce: Yes, they should be. Switching between GQL/Prisma I get it confused sometimes, because they use opposite conventions.

Tom-Pearce:
> Looks good from what I can understand. Couple of things:
> 
> * Should we have any tests written for our queries?
> * I've tried running the server locally and querying unassigned students - I get back a 'Not authorised' message. Do I need to add any env-vars to run locally, and also is there a test DB that I would need to point to?

Yes, I still need to look into the testing side of it all. Hopefully I can do it soon.

Yeah, so theres a bit of initial setup for the API. Additionally you SHOULD be getting unauthorised because you likely haven'y called the login API first to get a token, and then passing that token in the Authorization header. You also won't be able to create a user without being logged in first so would need to disable the shield rules on auth API, create a user, then you can login. So that error is expected behaviour.

If you want I'm happy to spend 15 minutes going through it with you. I did it with Steven recently and it's mainly just giving you my .env and creating you a user on the DB. Then you should have a working copy of the API.

packages/osc-api-admin/src/schemas/schema.ts
mattwendzina: Do the courseId/tutorId under `type CourseAllocations` also need to be non-nullable?

Tom-Pearce: `CourseId` yes, tutor no, because they may be unassigned. In the case of this query there will always be 1 course without a tutor as it's specifically pulling students with an course with no tutor.

mattwendzina:
> > Looks good from what I can understand. Couple of things:
> > 
> > * Should we have any tests written for our queries?
> > * I've tried running the server locally and querying unassigned students - I get back a 'Not authorised' message. Do I need to add any env-vars to run locally, and also is there a test DB that I would need to point to?
> 
> Yes, I still need to look into the testing side of it all. Hopefully I can do it soon.
> 
> Yeah, so theres a bit of initial setup for the API. Additionally you SHOULD be getting unauthorised because you likely haven'y called the login API first to get a token, and then passing that token in the Authorization header. You also won't be able to create a user without being logged in first so would need to disable the shield rules on auth API, create a user, then you can login. So that error is expected behaviour.
> 
> If you want I'm happy to spend 15 minutes going through it with you. I did it with Steven recently and it's mainly just giving you my .env and creating you a user on the DB. Then you should have a working copy of the API.

Cool. Well as long as we're good with testing coming further down the line then LGTM!

PR #686, opened 2023-02-07
Feature/academic hub setup

stevenbister:
## 📝 Description

Sets up initial routes and links up the auth api so we can do a simple login.

N/a

## 🚀 New behavior

We've installed Apollo Client and set it up inside `/lib/apollo/getClient.server.ts` but instead of using the providers and hooks we can use it in this way to run inside Remix's loaders and actions. You can see this being used in the `login.tsx`

We've also created a bunch of initial routes which contain basic placeholder markup to identify them, it's intended to be removed once work starts on them.

## 📝 Additional Information

Not added any tests yet, once the auth flow is finalised and we're returning more user data will setup integration/e2e tests.

Any TODO's are intentional and reminders for upcoming work.

### DIFF
--- a/packages/osc-academic-hub/app/root.tsx
+++ b/packages/osc-academic-hub/app/root.tsx
@@ -17,7 +17,6 @@ import oscUiSwitchStyles from 'osc-ui/dist/src-components-Switch-switch.css';
 import styles from 'osc-ui/dist/src-styles-main.css';
 import React, { useEffect } from 'react';
 import { checkConnectivity } from '~/utils/client/pwa-utils.client';
-import { getUser } from './session.server';
 import { getColorScheme } from './utils/colorScheme';
 
 let isMount = true;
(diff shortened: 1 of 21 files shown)


packages/osc-academic-hub/app/root.tsx
```diff
@@ -117,7 +116,7 @@ export default function App() {
                     isMount: mounted,
```
mattwendzina, line 116: Not sure why I can't leave a comment on the line, but can remove the `React.` from `React.useEffect` on line 109

mattwendzina: Line 110 - Looks like `mounted` can be a const

stevenbister: Ah yeah, I think this is leftover from the initial setup as it's in ecommerce as well from the looks of it. You happy for me to open up a ticket and tackle this and the `React.useEffect` thing in a separate PR?

PR #690, opened 2023-02-10
Feature/assignments api

Tom-Pearce:
## 📝 Description

Authenticates S3 client, and exposes an endpoint to get a signed URL to upload a file to.

None

## 🚀 New behavior

Exposes an endpoint to get a signed URL to upload a file to.

## 📝 Additional Information
We need to keep the AWS package versions in sync otherwise it causes TS errors.


packages/osc-api-assignments/src/server/permissions.ts
```diff
@@ -0,0 +1,12 @@
...
+import { isAuthenticated } from './rules';
+
+export const shieldPermissions = shield(
+    {
+        Query: {
+            uploadUrl: isAuthenticated,
+        },
+        // Mutation: {},
+    },
+    { debug: true }
```
mattwendzina, line 11: Have you intentionally left `debug` in here?

packages/osc-api-assignments/src/resolvers/resolver.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { AssignmentsContext } from '~/types/interfaces';
+import { baseParams, getSignedUrlParams } from '~/utils/s3';
+
+export const resolvers = {
+    Query: {
+        uploadUrl: async (_: undefined, __: undefined, { s3Client }: AssignmentsContext) => {
+            // Generate AWS PUT object command
+            const command = new PutObjectCommand(baseParams);
+            // Request a signed URL for client to upload the file to S3
+            return await getSignedUrl(s3Client, command, getSignedUrlParams);
```
mattwendzina, line 12: Not exactly sure how error handling is dealt with inside of resolvers. Don't think it's somewhere you can wrap things in a try catch. Just thinking whether we'd need any error handling around this `getSignedUrl` for example for when things go awry? Perhaps this gets picked up elsewhere though?

mattwendzina (changes requested):
Added a couple of comments mate.
Same as Admin API - guessing this needs some testing around it (and lms API). Wonder if it's worth getting some tickets created around this if it's not being done now?

packages/osc-api-assignments/src/server/permissions.ts
Tom-Pearce: No, I was having issues and forgot it was there, will remove it.

packages/osc-api-assignments/src/resolvers/resolver.ts
Tom-Pearce: Sorry, we need to `return` an error, not throwing them, but GQLShield will also catch any thrown errors and return them nicely as mentioned

mattwendzina: Cool, sounds good mate 👍

mattwendzina (changes requested):
Looks good mate. Guessing you're leaving testing out then for these atm, like the others and will come back to it once they get fleshed out a bit more? 
Assuming that's the case, happy to give it a 👍

Tom-Pearce:
> Looks good mate. Guessing you're leaving testing out then for these atm, like the others and will come back to it once they get fleshed out a bit more? Assuming that's the case, happy to give it a 👍

Yeah, I need to find some time to at least build some contract testing in for the APIs, and we can do unit/integration later as we extend the functionalities.

mattwendzina:
> > Looks good mate. Guessing you're leaving testing out then for these atm, like the others and will come back to it once they get fleshed out a bit more? Assuming that's the case, happy to give it a 👍
> 
> Yeah, I need to find some time to at least build some contract testing in for the APIs, and we can do unit/integration later as we extend the functionalities.

Sounds good mate.

mattwendzina (approved):
LGTM

hi-im-si:
Nice work @Tom-Pearce 🎉  
Might be worth creating an issue in regard to setting up contract testing as referenced by @mattwendzina to ensure we have this logged for future reference.

PR #691, opened 2023-02-10
Feature/hero component

stevenbister:
- Closes #682 
- Closes #683 
- Closes #684 

## 📝 Description

Adds the `<Hero>` component. I also ended up updating the `o-container` class as well to account for the wider gutters we have in the design. There were a few files/types etc. in there that I'd brought across from my POC when we set all of this up so I've either removed them or updated where necessary.

A few other extras

- There might be a few TS errors these should be fixed by #709 
- `<PolymorphicComponentProps>` I've created a generic for making polymorphic components (`as` props stuff) I've had a few use cases now and I imagine this will pop up again. So hopefully this will make typescript-ing those easier going forward.

## ⛳️ Current behavior (updates)

N/a

## 🚀 New behavior

Adds `<Hero>` component

## 📝 Additional Information

I've left a couple of TODOs in there intentionally, one for adding the flourishes & one for the massive font size

PR #692, opened 2023-02-11
Feature/text gird

stevenbister:
- Closes #

## 📝 Description

Adds `<TextGrid>` component, which displays content in a responsive grid. There is an prop to add an inline heading otherwise the heading will sit "block" above it.

To create the grid I'm using the `auto-fit` keyword to make it responsive without the need for media queries. This will also allow us to omit needing a prop/modifier class to handle swapping the number of columns. As the columns will be dependent on the width of the container. I don't think it's necessary for the CMS users to be able to set a number of columns as looking at the design it's only use that has two columns is in the product page and I foresee the container it sits in being narrower than the default one.

A couple of things to note with regards to the `ecommerce` component.
- TS error on the `<Content />` component inside `<TextGrid />`. I've fixed the cause of this error, with the `Maybe<T>` generic, in #691 so that should get sorted once merged
- The `<Icon />` is working however it might not show anything, this is down to an issue outlined and fixed in #685, so again that will work once that's merged in. 

N/a

## 🚀 New behavior

Adds `<TextGrid>` component.


mattwendzina (approved):
LGTM. I'm guessing those things in the description that have now been crossed out are because they no longer apply/are sorted?

stevenbister:
> LGTM. I'm guessing those things in the description that have now been crossed out are because they no longer apply/are sorted?

Sweet! Thanks man 😎 And yeah they're all sorted now

hi-im-si (approved):

PR #693, opened 2023-02-13
fix: workflow fixes

aaacafe786:
Closes # 

## 📝 Description

fix bugs deploying osc-ecommerce, studio and academic hub

- updated to docker/setup-buildx-action@v2 v0.9.1
- change conditionals for studio branches
- added env variable to studio (sanity_studio_ci) and dockerfiles
- added env.production.sample to studio, please note, you will need an env.production file in studio
- updated nvmrc
- updated env.d.ts in studio
- updated token imports in studio
- added @radix-ui/react-icons as a package to studio

wont build previews/etc.
error finding image when deploying ecommerce/academic hub
error importing tokens locally and remotely in studio
error missing package @radix-ui/react-icons in studio

## 🚀 New behavior
will build previews/etc.
no issues with workflows

no


packages/osc-studio/components/inputs/ColorPicker.tsx
```diff
@@ -5,8 +5,11 @@ import { SearchIcon } from '@sanity/icons';
 import { Autocomplete, Box, Card, Flex, Text } from '@sanity/ui';
 import { uuid } from '@sanity/uuid';
 import React from 'react';
-import colors from '../../../../tokens/colors';
 import { capitalizeFirstLetter } from '../../utils/capitalizeFirstLetter';
+const colors =
+    process.env.SANITY_STUDIO_CI !== 'true'
+        ? require('../../../../tokens/colors')
+        : require('../../tokens/colors');
```
hi-im-si, line 12: is there any way we can avoid these relative paths - they can become extremely messy and problematic in production. could we utilise tsconfig to define some base paths?

packages/osc-studio/schemas/blocks/body.tsx
```diff
@@ -1,6 +1,12 @@
...
-import colors from '../../../../tokens/colors';
-import sizes from '../../../../tokens/fluid-scale';
+const sizes =
+    process.env.SANITY_STUDIO_CI !== 'true'
+        ? require('../../../../tokens/fluid-scale')
+        : require('../../tokens/fluid-scale');
+const colors =
+    process.env.SANITY_STUDIO_CI !== 'true'
+        ? require('../../../../tokens/colors')
+        : require('../../tokens/colors');
```
hi-im-si, line 9: is there any way we can avoid these relative paths - they can become extremely messy and problematic in production. could we utilise tsconfig to define some base paths?

hi-im-si (changes requested):

packages/osc-studio/components/inputs/ColorPicker.tsx
stevenbister: Personally not sure how tsconfig would work (need to do some reading) but perhaps we move the tokens into a package and export them? So it would work something like this?
```tsx
import { colors } from 'osc-tokens'
```
Not sure if that would be missions to do though 🤔

packages/osc-studio/schemas/blocks/body.tsx
aaacafe786: hey, been playing around with this, the issue is that the tokens folder outside of osc-studio so tsconfig does not recognise the file. i think stevens idea to move the tokens into a package is probably the cleanest way to fix this, but in the meantime do you mind if we leave it as is?

PR #706, opened 2023-02-15
fix: add SpriteSheetProvider and re-write test

### LINKED ISSUE #705: Fix Failing TextArea test
### Link task to story

Story: 

### Description

TextArea is failing a test due to new Icons and Spritesheet.

Test needs updating in order to fix this.

mattwendzina:
Closes #705 

## 📝 Description

Fix failing TextArea test

Fails on Icon test

## 🚀 New behavior

Passes Icon test

### DIFF
--- a/packages/osc-ui/src/components/TextArea/TextArea.spec.tsx
+++ b/packages/osc-ui/src/components/TextArea/TextArea.spec.tsx
@@ -1,7 +1,7 @@
 import { render, screen } from '@testing-library/react';
 import React from 'react';
+import { Icon, SpritesheetProvider } from '../Icon/Icon';
 import { TextArea } from './TextArea';
-import { ExclamationTriangleIcon } from '@radix-ui/react-icons';
 
 test('should render a textarea component with a label', () => {
     render(<TextArea id="enquiry" name="Enquiry" />);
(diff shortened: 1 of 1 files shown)


packages/osc-ui/src/components/TextArea/TextArea.spec.tsx
```diff
@@ -18,27 +18,39 @@ test('should render an asterisk when input field is required', () => {
...
+                name="Enquiry"
+                required={true}
+                wasSubmitted={true}
+            />
+        </SpritesheetProvider>
     );
-    expect(screen.getByText('Exclamation Triangle Icon')).toBeInTheDocument();
+
+    expect(document.querySelector('use')).toBeInTheDocument();
+    expect(document.querySelector('use')).toHaveAttribute('href');
```
stevenbister, line 37: Rather than testing the existence of just the href, I would test that the href value is what you'd expect. Something like this:
```suggestion
    expect(document.querySelector('use')).toHaveAttribute('href', ./spritesheet.svg#exclamation-mark'');
```
Basically just asserting that the icon is what you'd expect

stevenbister (changes requested):
Probably a bit of a nitpick but added a comment to the icon test. Otherwise looks good man!

packages/osc-ui/src/components/TextArea/TextArea.spec.tsx
mattwendzina: Ah nice. I did actually try doing this but was getting some weird errors, didn't realise they needed to be comma separated. Updated 👍

stevenbister: Nice! Yeah that comma is a bit of a gotcha 😅

stevenbister (approved):

PR #710, opened 2023-02-17
Bugfix/content unknown attributes

stevenbister:
Originally part of https://github.com/Open-Study-College/osc/pull/691 but pulled it out as it impacts more than just that PR

## 📝 Description

I noticed that some of the props were getting applied as html attributes, which was causing some invalid markup
image

This was down to the spreading of the `other` prop, which after reviewing didn't appear to be needed, so I've removed it.

Props were getting spread through the components `other` prop, and causing them to be applied as html attributes.

## 🚀 New behavior

Removes `other` prop to prevent this from happening.

### DIFF
--- a/packages/osc-ui/src/components/Content/Content.tsx
+++ b/packages/osc-ui/src/components/Content/Content.tsx
@@ -142,7 +142,6 @@ export const Content = (props: Props) => {
         paddingBottom,
         value,
         buttons,
-        ...other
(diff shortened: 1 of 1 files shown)


aaacafe786:
LGTM

aaacafe786 (approved):

hi-im-si (approved):

Issue #253, opened 2022-10-06
Content Media: Make dynamic with Sanity


hi-im-si:
### Link task to story

Story: #252

### Description

n/a

PR #715, opened 2023-02-20
Feature/content media

stevenbister:
Closes #251 

## 📝 Description

Adds the content media component to osc-ui. The component is composed of `<ContentMedia>` and `<ContentMediaBlock>` subcomponent.

Most of the positioning is controlled by the `<ContentMediaBlock>` through the use of `cols` and `align` as well as applying classes directly when needed. The variant prop will also impact the styling slightly in some cases.

A couple of things to be aware of:
1. The padding on the form media story is a bit out, this is because I'm trying to use a utility class but there's a bug preventing it from working. See: #714 
2. The button variant isn't getting passed through to the `<Content />` component correctly so the buttons won't match the design exactly. I fix this in #691 so when that get's merged it will filter through.

N/a

## 🚀 New behavior

Adds the content media component to osc-ui

## 📝 Additional Information

Sanity ticket #253 will be handled in a separate PR


packages/osc-ui/src/components/Carousel/Carousel.tsx
```diff
@@ -120,22 +118,7 @@ export const Carousel = (props: Props) => {
...
-                },
-            },
-            [`(min-width: ${rem(mq['desk-lrg'])}rem)`]: {
-                slides: {
-                    origin: slideOrigin,
-                    perView: 3,
-                    spacing: gap,
-                },
-            },
-        },
```
aaacafe786, line 138: did you mean to remove this?

aaacafe786 (approved):
LGTM, no need to test alignment and positioning, so tests can remain simple

packages/osc-ui/src/components/Carousel/Carousel.tsx
stevenbister: Yes I did. I realised that having the default breakpoints was taking away from the editing experience, so now it should be one slide per page unless the breakpoints are added

Issue #184, opened 2022-09-23
Make dynamic with sanity


EnaamAlum:
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/183

### Description

n/a

PR #716, opened 2023-02-21
Feature/footer component

### LINKED ISSUE #185: Build footer component
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/183

### Description

E2E
Integration

Storybook series

stevenbister:
Closes #185 

## 📝 Description

Adds the footer component to osc-ui and storybook.

The `FooterMenu` subcomponents can all be toggled into an accordion by passing the `isAccordion` prop, I'm using the `StrictUnion` here to allow us to pass the `Accordion` props through once this prop is applied.

To swap the menus over to an accordion on smaller screens I'm having to do this in Storybook (and will have to in ecommerce etc.) and swap everything at once. I found swapping them with `useMediaQuery` in the individual components caused `<Accordion>` to error on the first screen resize. Probably because each one would render at a slightly different time so the underlying `useContext` wouldn't always find the provider. I'm happy with this approach, just thought I'd highlight it and gives us the flexibility to not use accordions if we decide to as a nice extra :)

N/a

## 🚀 New behavior

Adds footer component to osc-ui

## 📝 Additional Information
Sanity ticket #184 will be handled in a separate PR


aaacafe786 (approved):
LGTM, very clean

1) add abit more info to storybook about the different combinations of props each component takes (eg,
``` <FooterMenu isAccordion collapsible defaultValue="0" type="single">
```
)

hi-im-si (changes requested):

PR #746, opened 2023-02-24
Form Component Validation

### LINKED ISSUE #745: Server and Client Validation for Forms
### Link task to story

Story: #723

### Description

Create server and client side validation using Zod schema.

mattwendzina:
Closes #745 

## 📝 Description

Add client and server validation to the form components. Each form component has been given the following new props:
- Errors: string[]
- SetErrors: Dispatch<SetStateAction<any>>  (Shouldn't be any, need to update this)
- Schema- ZodSchema

1) Errors  - Will be a list of errors set in state in a parent, most likely server errors. 
2) SetErrors - The corresponding dispatch for Errors which will allow for errors to be updated on the client
3) Schema- Used for validation - Will be using ZodSchema, but this could be replaced for a different schema if required. This will be the same schema used for both server and client validation to keeps things simple and consistent.

The form inputs only had basic validation in order to work on the styling for error handling 

## 🚀 New behavior

See above

## 📝 Additional Information
There are a few additional things to note as I did some cleanup/refactoring as I went along:

- TextInput:
1. Id and name weren't set to required, but need to be as they won't function correctly without them.
2. Use of Icons (for errors and otherwise has been tidied up - was still using a radix icon)
3. Added <VisuallyHidden/> for Button Icon as this was an accessibility violation (there was no text)
4. Changed text colour on Quaternary style to `neutral-700` as it was a contrast violation

- TextArea:
1. Now imports the <InputError/> from TextInput to reduce code

- Switch:
1. I've added option for a validation error message - Looked at some examples and saw that there could be issues where for example a switch is used to set a preference on the server - If this fails the should be reset and helpful to give the user a message. Perhaps we don't forsee this usage, but as I was adding in all the error validation anyway I thought I'd just update it. Can remove though.
2. Added the option to add a name to the switch, as well as a required description along with <VisuallyHidden/> component as this was a accessibility violation.


packages/osc-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -49,27 +57,34 @@ export const Checkbox = forwardRef<ElementRef<typeof CheckboxPrimitive.Root>, Ch
...
+        }, [checked]);
 
         const modifiers = useModifier('c-checkbox__container', variants);
         const checkboxClasses = classNames('c-checkbox__container', modifiers);
 
         return (
             <fieldset
                 className={
-                    displayError
+                    errors
```
stevenbister, line 87: I think it might be worth tweaking this to `errors.length > 0` as if you get an empty array returned then this will still evaluate as true and you'll get some potentially unexpected results.

packages/osc-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -94,9 +109,9 @@ export const Checkbox = forwardRef<ElementRef<typeof CheckboxPrimitive.Root>, Ch
                     </CheckboxPrimitive.Indicator>
                 </CheckboxPrimitive.Root>
                 <Label name={value} htmlFor={id} required={required} />
-                {displayError ? (
+                {errors ? (
```
stevenbister, line 112: See previous comment about length

packages/osc-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -94,9 +109,9 @@ export const Checkbox = forwardRef<ElementRef<typeof CheckboxPrimitive.Root>, Ch
...
                     <div className="c-checkbox__error-message" role="alert">
-                        {errorMessage}
+                        {errors}
```
stevenbister, line 114: If you have more than one item in the array they get placed together without any space e.g.
image
Might be worth mapping over the array and wrapping each one in a `span` or something with a bit of padding or something to give it some space.
I reckon `<span className="u-pr-2xs">` would work nicely

packages/osc-ui/src/utils/clientSideValidation.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Validates a form input on the client using a given schema, and sets any errors
+ *
+ * @param id A string id of the field to be validated
+ * @param schema A validation schema, e.g. Zod
+ * @param setErrors A dispatch to set any errors
+ * @param value The value of the field to be validated
+ * @returns null
+ */
+export const clientSideValidation = (
```
stevenbister, line 12: Think it's worth adding some tests for this? Or expanding the component tests to include multiple messages or to assert that the validation message clears when the input meets the criteria?

packages/osc-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -49,27 +57,34 @@ export const Checkbox = forwardRef<ElementRef<typeof CheckboxPrimitive.Root>, Ch
...
 
-        const errorMessage = getFieldError(checked, required);
-        const displayError = wasSubmitted && errorMessage;
+        useEffect(() => {
+            // Client side error handling - Sets any errors on an input in
+            // accordance with the schema validation
+            if (errors) {
+                clientSideValidation(id, schema, setErrors, checked);
+            }
+        }, [checked]);
```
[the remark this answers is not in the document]
stevenbister: Think this is in all of them actually

stevenbister (changes requested):
Looking good! Couple of questions/thoughts. I've added most comments to the checkbox but they apply to all of the components.

Also agree with the switch changes you've made, makes sense to reset it if it fails

packages/osc-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,9 +1,10 @@
 import * as CheckboxPrimitive from '@radix-ui/react-checkbox';
-import type { ComponentPropsWithRef, ElementRef } from 'react';
-import React, { forwardRef, useState } from 'react';
+import type { ComponentPropsWithRef, Dispatch, ElementRef, SetStateAction } from 'react';
+import React, { forwardRef, useEffect, useState } from 'react';
+import type { ZodSchema } from 'zod';
```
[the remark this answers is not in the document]
mattwendzina: Yeah, makes sense, thanks for the link. Updated with Zod in the `osc-ui` package. Guess it should be added to `osc-ecommerce` when it's being explicitly used there.

packages/osc-ui/src/components/Checkbox/Checkbox.tsx
mattwendzina: Good point. I've added this into all the components. Also added a check on the if block to ensure there is also a schema and a dispatch before running the `clientSideValidation` function

packages/osc-ui/src/components/Checkbox/Checkbox.tsx
mattwendzina: See above

packages/osc-ui/src/components/Checkbox/Checkbox.tsx
mattwendzina: Yeah, this works well thanks. I've added an example onto the TextInput validation, but added it to all the components.
Also realised a refactor was needed on Switch component - all props were getting spread onto the DOM elements, so have updated so only the rest get passed in.

packages/osc-ui/src/utils/clientSideValidation.ts
mattwendzina: Yep, good idea. I've added some more into the TextInput tests. 

I'm not sure whether to add these across the suite of form inputs or not... 

If something breaks in the function it should get picked up by the TextInput tests and it means potentially only needing to fix the tests in TextInput rather than having to go through all of them (assuming a fix is required in the actual tests). But then maybe it's worth just having them in all of them for some edge cases that I may not be thinking about?

What do you think?

packages/osc-ui/src/components/Select/Select.spec.tsx
```diff
@@ -38,3 +39,20 @@ test('should render a placeholder in the select component', () => {
     expect(screen.getByText('Please Select')).toHaveTextContent('Please Select');
     expect(container.querySelector(`button[data-placeholder]`)).toBeInTheDocument();
 });
+test('should render an error message when errors are present', () => {
+    const setErrorsMock = vi.fn();
+    render(
+        <SpritesheetProvider>
```
stevenbister, line 45: If you import the render function from `test-utils` you can drop the spritesheet provider off these tests
https://github.com/Open-Study-College/osc/wiki/OSC-UI#testing 

Although you might need to rebase this branch first 😅

packages/osc-ui/src/components/DatePicker/DatePicker/DatePicker.tsx
```diff
@@ -53,6 +80,11 @@ export const DatePicker = (props: DatePickerProps) => {
                     </ReactAriaDialog>
                 </ReactAriaPopover>
             )}
+            {errors ? (
```
stevenbister, line 83: Do we need to check the length of the array here and loop the output on line 85?

packages/osc-ui/src/components/DatePicker/DatePicker/DatePicker.tsx
```diff
@@ -34,10 +57,14 @@ export const DatePicker = (props: DatePickerProps) => {
                 // Manually setting random ID due to bug on duplicate Ids:
                 // https://github.com/adobe/react-spectrum/issues/3969
                 id={dateFieldId}
+                aria-invalid={props.errors ? true : false}
```
stevenbister, line 60: Is there a reason for not using the destructured value here? Also; might need to check against the length of the array

stevenbister (changes requested):
Sorry spotted a few other things, also the test for the datepicker is failing on my local run, not quite sure what's changed? Looks like it's a `div` that's holding that year now
```
❯ Object.getElementError ../../node_modules/@testing-library/dom/dist/config.js:40:19
 ❯ ../../node_modules/@testing-library/dom/dist/query-helpers.js:90:38
 ❯ ../../node_modules/@testing-library/dom/dist/query-helpers.js:62:17
 ❯ ../../node_modules/@testing-library/dom/dist/query-helpers.js:111:19
 ❯ src/components/DatePicker/DatePicker/DatePicker.spec.tsx:97:30
     95|     await user.click(decButton);
     96|
     97|     const yearButton = screen.getByRole('button', { name: '2023' });
       |                              ^
     98|     await user.click(yearButton);
     99|

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯

Test Files  1 failed | 37 passed (38)
     Tests  1 failed | 175 passed (176)
      Time  9.53s (in thread 12.54s, 75.99%)
```

packages/osc-ui/src/utils/clientSideValidation.ts
mattwendzina: Yeah, I think it would be okay to just use the TextInput as representative of the others 👍

mattwendzina:
> Sorry spotted a few other things, also the test for the datepicker is failing on my local run, not quite sure what's changed? Looks like it's a `div` that's holding that year now
> 
> ```
> ❯ Object.getElementError ../../node_modules/@testing-library/dom/dist/config.js:40:19
>  ❯ ../../node_modules/@testing-library/dom/dist/query-helpers.js:90:38
>  ❯ ../../node_modules/@testing-library/dom/dist/query-helpers.js:62:17
>  ❯ ../../node_modules/@testing-library/dom/dist/query-helpers.js:111:19
>  ❯ src/components/DatePicker/DatePicker/DatePicker.spec.tsx:97:30
>      95|     await user.click(decButton);
>      96|
>      97|     const yearButton = screen.getByRole('button', { name: '2023' });
>        |                              ^
>      98|     await user.click(yearButton);
>      99|
> 
> ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
> 
> Test Files  1 failed | 37 passed (38)
>      Tests  1 failed | 175 passed (176)
>       Time  9.53s (in thread 12.54s, 75.99%)
> ```

Thanks mate, going through these now. DatePicker tests sorted (there was a failure on Range too). I had set `shouldCloseOnSelect` to `true` as needed this for the forms. It was hardcoded and meant to go back and make it a prop and forgot lol. Have done that now and updated the tests with this 👍

packages/osc-ui/src/utils/clientSideValidation.ts
stevenbister: Sounds good to me

packages/osc-ui/src/components/Select/Select.spec.tsx
mattwendzina: All done. Updated this on the other form components too.

packages/osc-ui/src/components/DatePicker/DatePicker/DatePicker.tsx
mattwendzina: Yeah, good shout. Done.

packages/osc-ui/src/components/DatePicker/DatePicker/DatePicker.tsx
mattwendzina: Yeah, think I just missed this. Destructured now 👍

packages/osc-ui/src/components/DatePicker/DatePicker/DatePicker.tsx
```diff
@@ -34,10 +57,14 @@ export const DatePicker = (props: DatePickerProps) => {
...
             >
                 <DateField
                     {...fieldProps}
+                    dateFieldId={dateFieldId}
                     buttonProps={buttonProps}
+                    errors={errors}
+                    aria-describedby={props.errors ? `${dateFieldId}-error` : undefined}
```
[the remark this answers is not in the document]
mattwendzina: Sorted as above

packages/osc-ui/src/components/DatePicker/DateRangePicker/DateRangePicker.tsx
```diff
@@ -203,29 +244,45 @@ export const DateRangePicker = (props: DateRangePickerProps) => {
...
         <div className="c-datepicker c-datepicker__range">
             <label className="c-label" {...labelProps}>
                 {props.label}
             </label>
             <div {...groupProps} ref={ref}>
-                <div className="c-datepicker__date-field-container">
+                <div
+                    className="c-datepicker__date-field-container"
+                    id={dateFieldId}
+                    aria-invalid={errors ? true : false}
```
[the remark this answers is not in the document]
mattwendzina: Yep, done

packages/osc-ui/src/components/DatePicker/DateRangePicker/DateRangePicker.tsx
```diff
@@ -203,29 +244,45 @@ export const DateRangePicker = (props: DateRangePickerProps) => {
...
+                >
                     <div className="c-datepicker__date-field-inner-container">
-                        <DateField buttonProps={buttonProps} {...startFieldProps} />
+                        <DateField
+                            aria-describedby={errors ? `${dateFieldId}-error` : undefined}
```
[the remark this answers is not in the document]
mattwendzina: 👍

packages/osc-ui/src/components/DatePicker/DateRangePicker/DateRangePicker.tsx
```diff
@@ -250,6 +307,11 @@ export const DateRangePicker = (props: DateRangePickerProps) => {
                     </ReactAriaDialog>
                 </ReactAriaPopover>
             )}
+            {errors ? (
```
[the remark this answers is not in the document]
mattwendzina: 👍

packages/osc-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -73,45 +99,18 @@ export const TextInput = forwardRef<HTMLInputElement, Props>((props: Props, forw
...
+                    errors && errors.length > 0
                         ? `${containerClasses} c-input__container--error`
                         : `${containerClasses}`
                 }
             >
                 <input
-                    aria-invalid={displayError ? true : false}
-                    aria-describedby={displayError ? `${id}-error` : undefined}
+                    aria-invalid={errors ? true : false}
+                    aria-describedby={errors ? `${id}-error` : undefined}
```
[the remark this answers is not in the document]
mattwendzina: 👍

packages/osc-ui/src/components/Switch/Switch.tsx
```diff
@@ -35,53 +64,75 @@ export interface Props extends SwitchProps {
...
-                        />
-                    ) : null}
-                </span>
-            </SwitchPrimitive.Thumb>
-        </SwitchPrimitive.Root>
-    );
-});
+        return (
+            <>
+                <div
```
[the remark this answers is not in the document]
mattwendzina: Ah, good spot. Yep, so added data-state to the parent. I'm just pointing to the value of the toggle here though - so just true/false - updated the scss to reflect this and all is working.

PR #755, opened 2023-02-28
Feature/react testing library custom renderer utility

stevenbister:
## 📝 Description

I discovered recently that you can create a wrapper around rtl's render function which allows you to add providers and other components around the component you want to render. https://testing-library.com/docs/react-testing-library/setup/#custom-render

I thought it would be a good idea for us to adopt this as we are starting to need a bit of boilerplate around our tests with things like the `<MemoryRouter>` for testing components that contain Remix's `<Link>` and the `<SpriteSheetProvider>` provider for when we have icons.

Currently when writing a test you will import the render function from react testing library and wrap your test with `<MemoryRouter>` and `<SpritesheetProider>` which may look something like this:

```tsx
// ... other imports
import { render, screen, waitFor } from '@testing-library/react';
import { SpritesheetProvider } from '../Icon/Icon';
import { MemoryRouter } from 'react-router-dom';

test('', () => {
    render (
        <MemoryRouter>
            <SpritesheetProvider>
                 <MyComponent />
            </SpritesheetProvider>
        </MemoryRouter>
    )
})
```

## 🚀 New behavior

Now rather than bringing in all of that boilerplate you can do this:

```tsx
// ... other imports
import { render } from 'test-utils';

test('', () => {
    render ( <MyComponent />)
})
```

Rather than importing the render function from `@testing-library/react` now you will import it from `test-utils` like this
```tsx
import { render } from 'test-utils';
```

## 📝 Additional Information

I've updated the tsconfig and the vite config so the path to the utility file is just `test-utils`

I haven't updated the tests for the Icon as I would want to test the `SpritesheetProvider` explicitly there.

### DIFF
--- a/packages/osc-ui/__test__/test-utils.tsx
+++ b/packages/osc-ui/__test__/test-utils.tsx
@@ -1,0 +1,20 @@
+import type { RenderOptions } from '@testing-library/react';
+import { render } from '@testing-library/react';
+import type { ReactElement } from 'react';
+import React from 'react';
+import { MemoryRouter } from 'react-router-dom';
+import { SpritesheetProvider } from '../src/components/Icon/Icon';
+
+const AllTheProviders = ({ children }: { children: React.ReactNode }) => {
+    return (
+        <MemoryRouter>
+            <SpritesheetProvider>{children}</SpritesheetProvider>
+        </MemoryRouter>
+    );
+};
+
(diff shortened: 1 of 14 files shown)


hi-im-si:
This looks a really useful addition and certainly simplifies the structure of the tests. No brainer for me 👍🏼

hi-im-si (approved):

mattwendzina:
Agree, this looks like a winner 👌

mattwendzina (approved):
LGTM

Issue #712, opened 2023-02-19
Deleting package-lock.json file causes remix apps to error with invalid hooks error


stevenbister:
### Description

When I remove the node_modules and package-lock file to do a fresh install, trying to run a remix apps causes them to fail with Invalid hook call error

### Link to Reproduction

http://localhost:3000

### Steps to reproduce

1. In your terminal run `rm -rf node_modules package-lock.json`
2. Run `npm run setup` (or `npm i`)
3. Run `npm run dev:osc-academic-hub`

You will be greeted with
<img width="1171" alt="image" src="https://user-images.githubusercontent.com/23461173/219942924-cc6c5c2c-4249-497a-ae1b-1b0405dd4ae1.png">

### Release Version

0.42.0

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

If you `git restore package-lock.json` it and run the install again it stops the error.

I've tried removing `osc-studio` from the monorepo to test as that's the only package that is obviously using a different react version (17) to the remix apps (18). This didn't make a difference however :/ I guess because it's not a "workspace" that makes sense as it shouldn't make it into the root lock file.

I am pretty stumped by this. Removing all packages except for academic hub still causes the app to error, which makes me think there is something deeper going on and for whatever reason the packages aren't being hoisted or linked correctly. I think using a combination of lerna and npm workspaces _might_ be contributing but that is purely speculation.

As an aside I did also notice that `osc-ui` itself doesn't have React as a dependency so must be relying on it from another package, this is a cross package dependency which is an antipattern for monorepos ([see packages have strong boundaries](https://monorepo.guide/thinking-in-monorepos) so potentially another contributor.

stevenbister:
Attempting to move the workspaces into the lerna.json then causes husky to fail to install
```
npm i

> osc@0.42.0 prepare
> husky install

sh: husky: command not found
npm ERR! code 127
npm ERR! path /Users/stevebister/SITES/osc
npm ERR! command failed
npm ERR! command sh -c husky install

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/stevebister/.npm/_logs/2023-02-19T11_46_18_855Z-debug-0.log
```


Whyyyyyyyyy!

stevenbister:
Actually this causes a lot of things to fail when running commands such as `npm run dev:osc-academic-hub`

```
npm run dev:osc-academic-hub

> osc@0.42.0 dev:osc-academic-hub
[...]

Issue #703, opened 2023-02-14
Create a package for the tokens folder


aaacafe786:
### Description

move the tokens into a package and export them, so it would work something like this: 
```import { colors } from 'osc-tokens'```

### Problem Statement/Justification

we must specify two seperate paths when importing tokens into osc-studio, we also need to use relative paths to account for the fact we move the package to the root

### Proposed Solution or API

- clone osc-ui and rename to osc-tokens, 
- import osc-tokens inside osc-studio eg:
``` "osc-tokens": "file:packages/osc-tokens", ```
- edit workflows so osc-tokens is not deleted

### Alternatives

we could include the tokens folder as part of osc-ui

### Additional Information

this might not work until we upgrade sanity to the next version, as currently it is a standalone project in the monorepo

Issue #711, opened 2023-02-17
Improve caching efficiency when building packages locally and remotely


aaacafe786:
### Description

alot of the commands in the root package.json have the following flag "--skip-nx-cache". this is deffo not required locally and may only be required for some packages (osc-ui) during the workflow build steps. we need to figure out what packages can be cached

this will likely be accomplished by testing each command locally and remotely without the flags in order to determine a pattern (quite a long process)

### Problem Statement/Justification

caching might be able to be improved/applied to more steps

### Proposed Solution or API

test each command locally and as part of the workflows without the "--skip-nx-cache" flag, do not forget to test multiple times as caching only occurs when the build has not changed

aaacafe786:
we can also do abit of reading into the nx.json setup file and try to improve it there

PR #704, opened 2023-02-14
fix: moved tokens to osc-ui

aaacafe786:
Closes # 

in an attempt to update the import paths for the tokens in osc-studio, i moved the tokens to osc-ui, this can be cloned/stripped and renamed as osc-tokens (if necessary).
please note, the CI build process remains the same, we do need the package to be part of the monorepo in order to import a package

i got this to work with node v18.14.1 and osc-studio as a package by adding the following code to the root package.json
  ```
"overrides": {
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "node-sass": "^8"
  },
```

you will notice i have removed react/react-dom and associated types as a dependency from osc-ui, this is because it interferes with the ci build process, the react version used by osc-ui components is the one in the root node modules, you can confirm this by hovering over the 'react' part of the import script

> Add a brief description

import paths for tokens are very confusing

> Please describe the current behavior that you are modifying

moved the tokens folder into osc-ui

> Please describe the behavior or changes this PR adds

## 💣 Is this a breaking change (Yes/No):

YES, you need to move your .env file to ecommerce and academic hub

PR #760, opened 2023-03-05
Refactor/monorepo - isolate packages

stevenbister:
- Closes https://github.com/Open-Study-College/osc/issues/712
- Closes https://github.com/Open-Study-College/osc/issues/703

## 📝 Description

Due to the amount of iteration in the monorepo over time it was feeling like things were becoming a little flakey and stuff not quite working as you'd expect (see #712). I believe that a factor in this is that our workspaces aren't really isolated from each other, meaning that you could install a package in `osc-ecommerce` and then use it within `osc-ui`, for example. This not only makes the workspaces interdependent but also debugging harder than it has to be.

I've done a lot of reading and learning while doing this and one of the things that gets brought up is making sure to keep packages isolated from each other. [The Packages have strong boundaries](https://monorepo.guide/thinking-in-monorepos) section of this article does a good job of explaining this thinking.

I think that Al has done a great job of getting this set up and this I'm hoping to build upon that in several stages:

- [x] Isolate packages from each other
- [x] Remove Lerna and lean more into NX & NPM workspaces
- [ ] Update GitHub actions to try & consolidate them and speed them up by leaning into using the `nx affected`
- [ ] Deploy sites & run actions without having to move files around (push up entire monorepo and serve from the packages directory for example)
- [ ] Re-structure workspaces into apps for the websites, packages for ui related libraries and services for the apis
- [ ] Update Storybook to V7
- [ ] Remove Rollup from osc-ui so we're just using Vite. Vite uses rollup under the hood so _shouldn't_ be too much of a change

This PR should setup the groundwork for the above steps and makes a start on #711 

- Uses Lerna for scripts
- `osc-studio` is not included in the workspaces of the repo, and therefore acts separately. This was initially done to avoid an issue with react versions as the current Sanity version is a bit behind.
- `tokens` sits outside of workspaces and acts just as a directory rather than a package itself
- Cross dependent packages
- Global config npm packages imported through workspace packages
- Duplicate versions of config packages such as `nx` sitting in remix apps and osc-ui
- Packages installed in both dependencies and dev-dependencies in some cases

## 🚀 New behavior

I've gone through all of the packages, identified what we're using and moved some things around:

- Global setting packages such as `nx`, `eslint`, `prettier` etc. now live in the root `package.json`
- Updated Vite and React Testing Library so packages should all be on the same version
- Removed any unused dependencies
- Removed any duplicate deps that were in both the dependencies and dev-dependencies
- Added any missing extraneous dependencies to the workspace

With regards to Lerna; I decided to remove it and lean more into NX and NPM workspaces. Lerna didn't appear to be adding much in the way we were using it and it was just one more dependency and config to worry about and maintain.

I've updated all of the scripts in the root package.json to use NX and configured the nx.json a little further. More could probably be done here, as you can get quite in depth with it but for now I think it's good.

To get `osc-studio` working as a package I have upgraded the React version in the workspace to 18 and added an `overrides` to the root package.json as @aaacafe786 suggested in here: https://github.com/Open-Study-College/osc/pull/704. Once we upgrade to Sanity V3 we can ditch this.

There was also the issue of duplicate React versions when deleting and re-generating the lock file. After some digging I discovered it was due to Storybook (see my latest comment in #712). Managed to get it working by adding React 18 as a dependency in the root package.json, which forces Storybook to use that as the peer dependency. Not my ideal solution but I think we might be stuck with it until Storybook 7 comes out.

Our design tokens now works as a workspace, nothing has really changed here other than how you import the tokens and the fact we will no longer need to worry about different paths on local compared with prod.

I set up a base ts-config file in the root that all workspaces extend upon, I reckon we could add more to it and reduce the individual config files but felt out of scope and would need input from all of the team. Related to this I've added a typecheck script to all workspaces to make checking typescript a little easier than going into each file or waiting until CI runs.

I've mentioned already about cross dependent packages, to help us avoid that I've added the `import/no-extraneous-dependencies` plugin to our `eslint` config at the root.

Removed the seed and empty db file from the Prisma directory. We're not using them and the seed file doesn't work currently, once we're further along it might be worth adding back in.

Hopefully not too much of a breaking change here, only things to note are

- Packages must be installed into the specific workspace if you want to use it, eslint will error if not.
- Install packages using the `--workspace` flag rather than going into the workspace and installing directly (see the README in this PR for details)
- Lerna has been removed, so any Lerna scripts should be replaced with the NX equivalent 
- Design tokens are now a package so you will now need to install them like this:
  ```tsx
  import { colors } from 'osc-design-tokens'
  ```
- When importing a workspace to another workspace use the `"*"` instead of `"file:..."`
  ```json
   "osc-api": "*"
  ```

## 📝 Additional Information

When you pull this down to test it, it might be worth clearing out all of you're `node_modules` (both the root one and the ones in the workspaces). 

I've added a script called  `npm run clean` which will do this and remove the lock file for you. Just be warned when you run `npm run setup` it will take about 5-10 mins on this initial install as it need to re-generate the lock file. (You should only have to do this as a one off)

I've also made a note of all of our packages and where they're used: https://openstudycollege.sharepoint.com/:x:/s/Dev/Ec1GKgceTB1HkYv3eYJDCUwBd-88ucHWF7SXmUYkMk90XQ?e=VAa9Of 
Might be worth maintaining this to make it easier if we want to remove things?


hi-im-si (commented):
Initial inspection looks pretty good.

The comment on the storybook upgrade to v.7 - we could potentially look to upgrade now - seems like only squashing bugs - I expect an imminent release.

https://discord.com/channels/486522875931656193/1081585833603235931

Might be good for @Open-Study-College/developers to all discuss this and test together as I believe it will be a more robust foundation

stevenbister:
> The comment on the storybook upgrade to v.7 - we could potentially look to upgrade now - seems like only squashing bugs - I expect an imminent release.
> 
> https://discord.com/channels/486522875931656193/1081585833603235931

Oooo I hadn't seen this 👀

nx.json
```diff
@@ -1,29 +1,41 @@
...
+    "defaultBase": "main"
+  },
   "tasksRunnerOptions": {
     "default": {
       "runner": "@nrwl/nx-cloud",
       "options": {
-        "cacheableOperations": ["build", "test"],
+        "cacheableOperations": ["build", "test", "test:e2e"],
+        "parallel": 20,
         "accessToken": "NGY1YTkxMDItN2ViNS00MjgxLWIxOGEtMTdjZjc0Mzk1OGZmfHJlYWQtd3JpdGU="
```
mattwendzina, line 13: Is it okay that we have the access token directly in the code or should this be an env var?

packages/osc-api-admin/package.json
```diff
@@ -5,7 +5,8 @@
   "main": "src/index.ts",
```
mattwendzina, line 5: Sorry, not sure where to put this comment, so just putting at the top of first api package that has changes. 

Feel free to ignore this as it's just a thought with not a huge level of understanding as to how this holds together...

When I look at the packages folder I see a huge number of api packages. I was wondering whether it would make sense for all the api packages to sit in `osc-api` to keep them altogether in one place. Just seems like it would look a bit neater that way. Appreciate this may not be simple or even preferable lol, but just a thought I'd had. 

And with that said it does look like `osc-api` is it's own package anyway. Hard to know exactly what this one is doing given it's just a generic name, but by looks of it, it gets a user, and there is an auth function too. Looks like it gets pulled into all of the other apis, although I can see it only being used in the auth api. 

This is perhaps getting a bit too much into the weeds and something that @Tom-Pearce probably knows more about.  

The main point lol, is whether it may be useful to have all the api packages in their own folder to keep it looking a little neater.

nx.json
stevenbister: Ahh yeah that's a good point, tbh it was like that in the original monorepo so I didn't really think about it but yeah probs worth hiding it

packages/osc-api-admin/package.json
stevenbister: Yep! Defo agree (see "Re-structure workspaces into apps for the websites, packages for ui related libraries and services for the apis" in the PR notes 😉 ) 
But yeah ideally I'd like to structure the repo like this

- apps
    - osc-ecommerce
    - osc-academic-hub
    - osc-studio
- packages
    - osc-ui
    - osc-design-tokens
- services
    - osc-api
    - osc-api-admin
    - etc.

But we need to get the github actions/workflows a bit more maintainable first

package.json
```diff
@@ -63,25 +63,51 @@
...
-    "packages/osc-api-auth",
-    "packages/osc-api-ecommerce",
-    "packages/osc-api-assignments",
-    "packages/osc-api-crm",
-    "packages/osc-api-admin",
-    "packages/osc-api-lms",
-    "packages/osc-api-gateway"
-  ],
+  "dependencies": {
+    "react": "^18.2.0",
```
mattwendzina, line 67: Just to clarify, the only reason we have react at the top level here is because of the issue with storybook? Presumably this generally still sticks to the rule of installing inside of repos and have strong boundaries between packages and won't need to be here when we upgrade to Storybook v7?

packages/osc-api-admin/package.json
mattwendzina: Ah, yeah sweet 🙌

package.json
stevenbister: Yeah that's right, if we need React in a workspace then we should still be installing it in there as well.

I don't _fully_ understand why Storybook requires this 😅 but it seems to be a bit of a pain points for most people. My thoughts are that it's because we have Storybook itself in the root of the monorepo but that's how NX suggest to do it if you want to use it for multiple workspaces like we are.

mattwendzina:
Appears to be a console warning in storybook now:
```
`showName` is deprecated as `name` will stop having dual purposes in the future. Please specify a `title` in `globalTypes` instead. 
```
Not really urgent, but just flagging it. Was able to get rid of it by just adding a title in the globalTypes in `preview.js`

package.json
mattwendzina: Ah right, yeah that kinda makes sense with the fact that Storybook is in the root of the monorepo.

mattwendzina:
When I run `osc-ecommerce` it produces a number of files in the public fonts folder. Not sure if this is meant to behave like this, though I have a feeling that this may be something that is already happen if I remember correctly lol

stevenbister:
> Not really urgent, but just flagging it. Was able to get rid of it by just adding a title in the globalTypes in `preview.js`

Can you open a issue for this one?

stevenbister:
> When I run `osc-ecommerce` it produces a number of files in the public fonts folder. Not sure if this is meant to behave like this, though I have a feeling that this may be something that is already happen if I remember correctly lol

Yeah that's what it's doing currently, we need to move the fonts from osc-ui into Remix's public folder so the paths in the stylesheet match up/work. We _might_ be able to stop doing this once we update the deployments and stop moving thing around in production but I seem to remember this being a Remix requirement 🤔

mattwendzina:
Looks great mate, pulled it down. Followed all the instructions, ran the different apps and everything seemed to be working well! I'll open an issue on the console.warning for storybook. Other than getting that accessToken into an env var then I'm happy to give 👍 and get it in 🚢🚢

mattwendzina:
> > Looks great mate, pulled it down. Followed all the instructions, ran the different apps and everything seemed to be working well! I'll open an issue on the console.warning for storybook. Other than getting that accessToken into an env var then I'm happy to give 👍 and get it in 🚢🚢
> 
> Thanks mate 🎉
> 
> Just had a bit of a read through this https://nx.dev/nx-cloud/account/access-tokens and it looks like it's common to leave the token in the nx.json. What I think would be worth doing is changing it to a read-only token and then having a read-write one for CI. Looks like that would benefit the builds better as our local changes won't impact them while we're working on them.
> 
> Might be one for a separate task, probs makes sense to do it when we update the workflows (#779).
> 
> If you're happy with that?

Yep, sounds like a great option mate 👌

.github/actions/movePackageToRoot/action.yml
```diff
@@ -12,6 +12,3 @@ runs:
 
     - run: mv -v -f packages/${{ inputs.name }}/* .
       shell: bash
-
```
aaacafe786, line 15: why are we getting rid of this, it was an attempt to reduce extra unused files

aaacafe786:
Seems really good, only other small thing is that:

- Typecheck is failing for some packages locally

aaacafe786 (commented):
LTGM, 2 small things

aaacafe786 (approved):
LTGM, 2 small things

.github/actions/movePackageToRoot/action.yml
stevenbister: Because we've moved design tokens into the packages we want to keep it available (same as osc-ui) otherwise the builds were failing. I couldn't get the regex to work to prevent both that _and_ osc-ui from being deleted. 
I figured because we had discussed removing this action eventually anyway taking it out wouldn't hurt, plus it should shave a bit of time off deployment... probably

stevenbister:
> Seems really good, only other small thing is that:
> 
> * Typecheck is failing for some packages locally

Thanks man 😎 Yeah we've got a bunch of ts issues in main at the moment and I think adding this command has helped to highlight them. I'll open an issue and get it tackled separately from this PR

stevenbister:
Issue opened 👍 
https://github.com/Open-Study-College/osc/issues/783

PR #763, opened 2023-03-06
Bugfix/loading dynamic stylesheets

### LINKED ISSUE #662: Stylesheets not loading on first visit when navigating using Remix `<Link>` or `<NavLink>` components
### Description

When navigating using Remix `<Link>` or `<NavLink>` the dynamic stylesheets don't seem to be loading until after you refresh the page.

### Link to Reproduction

http://localhost:2000

### Steps to reproduce

If you're on a cached page you will need to do a hard refresh

https://user-images.githubusercontent.com/23461173/213688969-64d454b5-da8c-466f-b8de-e6b899d17c37.mov

### Release Version

0.26.1

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

I tested it by moving the stylesheets into the root and having them load on all pages and that seems to work so it appears to be related to the [dynamic links](https://github.com/sergiodxa/remix-utils#dynamiclinks) whether that's a bug in the package or to do with our implementation I'll need to investigate.

stevenbister:
- Closes #662 

## 📝 Description

I discovered a bug where if you navigate our site using the Remix `<Link>` or `<NavLink>` component the dynamic styles would not apply to the page unless you manually refreshed it.

I setup a minimal test on codesandbox and noticed that the dynamic styles would apply there as you'd expect. After a bit more poking around; I noticed that in a few examples and in the Indie Stack itself they weren't type guarding the `<Meta>` or `<Links>` components in the `root.tsx` file whereas we were. 

Removing the type guard seems to have solved it. This makes sense as the `<Link>` component adds [client-side routing](https://remix.run/docs/en/1.14.0/components/link), so by saying "render the `<DynamicLink>` if the document is undefined" we're forcing it to only add the `<DynamicLink>` if the server is doing the routing.

An alternative solution would be to only use `<a>` tags however we would then lose out on the benefits of CSR and being able to include things like [prefetch to the links](https://remix.run/docs/en/1.14.0/components/link#prefetch). But open to discussion on this.

- The `<Meta>`, `canonical`, `<Link>` & `<DynamicLink>` are type guarded with `typeof document === 'undefined'`
- `getTypes` function is strictly a server only function and lives in `sanity.server.ts`

## 🚀 New behavior

- Removes type guard from `<Meta>`, `canonical`, `<Link>` & `<DynamicLink>`
- Moves `getTypes` into `Module.tsx` so it's accessible in the client bundle

### DIFF
--- a/packages/osc-academic-hub/app/root.tsx
+++ b/packages/osc-academic-hub/app/root.tsx
@@ -83,8 +83,8 @@ const Document = ({ children }: DocumentProps) => {
     return (
         <html lang="en">
             <head>
-                {typeof document === 'undefined' && <Meta />}
-                {typeof document === 'undefined' && <Links />}
+                <Meta />
+                <Links />
             </head>
             <body>
                 {children}
--- a/packages/osc-ecommerce/app/components/Module.tsx
+++ b/packages/osc-ecommerce/app/components/Module.tsx
@@ -9,7 +9,6 @@ import islandGrid from 'osc-ui/dist/src-components-IslandGrid-island-grid.css';
 import popoverStyles from 'osc-ui/dist/src-components-Popover-popover.css';
 import textGridStyles from 'osc-ui/dist/src-components-TextGrid-text-grid.css';
 import videoStyles from 'osc-ui/dist/src-components-VideoPlayer-video-player.css';
-import { getTypes } from '~/models/sanity.server';
 import type {
     accordionModule,
     cardModule,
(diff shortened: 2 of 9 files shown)


mattwendzina:
Nice spot on this mate. Makes complete sense and having had a look at a few repos too I can't see any examples of where these are only being loaded serverside. 
So LGTM, only thing might be checking with @aaacafe786 as to whether there was a specific reason these had a guard on assuming he'd set up the repo originally with these checks on?

stevenbister:
Thanks @mattwendzina Yeah totally agree @aaacafe786 be good to get your thoughts on this :)

aaacafe786 (approved):
nice catch, looks good to me, this seems to be an artifact from chakra and the css nonsence

PR #769, opened 2023-03-07
Refactor/island grid items positioning

### LINKED ISSUE #765: Card island grid misaligned when there are three items
### Description

When adding three cards to the card island grid then the left item grows beyond the height of the two right items

### Link to Reproduction

http://localhost:2000/kitchen-sink

### Steps to reproduce

> **Note**
> This requires PR#668 to be merged/on that branch to view

1. Go to http://localhost:2000/kitchen-sink
2. Scroll down to the second Looking for some inspiration section with the orange cards

### Release Version

0.51.0

### Browser

Brave 1.48.171

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

image

stevenbister:
- Closes #765 

## 📝 Description

Tweaks the island grid css slightly so the height of the first item will match with the height of the items on the right. 
Currently setting of the max end of the rows as 4 causes the large item to expand past the end of the last item on the right if there are only two.

- Height of the "large" island is handled by `grid-row: 1 / 4`

## 🚀 New behavior

- Adds a `has-{n}-islands` class
- adds a loop to loop through the max islands and sets the value to each `has-{n}-islands` class
- positioning is then handled by these classes

### DIFF
--- a/packages/osc-ui/src/components/Card/card.scss
+++ b/packages/osc-ui/src/components/Card/card.scss
@@ -223,6 +223,7 @@ $card-fs-3xl: $fs-3xl;
             }
 
             #{$self}__footer {
+                align-items: center;
                 border-top: var(--card-border);
                 margin-block-start: $card-space-l;
                 padding-block: $card-space-l;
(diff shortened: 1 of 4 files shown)


hi-im-si (approved):
LGTM 🚀

PR #776, opened 2023-03-08
Refactor/content width

stevenbister:
- Closes #766 

## 📝 Description

As per the ticket I noticed that the content wasn't lining up with the edge of our container. To get around this I've replaced the max-width and the auto margin with just an `80ch` unit and then added a modifier to make the text full width when needed. I've added this as a control to Sanity so it gives the CMS users a bit of choice.

- Content has a margin and max-width pushing it into the centre of the screen, and causing it to not align with the edge of the container

## 🚀 New behavior

- Max width is set based on `ch` unit
- Margin removed
- Full-width modifier added


mattwendzina:
This LGTM, just a question as to why `80ch` is being used as the width when the option isn't full-width. Does it relate to the designs? There wouldn't be times when content would need to be greater than 80ch but not max-width?

stevenbister:
> This LGTM, just a question as to why `80ch` is being used as the width when the option isn't full-width. Does it relate to the designs? There wouldn't be times when content would need to be greater than 80ch but not max-width?

Yeah so I remembered reading somewhere that 70-80ch is a good rule of thumb for content width, so it's not too wide to read comfortably. Looking at the designs we shouldn't go over that as we are containing most of it within components but it's a nice safety net

mattwendzina:
> > This LGTM, just a question as to why `80ch` is being used as the width when the option isn't full-width. Does it relate to the designs? There wouldn't be times when content would need to be greater than 80ch but not max-width?
> 
> Yeah so I remembered reading somewhere that 70-80ch is a good rule of thumb for content width, so it's not too wide to read comfortably. Looking at the designs we shouldn't go over that as we are containing most of it within components but it's a nice safety net

Noice. Sounds good mate 👍

mattwendzina (approved):
LGTM

PR #777, opened 2023-03-09
Bugfix/duplicate stylesheets

### LINKED ISSUE #772: Duplicate stylesheets when loading components
### Description

When loading modules from the CMS we're having to list each stylesheet dependency per module, which is causing duplicate stylesheets to be loaded

### Link to Reproduction

http://localhost:2000/

### Steps to reproduce

1. Go into the Module.tsx file
2. See that Hero (for example) loads hero styles and content styles
3. Visit http://localhost:2000/
4. Inspect the page and you will see that content styles gets loaded twice
This is because there is a content component on the page also loading that stylesheet

image

### Release Version

0.52.0

### Browser

Brave 1.48.171

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

stevenbister:
- Closes #772 

## 📝 Description

When loading the stylesheets for the modules coming out of Sanity we will need to include both the stylesheet for the component and it's dependencies. For example the Hero is made up of `<Hero>`, `<Content>`, `<Carousel>` and `<Button>`. Currently when we add these stylesheets we are only adding them like this
```ts
case 'module.hero':
  styles.push(
    { rel: 'stylesheet', href: heroStyles }
  );
  break;
```
Which means the dependent styles will either be missing or loaded because we are using another module on the page.
To get around this we will do this
```ts
case 'module.hero':
  styles.push(
    { rel: 'stylesheet', href: heroStyles },
    { rel: 'stylesheet', href: carouselStyles },
    { rel: 'stylesheet', href: contentStyles },
    { rel: 'stylesheet', href: buttonStyles }
  );
  break;
```
However this will lead to loading duplicate stylesheets onto the page.

Either missing style dependencies or duplicate stylesheets.

## 🚀 New behavior

Adds a `getUniqueObjects` function which we can use to dedupe the styles array in `getComponentStyles`

### DIFF
--- a/packages/osc-ecommerce/app/components/Module.tsx
+++ b/packages/osc-ecommerce/app/components/Module.tsx
@@ -4,6 +4,7 @@ import oscUiAccordionStyles from 'osc-ui/dist/src-components-Accordion-accordion
 import alertStyles from 'osc-ui/dist/src-components-Alert-alert.css';
 import buttonStyles from 'osc-ui/dist/src-components-Button-button.css';
 import cardStyles from 'osc-ui/dist/src-components-Card-card.css';
+import carouselStyles from 'osc-ui/dist/src-components-Carousel-carousel.css';
 import contentStyles from 'osc-ui/dist/src-components-Content-content.css';
 import contentMediaStyles from 'osc-ui/dist/src-components-ContentMedia-content-media.css';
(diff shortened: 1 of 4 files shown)


mattwendzina (approved):
LGTM

PR #784, opened 2023-03-10
Feature/add icon picker to button

stevenbister:
## 📝 Description

While putting the demo page together I noticed that cms users can't add icons to the buttons so this PR aims to allow that.
This should help with setting up the Cards on the homepage which have icons on the quaternary button

- Buttons can't have icons

## 🚀 New behavior

- Add icon picker to button module in studio
- Update ecommerce `<Button>` and `<Content>` components to take an icon prop and display the icon

### DIFF
--- a/packages/osc-ecommerce/app/components/Button/Button.tsx
+++ b/packages/osc-ecommerce/app/components/Button/Button.tsx
@@ -1,4 +1,4 @@
-import { Button as OSCButton, CopyButton } from 'osc-ui';
+import { Button as OSCButton, CopyButton, Icon } from 'osc-ui';
 import type { buttonModule } from '~/types/sanity';
 
 type Props = {
@@ -21,6 +21,7 @@ export const Button = (props: Props) => {
         textToCopy,
         isInversed,
         variant,
+        icon,
     } = button;
 
     switch (type) {
@@ -36,6 +37,7 @@ export const Button = (props: Props) => {
                     {...rest}
                 >
                     {label}
+                    {icon && <Icon id={icon} />}
                 </OSCButton>
             ) : null;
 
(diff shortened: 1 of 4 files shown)


mattwendzina:
This LGTM. Did a test, tried different variants with the Icon and all good. 

My only slight issue is that there is no error validation for the Buttons inside the Content for the Type. Ie - I created a new button, but didn't set a type and the changes didn't pull through, I then noticed the an error on the button - see screenshot below. Fixed by selecting a type so not a big issue but not a great UX for the marketing team as they won't necessarily know what the problem is. 

My opinions aren't overly strong about it, or about it needing to be addressed now, I'm just flagging it to see what your view is. One option may be that a default is is set, in the same way that there is a default for the variant, but that's probably not ideal given that there isn't probably one standard type they will use. Anyway, be good to see what you think.

stevenbister:
> This LGTM. Did a test, tried different variants with the Icon and all good.
> 
> My only slight issue is that there is no error validation for the Buttons inside the Content for the Type. Ie - I created a new button, but didn't set a type and the changes didn't pull through, I then noticed the an error on the button - see screenshot below. Fixed by selecting a type so not a big issue but not a great UX for the marketing team as they won't necessarily know what the problem is.
> 
> My opinions aren't overly strong about it, or about it needing to be addressed now, I'm just flagging it to see what your view is. One option may be that a default is is set, in the same way that there is a default for the variant, but that's probably not ideal given that there isn't probably one standard type they will use. Anyway, be good to see what you think.

Yeah good shout, should defo validate that bit!

mattwendzina:
Screenshot 2023-03-15 at 10 46 18
Here's that screenshot = D

mattwendzina (approved):
LGTM!

PR #788, opened 2023-03-13
Bugfix/carousel height

stevenbister:
## 📝 Description

Removes the `height: 100%` set on the `c-carousel` introduced in #715

Currently all carousels will take this value and will cause some of them to take up much more height than necessary and causing some strange visual bugs e.g. the hero on the kitchen sink test page:

https://user-images.githubusercontent.com/23461173/224772116-0c401548-723e-4cc4-9d8c-2bd8852f9fcf.mov

## 🚀 New behavior

- Removes `height: 100%` from the carousel so it won't take up the height of it's flex/grid parent.


mattwendzina (approved):
LGTM

PR #794, opened 2023-03-16
Feature/Integrate hydrogen with osc-ecommerce

stevenbister:
## 📝 Description

Have added a initial setup of Hydrogen into osc-ecommerce -- this is just the setup and a few basic routes for us to get started with.

I have had to downgrade our Remix dependencies to `1.12.0` as they are a peer dependency of Hydrogen and `1.13.0` was causing it to fail the install.

As we aren't hosting on Oxygen we are unable to use the default server that ships when you install Oxygen, however we still need a server entry point and can't use the built in remix-server. After a recommendation from Shopify I have added a simple express server which is based off the [remix-express template](https://github.com/remix-run/remix/blob/main/templates/express/server.js).

image

Related to this -- because we aren't hosting on Oxygen we can't make use of the Shopify cli as that spins up an oxygen server which is incompatible with our express server. Having said that I have installed `@shopify/remix-oxygen` as a dev dependency to give us access to some types such as the `storefront` object. But we shouldn't be using it for any production code.

Something to note related to the `server.ts` stuff. Shopify doesn't enforce caching but recommends it. 
> The cache instance must follow the [Cache API interface](https://developer.mozilla.org/en-US/docs/Web/API/Cache). It can store data in-memory or using solutions like Redis. This parameter is optional but recommended.

I've not included it here as I know there was talk about implementing Redis so figured this can happen at the same time.

As we are no longer using the built in remix server we now run our dev process on nodemon and use the Remix cli to watch for file changes.

## 🚀 New behavior

- Adds `@shopify/hydrogen` package
- Integrates Shopify with existing products and collections routes
- Adds new routes
  - cart
    - Not hooked up or anything but the shell is there and ready
  - collections (all categories page)
    - The shell is there and it runs a basic query but we will need to hook it up to Sanity to make is CMS-able
-  Removed admin route
    - Not gonna be used AFAIK

## 📝 Additional Information

Give me a shout when you're reviewing as you'll need these secrets adding to your `.env`

```
PUBLIC_STOREFRONT_API_TOKEN
PRIVATE_STOREFRONT_API_TOKEN
PUBLIC_STOREFRONT_API_VERSION
PUBLIC_STORE_DOMAIN
```


packages/osc-academic-hub/package.json
```diff
@@ -29,10 +29,10 @@
...
     "@prisma/client": "^4.11.0",
     "@radix-ui/react-icons": "^1.2.0",
-    "@remix-run/node": "^1.13.0",
-    "@remix-run/react": "^1.13.0",
-    "@remix-run/serve": "^1.13.0",
-    "@remix-run/server-runtime": "^1.13.0",
+    "@remix-run/node": "^1.12.0",
+    "@remix-run/react": "^1.12.0",
+    "@remix-run/serve": "^1.12.0",
+    "@remix-run/server-runtime": "^1.12.0",
```
aaacafe786, line 35: does this need to be done inside the ecommerce package.json too?

packages/osc-ecommerce/package.json
```diff
@@ -9,11 +9,11 @@
     "build": "run-s build:*",
     "build:css": "PORT=2001 LOAD_PATH=app/styles PATH_TO_MAIN_SCSS=/app/styles/main.scss PATH_TO_DEST_MAIN_CSS=/app/styles/dest/main.css PATH_TO_COMPONENTS=/app/components node bin/postcssBuild.js",
     "build:remix": "cross-env NODE_ENV=production remix build && npm run copy-fonts:from-remix",
-    "dev": "cross-env NODE_ENV=development npm run copy-fonts:from-ui && npm run build && run-p dev:*",
-    "dev:css": "PORT=2001 LOAD_PATH=app/styles PATH_TO_MAIN_SCSS=/app/styles/main.scss PATH_TO_DEST_MAIN_CSS=/app/styles/dest/main.css node bin/postcss.js",
```
aaacafe786, line 13: add this back in pls

packages/osc-ecommerce/remix.config.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
     ignoredRouteFiles: ['**/.*', '**/*.css', '**/*.test.{js,jsx,ts,tsx}'],
     // serverDependenciesToBundle: [/^header.*/],
     watchPaths: ['../osc-ui/dist/*'],
-    devServerPort: 9002,
+    devServerPort: 8002,
```
aaacafe786, line 9: hey the default is 8002, if we run academic hub and ecommerce together there looks to be some problems, might need to have this in there

stevenbister: What problems are you getting, seems alright for me? I needed to set this to `8002` as it was causing a bunch of websocket issues in ecommerce

aaacafe786: Hey, try running e-commerce and academic hub locally on different terminals with devServerPort set to 8002 on both, the screen is blank on one of them if you navigate to the homepage

aaacafe786: @stevenbister maybe keep e-commerce as 8002 and change academic hub to something else, should buy some time to think

aaacafe786 (approved):
ltgm, testing passing and queries resolved

PR #795, opened 2023-03-20
Feature/design tokens vite

stevenbister:
## 📝 Description

Part one of the Sanity V3 updates.

With V3 now using Vite as the compiler it has started throwing an error when we try to import our design tokens
```
Uncaught error: require is not defined
http://localhost:3333/schemas/blocks/body.tsx?t=1679325463566:5:5
ReferenceError: require is not defined
    at http://localhost:3333/schemas/blocks/body.tsx?t=1679325463566:5:5
```
This is because Sanity is an SPA and `require` is not available so we need to provide the tokens as a module.

However because we need to make use of the `json-to-scss` package we need our input files to continue to use the `require` syntax. Therefore I have added Vite to bundle the package into both `es` and `cjs` syntaxes, stored in the `dist` directory.

I believe Vite is treeshakable by default as well so we might gain some perf improvements, albeit pretty minor in this case.

- Design tokens package simply exports the objects using `module.exports`

## 🚀 New behavior

- Adds Vite to bundle the package into `es` and `cjs` files
- Changes default exports to named exports
- Adds `build:osc-design-tokens` script

Should only be temporary but for the time being; if you wish to use tokens inside of Sanity you will need to directly target the `cjs` file like this
```ts
const { colors } = require('osc-design-tokens/dist/index.js');
```
But once V3 is implemented this should no longer be needed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -58866,7 +58866,447 @@
       "version": "1.0.0",
       "license": "ISC",
       "devDependencies": {
-        "json-to-scss": "^1.6.2"
+        "@rollup/plugin-commonjs": "^24.0.1",
+        "json-to-scss": "^1.6.2",
+        "vite": "^4.2.1"
+      }
+    },
+    "packages/osc-design-tokens/node_modules/@esbuild/android-arm": {
+      "version": "0.17.12",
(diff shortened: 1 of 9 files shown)


mattwendzina (commented):
LGTM

mattwendzina (approved):
LGTM!

Issue #729, opened 2023-02-24
Make HubSpot form dynamic with sanity


EnaamAlum:
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/723

### Description

n/a

Issue #728, opened 2023-02-24
Form testing


EnaamAlum:
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/723

### Description

n/a

PR #799, opened 2023-03-21
Feature/contact form in ecommerce hubspot integration

mattwendzina:
Closes https://github.com/Open-Study-College/osc/issues/724
Closes https://github.com/Open-Study-College/osc/issues/728
Closes https://github.com/Open-Study-College/osc/issues/729

## 📝 Description

This builds upon the `feature/form-component-validation-updates` & `contact-form-in-ecommerce` branches by adding the ability to dynamically import the form inputs directly from hubspot.  

N/A

## 🚀 New behavior

- Adds form component validation for server and client side using Zod
- Adds Hubspot integration:
1) Enables forms to be created and will pull in the form input and data
2) Allows these forms to be submitted back to Hubspot
- Adds Sanity integration:
1) Enables Marketing to select a form Hubspot
2) Adds settings to enable the form to be a slide-out form if required

No

## 📝 Additional Information

- Currently the formIds are being hardcoded. Eventually want these to be pulled in dynamically using the Hubspot Forms API. This will likely need a proxy and will be re-visited further down the line.


packages/osc-ecommerce/app/components/Forms/FormContainer.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const FormContainer = (props: FormContainerProps) => {
+    const { children, slideOut = false, slideOutText, variant } = props;
+    const variantModifier = useModifier('c-form__container', variant);
+    const classes = classNames('c-form__container', variantModifier);
+
+    return (
+        <div className={slideOut ? classes : 'c-form__container'}>
+            {slideOut ? (
+                <Button variant="quaternary" className="c-form__container--slide-out-btn">
```
stevenbister, line 31: I think rather than a modifier this would be an element, for example `c-form__slide-out-btn`

packages/osc-ecommerce/app/components/Forms/HubspotForm/HubspotForm.spec.tsx
```diff
@@ -0,0 +1,48 @@
...
+test('should render the Contact form', () => {
+    render(
+        <SpritesheetProvider>
+            <HubspotForm
+                formErrors={[]}
+                formFieldGroups={contactFormData.formFieldGroups}
+                setValidationErrors={() => {}}
+                submitText={textContent.submitText}
+                termsAndConditions={textContent.termsAndConditions}
+                titleAndDescription={textContent.titleAndDescription}
```
stevenbister, line 16: Looks like a few of these props aren't being used anymore?

packages/osc-ecommerce/app/components/Forms/HubspotForm/textContent.ts
stevenbister, line 1: If we update the test file to remove those props will we still need this file?

packages/osc-ecommerce/app/components/Forms/types.ts
stevenbister, line 1: Assuming that these types are pulled from the Hubspot docs? But have a few questions/suggestions
- Worth changing the comments to JSDoc?
- Do we need to include the deprecated ones? If so I think you can mark them as deprecated with JSDoc `@deprecated`
- Worth making the readonly ones readonly with ts? https://www.typescriptlang.org/docs/handbook/utility-types.html#readonlytype

Appreciate we might not be directly hitting these so happy to leave it as is if it's a lot of effort

packages/osc-ecommerce/app/components/Forms/utils.tsx
```diff
@@ -0,0 +1,113 @@
...
+                        id={hubspotFields.name}
+                        inputMode={hubspotFields.fieldType === 'phonenumber' ? 'numeric' : 'text'}
+                        name={hubspotFields.label}
+                        pattern={hubspotFields.fieldType === 'phonenumber' ? '[0-9]*' : undefined}
+                        placeholder={hubspotFields.placeholder}
+                        required={hubspotFields.required}
+                        schema={schema.pick({ [hubspotFields.name]: true })}
+                        setErrors={setValidationErrors}
+                        type={hubspotFields.fieldType}
+                        // variants={data.variants}
```
stevenbister, line 54: Should this be commented out?

packages/osc-ecommerce/app/routes/$slug.tsx
```diff
@@ -47,9 +66,48 @@ export const loader: LoaderFunction = async ({ request, params }) => {
         canonicalUrl,
         isPreview,
         query: isPreview ? PAGE_QUERY : null,
+        hubspotFormData,
     });
 };
 
+export const action: ActionFunction = async ({ request }: ActionArgs) => {
```
stevenbister, line 73: Rather than having this in here it might be worth moving it into it's own route (`actions/hubspot` or something) and updating the `RemixForm` action to post to that path.

Doing it in the the route means we'll have to duplicate this in each of the other route files (collections, products, posts etc.)

What do you think?

packages/osc-ecommerce/app/utils/hubspot.helpers.ts
```diff
@@ -0,0 +1,98 @@
+import { json } from '@remix-run/node';
+import { getValidationSchema } from '~/components/Forms/utils';
+import { hubspotFormsApiRequest } from '~/utils/server/hubspot.server';
+import { validateAction } from '~/utils/validation';
+
+export const shapeHubspotFormData = (
+    formFieldsData: Record<any, string>[],
+    formData: Record<any, string>
+) => {
+    debugger;
```
stevenbister, line 10: Can we take this out?

packages/osc-ecommerce/app/utils/hubspot.helpers.ts
```diff
@@ -0,0 +1,98 @@
...
+        return json(errorCases);
+    }
+
+    // If validation is passed - shape the hubspot form data to be submitted to Hubspot
+    const hubspotContactData = shapeHubspotFormData(hubspotFormFields, validatedFormInputData);
+
+    try {
+        const response = await hubspotFormsApiRequest(
+            'post',
+            `https://api.hsforms.com/submissions/v3/integration/secure/submit/${process.env.CONTACT_PORTAL_ID}/${formId}`,
```
stevenbister, line 86: Might be worth adding this env to the README if you've not already

packages/osc-ecommerce/app/utils/server/hubspot.server.ts
```diff
@@ -0,0 +1,48 @@
+import { Client } from '@hubspot/api-client';
+
+export const hubspotClient = () => {
+    return new Client({
+        accessToken: process.env.HUBSPOT_DEV_ACCESS_TOKEN,
```
stevenbister, line 5: Might be worth adding this env to the README if you've not already

stevenbister (changes requested):
Looking good man! Left a few comments and thoughts. I'm also getting an error when running locally
image

Think we might have to add any environment variables into GitHub/the workflows as well. Happy to have that in here or as a separate PR, whichever you think is easiest

packages/osc-ecommerce/app/components/Forms/HubspotForm/HubspotForm.spec.tsx
mattwendzina: Removed.

packages/osc-ecommerce/app/components/Forms/HubspotForm/textContent.ts
mattwendzina: Yeah good point, there was still some data being used, but moved it all into the mockData and organised it a bit better so it's clear what mockData is Hubspot and which is Sanity.

packages/osc-ecommerce/app/components/Forms/types.ts
mattwendzina: I think all the above is good. Done.

packages/osc-ecommerce/app/components/Forms/utils.tsx
```diff
@@ -0,0 +1,113 @@
...
+    index: number,
+    schema: ZodObject<ZodRawShape>,
+    setValidationErrors: Dispatch<SetStateAction<any>>,
+    validationErrors: Record<any, any>
+) {
+    let formInput:
+        | InputHTMLAttributes<HTMLInputElement>
+        | TextareaHTMLAttributes<HTMLTextAreaElement>;
+
+    // TODO - Need to work out the correct type for this!
```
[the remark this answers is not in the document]
mattwendzina: Updated

packages/osc-ecommerce/app/components/Forms/utils.tsx
mattwendzina: I think this was used on some of the other forms, but can't remember now lol. I've removed and will just add back in if needed.

packages/osc-ecommerce/app/routes/$slug.tsx
mattwendzina: Yep, thanks mate, this was part of the flow I hadn't fully figured out. Makes sense. Implemented. Took a while though as I hadn't used useFetcher so initially it was reloading the page with Remix Form, and then it took some more time to work out that the data gets returned via useFetcher rather than useLoaderData 😅 
Got there eventually though 😄

packages/osc-ecommerce/app/utils/hubspot.helpers.ts
mattwendzina: Ooops lol. Done.

packages/osc-ecommerce/app/utils/server/hubspot.server.ts
mattwendzina: Added.

packages/osc-ecommerce/app/routes/$slug.tsx
stevenbister: Nooiiice

packages/osc-ecommerce/app/routes/$slug.tsx
```diff
@@ -37,6 +39,22 @@ export const loader: LoaderFunction = async ({ request, params }) => {
...
+        | {} = {};
+    try {
+        const formData = (await getHubspotFormData(formId)) as HubspotFormData;
+        hubspotFormData = {
+            formFieldGroups: formData?.formFieldGroups,
+            submitText: formData?.submitText,
+        };
+    } catch (error) {
+        console.error('Unable to load form!');
+    }
```
stevenbister, line 56: This whole block will need adding to all of the routes as we'll want the option to use a form on any of them

stevenbister (changes requested):
Good stuff man, last couple of bits and I think we're there!

Something we'll need to think about doing is adding in success messages once the form submits "Thank you, we'll be in touch soon" kind of stuff, and how we handle that. Probs a separate task, I'll pop a note into Whimsical and perhaps we can discuss in the next sprint planning session?

packages/osc-ecommerce/app/components/Forms/utils.tsx
```diff
@@ -0,0 +1,112 @@
...
+    }, {});
+
+    return z.object(res);
+};
+
+export const transitionStates = (fetcher: Fetcher) => {
+    const isSubmitting = fetcher.state === 'submitting';
+    const isAdding =
+        fetcher.state === 'submitting' &&
+        fetcher.submission.formData.get('_action') === 'submitForm';
```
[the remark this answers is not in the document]
mattwendzina: Nice spot! Done.

packages/osc-ecommerce/app/components/Forms/HubspotForm/HubspotForm.tsx
```diff
@@ -0,0 +1,76 @@
...
+                {formFieldGroups?.map((data, index) => {
+                    return getInputType(
+                        data,
+                        index,
+                        validationSchema,
+                        setValidationErrors,
+                        validationErrors
+                    );
+                })}
+                <Button
```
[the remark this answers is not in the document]
mattwendzina: Good idea. Done.

packages/osc-ecommerce/app/routes/$slug.tsx
mattwendzina: I've refactored into one helper function and added to all same routes where modules are imported.

stevenbister (approved):
Nice man! LGTM 🚀

mattwendzina:
@stevenbister - I've had to make some additional changes..! After finishing I went and added a second form to start working on the newsletter to discover that it just displayed the contact form twice 😬 
Realised I was using a `.find()` on module.forms and so it only returned first one. I've changed this to a filter and then had to introduce some additional shaping and logic to ensure that the right form module gets the correct hubspot data.

mattwendzina:
Also needed to tweak the ids for the form components to make sure they are unique for the case of multiple forms where there could be overlap.

stevenbister:
> @stevenbister - I've had to make some additional changes..! After finishing I went and added a second form to start working on the newsletter to discover that it just displayed the contact form twice 😬 Realised I was using a `.find()` on module.forms and so it only returned first one. I've changed this to a filter and then had to introduce some additional shaping and logic to ensure that the right form module gets the correct hubspot data.

Ahaa oops! Tbf I didn't test putting multiple forms on a page 😅 I'll give it another once over now

packages/osc-ecommerce/app/routes/$slug.tsx
```diff
@@ -3,12 +3,13 @@ import { json } from '@remix-run/node';
 import { useLoaderData, useParams } from '@remix-run/react';
 import { useState } from 'react';
 import type { DynamicLinksFunction } from 'remix-utils';
+import type { HubspotFormData } from '~/components/Forms/types';
```
stevenbister, line 6: Eslint warning here `'HubspotFormData' is defined but never used`

stevenbister (changes requested):
Just that eslint warning but otherwise the new changes look good to me

packages/osc-ecommerce/app/routes/$slug.tsx
mattwendzina: Removed 👍

Issue #734, opened 2023-02-24
Build drawer component - Variant 1


EnaamAlum:
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/733

### Description

Standard Modal 
- this will display a modal in the centre of the screen. (Traditional Modal)
- should be able to switch between standard and alert dialog

Links
- https://www.radix-ui.com/docs/primitives/components/dialog
- https://www.radix-ui.com/docs/primitives/components/alert-dialog

Reviewer: Matt

PR #801, opened 2023-03-22
Refactor/modal

stevenbister:
- Closes https://github.com/Open-Study-College/osc/issues/734

## 📝 Description

Refactors the `Modal` component to make it more composable and set it up for the other drawer variations.

- Modal it buildable via props, and restricted to those

## 🚀 New behavior

- Modal is composable and can take any kind of content
- Can be programatically controlled
- Adds styles / animations
- Can be swapped over to an `AlertDialog` which interrupts the user with important content and expects a response

### DIFF
--- a/packages/osc-ui/package.json
+++ b/packages/osc-ui/package.json
@@ -30,6 +30,7 @@
     "@portabletext/react": "^2.0.2",
     "@radix-ui/react-accessible-icon": "^1.0.1",
     "@radix-ui/react-accordion": "^1.1.0",
+    "@radix-ui/react-alert-dialog": "^1.0.3",
     "@radix-ui/react-avatar": "^1.0.1",
     "@radix-ui/react-checkbox": "^1.0.2",
     "@radix-ui/react-dialog": "^1.0.2",
--- a/packages/osc-ui/src/components/Modal/AlertModal.stories.tsx
+++ b/packages/osc-ui/src/components/Modal/AlertModal.stories.tsx
@@ -1,0 +1,64 @@
+import type { Meta, Story } from '@storybook/react';
+import React from 'react';
+import { Button, ButtonGroup } from '../Button/Button';
+import type { AlertModalProps } from './AlertModal';
+import {
+    AlertModal,
+    AlertModalAction,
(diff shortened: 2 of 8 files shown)


mattwendzina:
Looks great mate 👍 

Only issue I seem to be having is with the accessibility voice over for the title. When I turn voice over on in storybook and open the modals the only content that is read out for me is the description. For example, on the Primary story for Modal, it reads out the description: "An optional accessible description to be announced when the dialog is opened."
But it does not read out the title: "An accessible title to be announced when the dialog is opened."

mattwendzina (commented):
Looks great mate 👍

Only issue I seem to be having is with the accessibility voice over for the title. When I turn voice over on in storybook and open the modals the only content that is read out for me is the description. For example, on the Primary story for Modal, it reads out the description: "An optional accessible description to be announced when the dialog is opened."
But it does not read out the title: "An accessible title to be announced when the dialog is opened."

mattwendzina:
> > Looks great mate 👍
> > Only issue I seem to be having is with the accessibility voice over for the title. When I turn voice over on in storybook and open the modals the only content that is read out for me is the description. For example, on the Primary story for Modal, it reads out the description: "An optional accessible description to be announced when the dialog is opened." But it does not read out the title: "An accessible title to be announced when the dialog is opened."
> 
> Hmmmm, are you using Chrome by any chance? I think there might be a bug either in VoiceOver or in Chrome/FF 🤔 It will read both the title and description in Safari but you're right, only the description in Chrome/FF 😅

Yeah, the issue for me was in Chrome

stevenbister:
Done a bit of digging and it looks like there's an open issue in Chrome which sounds like the same issue we've got here
- https://bugs.chromium.org/p/chromium/issues/detail?id=1404702&q=aria-labelledby&can=2

Given this and that the the title is handled by Radix not sure if there's much we can do here :/

mattwendzina (approved):
Based on fact that this bug is out of our hands, and posting issue/referencing the bug - LGTM 👍

hi-im-si (approved):

Issue #201, opened 2022-09-23
Auto complete - as you type it predicts your words


EnaamAlum:
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/196

### Description

n/a

PR #809, opened 2023-03-24
Feature: Algolia Workflows 

aaacafe786:
Closes [#201](https://github.com/Open-Study-College/osc/issues/201)

- The component is not available in OSC-ecommerce for testing purposes
- CICD Vitest workflows are not working for autocomplete

## 🚀 New behavior

- Connects the Autocomplete component to the CICD workflows, the component should now be available in https://pr-809-osc-ecommerce.fly.dev/search (temp)
- Vitest workflows should now be working
- fixed one test (findByTestId) rather than (findByText)


stevenbister:
@aaacafe786 Does this need rebasing against #786? Looks like the curved input is still here :S

stevenbister (approved):
Still not super happy about needing so many variations of the secrets as I don't understand why it works locally as normal?

But happy to approve the changes as it's achieving what we want

PR #818, opened 2023-04-03
Refactor/products route

stevenbister:
## 📝 Description

Updates the ecommerce site so that product pages are located at `/courses` rather than `/products`

- Products are located at the `/products` route

## 🚀 New behavior

- Products are now located at the `/courses` route
- Adds a `PATHS` constant which stores the path names (home, collections, products etc.)
- Updates any hard coded paths with the `PATHS` property


mattwendzina (changes requested):
Looks good to me mate. Just noticed a couple of TS Lint errors though:

- CourseCard.tsx - Line 99:
```
This JSX tag's 'children' prop expects a single child of type 'ReactNode', but multiple children were provided.ts(2746)
```
Looks like wrapping the children in a fragment would fix this.

- Courses/Collections $slug page, possibly other places on the `dynamicLinks` function: 
```
 Type 'object[]' is not assignable to type 'LinkDescriptor[]'.
    Type 'object' is not assignable to type 'LinkDescriptor'.
```

mattwendzina (approved):
LGTM

PR #822, opened 2023-04-04
Feature/add sanity preview kit

stevenbister:
## 📝 Description

Updates the ecommerce site to use [Sanity Preview Kit](https://github.com/sanity-io/preview-kit) to simplify how previewing draft content works with the CMS.

- Uses [Groq Store](https://github.com/sanity-io/groq-store) to stream draft content
- Have to swap page content out with draft content by setting it as state
- Previews are shown using the `?preview=true` query parameter

## 🚀 New behavior

- Uses [Sanity Preview Kit](https://github.com/sanity-io/preview-kit) to stream draft content using suspense
- Splits the page content out into `PageContent` and `PreviewContent` components 
    - Allows us to reuse the `Module` component loop across all pages
    - Also allows us to wrap in `PreviewSuspense` and use as a fallback
- Previews are now handled by sessions rather than query parameters. The session is generated by navigating to `/resources/preview`
- `PreviewContent` component is lazyloaded using `React.lazy` as we only want to load it when a CMS user is previewing.
- `resolveProductionUrl` function in `osc-studio` has been updated to pass the `/resources/preview` url which allows the session to be generated.
- Adds and styles a `PreviewBanner` component
- Updates page queries to exclude drafts. Previewkit will overlay the drafts on top of our query anyway so this should help to simplify the functions we need to interact with that data

## 📝 Additional Information

This is how you preview it in Sanity:

https://user-images.githubusercontent.com/23461173/230015212-804e079a-2a55-491b-ad54-72e62fceb3c8.mov

And any content you update should reflect on that screen without you needing to click publish (it might take a second for the saving spinner to finish initially)

There is currently a warning in the console when viewing preview mode that says `Cannot hydrate Suspense in legacy mode.` This is because we are using `ReactDOM.hydrate` rather than `ReactDOMClient.hydrateRoot`. As far as I can tell this appears to be fine, it just causes the whole page to refresh when a change is made in Sanity rather than it smoothly streaming in. Once we update to use `hydrateRoot` this behaviour should update nicely.


packages/osc-ecommerce/app/previewSession.server.ts
```diff
@@ -0,0 +1,12 @@
+import { createCookieSessionStorage } from '@remix-run/node'; // or cloudflare/deno
+
+const { getSession, commitSession, destroySession } = createCookieSessionStorage({
+    cookie: {
+        name: '__sanityPreviewSession',
+        httpOnly: true,
+        sameSite: 'lax',
+        secrets: ['s3cret1'],
```
hi-im-si, line 8: @Tom-Pearce would you recommend we passed in a secret this way or do you think it should be more encrypted?

hi-im-si (changes requested):
@stevenbister will need to update deployment URL

packages/osc-ecommerce/app/previewSession.server.ts
Tom-Pearce: I'd pull it from a `.env`

stevenbister: Cool, I've popped it into the .env file and added an example to the .sample/README too

Issue #851, opened 2023-04-26
Nested accordions with long titles cause overflow


stevenbister:
### Description

Looks like, related to this issue: https://github.com/Open-Study-College/osc/issues/823. Because the styles are getting cascaded then the smaller font-sizes aren't getting applied

### Steps to reproduce

<img width="436" alt="image" src="https://user-images.githubusercontent.com/23461173/234629861-64e7876e-8c64-490d-9793-b8b317fc6f51.png">

<img width="414" alt="image" src="https://user-images.githubusercontent.com/23461173/234629898-61e08cdf-f556-42be-8b00-c080bbbd794c.png">

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

Issue #823, opened 2023-04-06
Unable to apply primary styles to nested `Accordion`


stevenbister:
### Description

When I nest an `Accordion` within another `Accordion` and set the variant to `primary` I would expect to see the correct styles applied

### Link to Reproduction

http://localhost:6006/?path=/story/osc-ui-tabs--tab-to-accordion

### Steps to reproduce

1. Change the viewport to large or small mobile
2. Open the panel called "two"
3. The inner accordion styles shouldn't match the outer

### Release Version

0.65.0

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

I think this is because there isn't a `--primary` modifier class so it is simply inheriting the styles from the parent `Accordion` as it works with `--tertiary` as expected. Hopefully refactoring the scss to include this modifier should solve this issue 🤞

PR #852, opened 2023-04-27
Bugfix/accordion modifier styles

stevenbister:
- Closes https://github.com/Open-Study-College/osc/issues/823
- Closes https://github.com/Open-Study-College/osc/issues/851

## 📝 Description

Adds a context provider to the the accordion which passes the variant prop through to the subelements. This allows us to apply the same variant modifier across the board, in turn lets us style that modifier directly rather than relying on the cascade. This fixes the issue outlined in #823 as the `--secondary` and `--tertiary` modifiers would override `--primary` if it was nested as they were lower in the cascade. This was then leading to some pages to overflow as the nested accordion wasn't inheriting the correct styles and so the font was too big.

You can now see this working here on a product page in ecommerce
image

- `Accordion` styles rely solely on parent accordion variant modifier class

## 🚀 New behavior

- `Accordion` has a context provider
- Subelements each have their own variant modifier we can style against

### DIFF
--- a/packages/osc-ui/src/components/Accordion/Accordion.spec.tsx
+++ b/packages/osc-ui/src/components/Accordion/Accordion.spec.tsx
@@ -117,6 +117,16 @@ test('changes the icon to a chevron', () => {
 });
 
 test('renders the correct variant classname', () => {
+    const classes = [
+        'c-accordion',
+        'c-accordion__icon',
+        'c-accordion__header',
+        'c-accordion__trigger',
+        'c-accordion__item',
+        'c-accordion__content',
+        'c-accordion__text',
+    ];
+
     const { rerender } = render(
         <Accordion type="single" variant="secondary">
             <AccordionItem value="O">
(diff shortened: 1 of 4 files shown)


mattwendzina (commented):
This looks good mate. Only thing I've noticed, which is accordion related is that the outer accordion doesn't seem to give you the option to open and close the same drawer. You can only close one by opening another. Whereas the inner accordion allows you to open and close the same drawer. 
I've attached a video below to show what I mean. Is this the intended behaviour?

stevenbister:
> This looks good mate. Only thing I've noticed, which is accordion related is that the outer accordion doesn't seem to give you the option to open and close the same drawer. You can only close one by opening another. Whereas the inner accordion allows you to open and close the same drawer. I've attached a video below to show what I mean. Is this the intended behaviour?

Oh yeah that's intended, I've just not added the `collapsible` prop to the outer one

mattwendzina (approved):
LGTM

PR #861, opened 2023-05-10
Fixes from Sanity Script work

mattwendzina:
## 📝 Description

A few fixes related to importing the content from Wordpress to Sanity

## 🚀 New behavior

- Fix Sanity validation error when content in Accordion is left empty
- Fix bug in CollectionCard.tsx where JSX isn't being returned 
- Enable content to inherit the `o-container--flush` class when fullWidth is set to true in Sanity


stevenbister (approved):
LGTM

PR #863, opened 2023-05-11
feat: add remix routes for awards and create Sanity query

mattwendzina:
## 📝 Description

Create new Remix routes for handling the award pages

No route for handling Awarding bodies

## 🚀 New behavior

2 new routes:
- awards/index.tsx
- awards/$slug.tsx

## 📝 Additional Information

The `index.tsx` is simply redirecting to the homepage as this page is not needed

### DIFF
--- a/packages/osc-ecommerce/app/queries/sanity/fragments/buildUrls.ts
+++ b/packages/osc-ecommerce/app/queries/sanity/fragments/buildUrls.ts
@@ -21,6 +21,6 @@ export const buildUrls = groq`
         "slug": "/${PATHS.PRODUCTS}/" + store.slug.current,
     },
     (_type == "award") => {
-        "slug": "/${PATHS.AWARDS}/" + store.slug.current,
+        "slug": "/${PATHS.AWARDS}/" + slug.current,
     }
 `;
--- a/packages/osc-ecommerce/app/queries/sanity/fragments/buildUrls.ts
+++ b/packages/osc-ecommerce/app/queries/sanity/fragments/buildUrls.ts
@@ -19,5 +19,8 @@ export const buildUrls = groq`
     },
     (_type == "product") => {
         "slug": "/${PATHS.PRODUCTS}/" + store.slug.current,
+    },
+    (_type == "award") => {
+        "slug": "/${PATHS.AWARDS}/" + store.slug.current,
     }
 `;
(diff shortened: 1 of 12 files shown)


mattwendzina:
Thanks @stevenbister - Made those updates, and have also updated the type to "Award" and cascaded that name change through the other files.

packages/osc-studio/utils/resolveProductionUrl.ts
```diff
@@ -53,6 +53,9 @@ export function resolveProductionUrl(doc: SanityDocumentWithSlug) {
         case 'product':
             url.pathname = `${PATHS.PRODUCTS}/${slug}`;
             break;
+        case 'award':
+            url.pathname = `${PATHS.PRODUCTS}/${slug}`;
```
stevenbister, line 57: Typo?

packages/osc-ecommerce/app/queries/sanity/fragments/buildUrls.ts
```diff
@@ -19,5 +19,8 @@ export const buildUrls = groq`
     },
     (_type == "product") => {
         "slug": "/${PATHS.PRODUCTS}/" + store.slug.current,
+    },
+    (_type == "award") => {
+        "slug": "/${PATHS.AWARDS}/" + store.slug.current,
```
stevenbister, line 24: Think you just need to pop the store part off here so it would look like:

```suggestion
        "slug": "/${PATHS.AWARDS}/" + slug.current,
```

The store bit is specific for Shopify pages

stevenbister (changes requested):
Looks good man, couple of typos but nothing major.

Was thinking, do you reckon it's worth changing this to say "Awarding Bodies"?

image

mattwendzina:
> Looks good man, couple of typos but nothing major.
> 
> Was thinking, do you reckon it's worth changing this to say "Awarding Bodies"?
> 
> image

Yeah, agree. Updated.

My bad on those typos, clearly rushing..! 😅 Sorted now.

stevenbister (changes requested):
Looks good! Sorry, noticed one more step that needs adding, in `lib/sanity/config.ts` we need to add `award` to the `includeTypes` property in the `previewConfig`.

I'm gonna add all of these steps into the wiki for next time 😅

stevenbister (approved):
👍

PR #865, opened 2023-05-12
Feature/add forms to content media

mattwendzina:
## 📝 Description

Adds Forms into Content Media

Currently Content Media only allows Images

## 🚀 New behavior

This allows Forms to be added to Content Media

No

### DIFF
--- a/packages/osc-ecommerce/app/components/ContentMedia/ContentMedia.tsx
+++ b/packages/osc-ecommerce/app/components/ContentMedia/ContentMedia.tsx
@@ -188,8 +188,9 @@ const ContentMediaBlockModule = (props: ContentMediaBlockProps) => {
                         );
                     } else if (media?._type === 'module.forms') {
                         const moduleForm = media as formModule;
-                        mediaBlock = <Forms module={moduleForm} key={moduleForm._key} />;
+                        return <Forms module={moduleForm} key={moduleForm._key} />;
                     }
+                    return null;
                 })}
             </Carousel>
         );
--- a/packages/osc-ecommerce/app/components/ContentMedia/ContentMedia.tsx
+++ b/packages/osc-ecommerce/app/components/ContentMedia/ContentMedia.tsx
@@ -10,7 +10,8 @@ import {
     rem,
     useSpacing,
 } from 'osc-ui';
-import type { contentMediaModule, contentMediaSlide } from '~/types/sanity';
+import type { contentMediaModule, contentMediaSlide, formModule } from '~/types/sanity';
+import { Forms } from '../Forms/Forms';
 
 const perView = (perView: number | undefined) => (perView ? perView : 1);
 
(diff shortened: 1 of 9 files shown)


packages/osc-ecommerce/app/components/ContentMedia/ContentMedia.tsx
```diff
@@ -140,47 +141,42 @@ const ContentMediaBlockModule = (props: ContentMediaBlockProps) => {
...
+                            spacing: 16,
                         },
-                    }}
-                >
-                    {media?.mediaType.map((media) =>
-                        media?.image ? (
+                    },
+                }}
+            >
+                {media?.mediaType.map((media) => {
```
stevenbister, line 177: Getting an eslint error on this line, doesn't look like we're returning anything in some cases

packages/osc-ecommerce/app/components/Forms/Forms.tsx
```diff
@@ -93,9 +93,9 @@ const Form = (props: FormProps) => {
     );
 };
 
-export const Forms = (props: { module: formModule }) => {
+export const Forms = (props: { module: formModule; addContainer?: boolean }) => {
     // Module data coming from Sanity
-    const { module } = props;
+    const { module, addContainer = false } = props;
```
stevenbister, line 98: Is there a reason for using this prop over adding the container class directly like the other components?

stevenbister (changes requested):
Nice man, one eslint warning, and a question from me :)

Do you also think it's worth adjusting the space between the two columns when the form is set? They feel a bit close at the moment 🤔

packages/osc-ecommerce/app/components/Forms/Forms.tsx
mattwendzina: As discussed, some cases, like ContentMedia won't want a container.

mattwendzina:
> Nice man, one eslint warning, and a question from me :)
> 
> Do you also think it's worth adjusting the space between the two columns when the form is set? They feel a bit close at the moment 🤔

I found a way to add this conditionally in `ContentMedia.tsx` and have added a 40/60 split to the script for whenever ContentMedia has a form. I believe this creates the level of space we discussed. See screenshot below:

Screenshot 2023-05-15 at 15 04 32

stevenbister:
> I found a way to add this conditionally in `ContentMedia.tsx` and have added a 40/60 split to the script for whenever ContentMedia has a form. I believe this creates the level of space we discussed. See screenshot below:
> 
> Screenshot 2023-05-15 at 15 04 32

Sorry mate, not sure if this is quite working for me? The `start-7` bit is there but the other column is still the same width? Do I need to do something to set it? 

Also; if we swap the order, so form on the left and content on the right then it goes a bit wonky
image

These are the settings for the slide:
image

stevenbister (approved):
LGTM :)

PR #866, opened 2023-05-15
Feature/product e2e

### LINKED ISSUE #915: Add initial e2e tests for the productt flow.
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/914

### Description

Must cover the following steps:
Product variant can be selected
Product slide out syncs with main product form
product can be added to cart

stevenbister:
- Closes #915 

## 📝 Description

Adds the initial e2e tests for the product page flow. I'm saying initial here as I expect these to expand as and when the cart is built out further and the mini cart is added. At the moment they simply test if you can add a product and it's variants to the cart, and that the drawer form is kept in sync with the main form on the page.

In doing this I noticed that the buttons in the form weren't getting disabled when the form is transitioning, meaning that if you changed the variant you could still potentially submit the form before the form finished updating and end up adding the wrong variant to the cart. Although this is probably noticeable because Playwright moves through the click events pretty fast I figured adding that disabled props to the button couldn't hurt.

I also updated the Playwright version in ecommerce so we can use the UI mode they've added which makes creating the tests easier with a visual reference of what is happening.

- Using playwright version 1.31
- Buttons aren't disabled when form is transitioning

## 🚀 New behavior

- Adds e2e tests for product flow
- Updates playwright to 1.33
- Disables buttons when form is in a transitioning state


mattwendzina (commented):
Nice mate, looks good.
Getting one failure though:
Last test in navbar.spec.ts - 'clicking outside of the menu closes it' - error is:
`Timed out 5000ms waiting for expect(received).not.toBeVisible()`

stevenbister:
> Nice mate, looks good. Getting one failure though: Last test in navbar.spec.ts - 'clicking outside of the menu closes it' - error is: `Timed out 5000ms waiting for expect(received).not.toBeVisible()`

Ah yeah, I've updated the locator on that one, I think the click was happening in the middle of the dropdown :/

mattwendzina (approved):
LGTM!

Issue #779, opened 2023-03-09
Update GitHub actions


stevenbister:
### Description

It would be great to refactor the GitHub actions and workflows to make them a bit more maintainable/faster

### Problem Statement/Justification

- Actions and workflows are difficult to maintain & edit
- Hard to see at a glance which actions have passed/failed as some are intended to fail for a successful deploy

### Proposed Solution or API

See an example of how NX suggest using this here: https://www.youtube.com/watch?v=GT7XIwG1i5A&list=PLakNactNC1dH38AfqmwabvOszDmKriGco&index=11

Also look into changing the NX token in our nx.json so we have a read only locally and read/write in CI
https://nx.dev/nx-cloud/account/access-tokens

Issue #871, opened 2023-05-16
SEO Pane in Sanity not working


stevenbister:
### Description

At the moment the seo pane in Sanity hangs when accessing it

### Link to Reproduction

https://osc-studio.fly.dev/

### Steps to reproduce

Go to any of the pages in Studio
Click on SEO at the top

### Release Version

0.86.2

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

The reason this looks like it's happening is because when we updated the preview pane stuff we changed the urls in the resolveUrl file, which was being shared between the two panes. Therefore the SEO one is now failing because the url is wrong. To fix will be a case of allowing the SEO pane to access different urls to the preview pane

PR #872, opened 2023-05-16
Refactor/workflows

stevenbister:
- Closes: #908 
- Closes: #779 
- Closes: #910 

## 📝 Description

Refactors our workflows and Dockerfiles to (hopefully) make them a bit more stable and maintainable. I've tried to reduce the complexity by removing the composite actions and replacing them with larger [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows).

I've removed the Dockerfiles directory and moved the Dockerfiles into each app that needs them. As each file is slightly different it felt like it made more sense to co-locate them and leans into the package isolation principle.

I've also remove the pr toml files as this was causing issues with deployments. The initial need for them was unclear and the builds seem to run fine without them. Along the same lines I've also removed the volumes stuff from the main toml file and the workflows. As we're not hosting a db on Fly or (AFAIK) using any storage on there I don't think we need to have them.

Workflows will now work like this:

There are a few general workflows which will run whenever a PR is opened
- Chromatic
- Tests, although these now use `nx affected` so should only run the tests for the apps/packages that have had changes.
- Codeball

We then have one workflow for each app (ecommerce, studio and academic hub), these will run when a PR is opened or the branch is merged into main and will run a workflow based on the matching conditions. Using ecommerce as an example
- If the PR is open and has the label of `build:osc-ecommerce` run the deploy preview workflow, which builds the PR app on Fly
- If the PR is closed or the label `build:osc-ecommerce` is removed then destroy the PR app on Fly
- If the PR is merged and has the label  `build:osc-ecommerce` deploy the production app to Fly

The only one with an extra step at the moment is academic hub which will create the PlanetScale branches/deployment requests alongside building the app.

Finally I've had to disable the SEOPane plugin in Sanity for the time being as when deploying the app it was causing an error. [I've opened a ticket here](https://github.com/Open-Study-College/osc/issues/871).

- Workflows are handled by composite actions with a lot of duplication of passing secrets through to each step
- Linting, tests and deployments are all part of one workflow file, which only runs for ecommerce and academic hub apps

## 🚀 New behavior

- Composite actions have been removed and the steps moved into reusable workflows
- Linting and tests have been moved into their own workflow, using `nx affected` to decide which tests should be run
- SEOPane in osc-studio has been disabled

When you open a PR you will need to add the labels `build:<package-name` for it to build the app (this won't always be necessary).

## 📝 Additional Information

I _think_ I've managed to cut down on some of the build time with the Dockerfiles and it does seem to be caching but I'm not confident that the first run on each PR won't be loooong. Will monitor it and maybe address further down the line. You can build Dockerfiles in Fly so that could be an option but I had issues with permissions with building Sanity initially.

I've added a docker-compose file to the root which isn't used in the actions but makes building and testing locally a bit easier I found

I've not built academic hub in this example as that one will fail due to the `API_GATEWAY_URL` not existing yet


stevenbister:
It's also really noticeable now that we have failing tests and are rubbish at Typescript 😂

packages/osc-ecommerce/app/components/Footer/Footer.tsx
```diff
@@ -68,7 +68,7 @@ export const SiteFooter = (props: SiteFooterProps) => {
                     ) : null}
                 </div>
 
-                {socials.length > 0 ? (
+                {socials && socials?.length > 0 ? (
```
mattwendzina, line 71: Are these changes and the changes in root.tsx just a bit of boy scouting?

mattwendzina (commented):
I've read through the PR notes and looked through the code. I don't really have any expertise here so can't add much value, but at least had a read through and looks good.

Left one small comment about what just looks like some boy scouting.

Only other thing - Is it worth updating the ReadMe with some explanation of how things now work?

packages/osc-ecommerce/app/components/Footer/Footer.tsx
stevenbister: Ahhh yeah, when I deployed the production dataset the site broke because we didn't have any data in these areas so needed to add these in. Figured it wouldn't hurt to leave them here as it's somewhat related

stevenbister:
Thanks @mattwendzina, appreciate you looking! In regards to

> Only other thing - Is it worth updating the ReadMe with some explanation of how things now work?

Yep good shout, was thinking of adding it to the Wiki once this gets merged but there is a section in there already about actions so I'll update it

mattwendzina (approved):
I'm giving it the 👍 though I think this will at least one other..! I can't wait to have a pipeline that can actually tells us where things are really failing 😄

PR #898, opened 2023-05-17
Refactor/price component

stevenbister:
## 📝 Description

Started working on the cards for the cart and noticed that the price part of it matches the one on the product form and within the course cards. So I've moved this out of `osc-ecommerce` and created a `<Price>` component inside `osc-ui` instead. ~~It's slightly different to the ones in the product card so I've left that as is for now, although we could expand this to handle those if needs be later on.~~ (Changed my mind lol)

I'm anticipating this getting used in the mini cart from the looks of things, so I've imported the stylesheet into the `root.tsx` file ready for when that gets built.

- `<Price>` component sits in `osc-ecommerce` and is coupled closely to the use of Hydrogen's `<Money>` component.

## 🚀 New behavior

- `<Price>` has been moved into `osc-ui` and set up to take components as children/props so we can pass the `<Money>` component through when needed.


packages/osc-ui/src/components/Price/Price.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const saleModifier = useModifier('c-price', 'on-sale');
+    const classes = classNames('c-price', compareAtPrice ? saleModifier : '', className);
+
+    const sizeModifier = useModifier('c-price__item', size);
+    const itemClasses = classNames('c-price__item', sizeModifier, 'u-mb-0');
+
+    // Here we're creating a slot for the compareAtPrice element to sit in.
+    // This helps us to keep the amount of nested elements to a minimum
+    // as well as merge the required classes.
+    const SaleComponent = compareAtPrice ? Slot : 'span';
```
mattwendzina, line 49: Probably being stupid here, but looks like the SaleComponent only gets rendered if compareAtPrice is true, so I can't see a scenario where the SaleComponent could be a span?

mattwendzina (commented):
Left one small comment, but other than that, looks good 👍

packages/osc-ui/src/components/Price/Price.tsx
stevenbister: Aha yeah good spot! Dropped that

mattwendzina (approved):
LGTM

PR #907, opened 2023-05-17
Feature/cart item card

stevenbister:
- Closes: #917 
- Closes: #918 
- Closes: #919 

## 📝 Description

Adds the first cart component, this is the "line item" in the cart (cards on the left)
image

I've refactored a few elements in the cards, so they take new props/modifier classes to change the styles rather than being dependant on their parent. For example the `Card` now has a `hasShadow` and `hasBorder` prop which applies a modifier class, rather than the shadow or border coming from a specific card type. This has allowed me to build up this component by just changing props and adding a few utility classes here and there. I'm also thinking these modifier can potentially be added to Sanity to give CMS users a bit more control over styles (e.g. there are some cards on the homepage that don't have a shadow).

I've added a couple of modifiers to the `Select` and a new variant to the `Accordion` component to match up the design.

- Shadows and borders are controlled by cart variant
- Small title styles are handled by `is-small` class
- Title and subtitle position/colour are handled by parent class

## 🚀 New behavior

- Shadow and borders are controlled by props/modifier classes
- Adds a transparent modifier class
- Replaces `is-small` with `--small` modifier
- Title and subtitle position/colour are now handled by modifier class
- `Select` now has a wrappable inline variant and the label can be darkened to the secondary colour
- Adds a quaternary variant to the accordion 

## 📝 Additional Information

This only adds the component to Storybook, will add it to the cart in osc-ecommerce once the other components are built and I start building out the page

### DIFF
--- a/packages/osc-ecommerce/app/components/Cards/BioCard.tsx
+++ b/packages/osc-ecommerce/app/components/Cards/BioCard.tsx
@@ -22,7 +22,7 @@ export const BioCard = (props: Props) => {
         : data?.image?.image?.secure_url;
 
     return (
-        <Card>
+        <Card hasShadow>
             {data?.image ? (
                 <CardImage isRounded>
                     {src && data?.image?.image ? (
--- a/packages/osc-ecommerce/app/components/Cards/CollectionCard.tsx
+++ b/packages/osc-ecommerce/app/components/Cards/CollectionCard.tsx
@@ -23,7 +23,7 @@ export const CollectionCard = (props: Props) => {
     const { theme, featuredImage } = data?.reference;
 
     return (
-        <OSCCollectionCard size={data?.variant}>
+        <OSCCollectionCard size={data?.variant} hasShadow>
             {featuredImage?.src ? (
                 <CardImage>
                     <Image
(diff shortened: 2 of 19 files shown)


mattwendzina (commented):
Looks great mate. 

Just a couple of things:
- There's an accessibility violation on the Cart Card - Color contrast issue. I think it's on the Save for later button. Looks like it probably needs to be neutral-700 rather than 600?
- Tabbing on the nested accordion - when the nest is closed I'd expect it to tab straight over it, but it still tabs through the inner accordion items which may be a bit confusing for a screen reader. Not the biggest of issues, but if it's a relatively easy fix then this may be worth doing?

stevenbister:
Thanks man! 
> * There's an accessibility violation on the Cart Card - Color contrast issue. I think it's on the Save for later button. Looks like it probably needs to be neutral-700 rather than 600?

There was two, one was this which I've updated but the other is on the pink colour here
image
As that's basically the main brand colour I'm not sure whether we can change that 😅 @hi-im-si any thoughts?

> * Tabbing on the nested accordion - when the nest is closed I'd expect it to tab straight over it, but it still tabs through the inner accordion items which may be a bit confusing for a screen reader. Not the biggest of issues, but if it's a relatively easy fix then this may be worth doing?

Ah yeah I see what you mean, I'm gonna make a note of this as I think I'll need to refactor the accordion slightly and remove the `forceMount` prop on that content section. But that will require a bit of tweaking to how the animations fire

PR #928, opened 2023-05-17
Bugfix/getHubspotForms causing pages without modules to crash

stevenbister:
## 📝 Description

Noticed that if pages don't have any modules defined from Sanity then the `getHubspotForms` method throws an error

image

Have added an early return if there are no modules which should get around this

- Pages throw an error if there are no defined CMS modules

## 🚀 New behavior

- Early return in `getHubspotForms` will prevent error from being thrown

### DIFF
--- a/packages/osc-ecommerce/app/utils/hubspot.helpers.ts
+++ b/packages/osc-ecommerce/app/utils/hubspot.helpers.ts
@@ -142,9 +142,9 @@ export const validateAndSubmitHubspotForm = async (
 export const getHubspotForms = async (page: SanityPage) => {
     let forms: formModule[] = [];
 
-    // If contentMedia has any forms then push them to forms array
     if (!page.modules) return;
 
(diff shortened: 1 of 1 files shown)


packages/osc-ecommerce/app/utils/hubspot.helpers.ts
```diff
@@ -143,6 +143,8 @@ export const getHubspotForms = async (page: SanityPage) => {
     let forms: formModule[] = [];
 
     // If contentMedia has any forms then push them to forms array
+    if (!page.modules) return;
```
mattwendzina, line 146: Perhaps the comment should stay below this? ie, on line 147

stevenbister: Ah yeah good spot

mattwendzina (approved):
LGMT

PR #929, opened 2023-05-18
Feature/checkbox size

stevenbister:
## 📝 Description

Adds a size modifier to the checkbox and label. Noticed that part of this component
image

Is a big checkbox so I've added some modifier classes to allow us to create checkboxes at this size. Rather than going from m - l I've jumped straight up to xl as that gives us some room for an inbetween size if we ever need one.

- Checkboxes are all a standard size

## 🚀 New behavior

- Adds props and modifier classes to checkbox and label


mattwendzina (approved):
LGTM

PR #930, opened 2023-05-18
Feature/callout secondary

### LINKED ISSUE #920: [S59.1]: Add on Callout
### Link to Epic

### Description

n/a

### Linked Tasks

_No response_

stevenbister:
- Closes #920 

## 📝 Description

Adds a new variant to the callout banner component along with a few new modifiers to adjust the styles to match the add on banner in the Cart page design.

The offset value is a bit of a magic number and only really works with the xl checkbox variant. I think this is _probably_ fine as I can't see anywhere else this style is being used.

- Singular variant callout banner

## 🚀 New behavior

- Adds variant prop and primary/secondary modifiers
- Adds context to component so we can pass the variant through
- Adds an offset prop to the content and footer to push things inline when using components such as the checkbox

### DIFF
--- a/packages/osc-ui/src/components/CalloutBanner/CalloutBanner.stories.tsx
+++ b/packages/osc-ui/src/components/CalloutBanner/CalloutBanner.stories.tsx
@@ -1,6 +1,7 @@
 import type { Meta, Story } from '@storybook/react';
 import React from 'react';
 import { Button } from '../Button/Button';
+import { Checkbox } from '../Checkbox/Checkbox';
 import { Image } from '../Image/Image';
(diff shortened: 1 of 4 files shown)


mattwendzina (commented):
Looks good mate 👍 

Only thing is I've noticed that the content is spilling out of the banner for the Primary banner when it's set to small mobile in Storybook.

stevenbister:
> Only thing is I've noticed that the content is spilling out of the banner for the Primary banner when it's set to small mobile in Storybook.

Ooops! That long email address is doing it in 😂 Should be sorted now

mattwendzina (approved):
LGTM!

PR #933, opened 2023-05-23
Feature/cart

stevenbister:
- Closes: https://github.com/Open-Study-College/osc/issues/945
- Closes: #946 
- Closes: https://github.com/Open-Study-College/osc/issues/947
- Closes: #944

## 📝 Description

Adds the following cart components to the cart route and sets up the layout.

- #898 
- #907 
- #931 

I moved the layout of the cart into it's own `CartLayout` component rather than keeping it in the `cart` route to try and keep that file a bit leaner.

In order to be able to pull the content from Sanity into the cart I've added a few utilities to the `cart.helpers` and `storefront.helpers` files. These find and transform the ids from the products to match the ids in Sanity so we can query the products content. We then push that data into the carts lines array. If you want to see the output of that data you can log the `cart` object in the `CartLayout` component and add http://localhost:2000/courses/aat-level-2-and-3-accounting to the cart.
Because we're grabbing the intro bit of content I've also added a `stripMarks` utility which will remove any of the styling utilities added from the cms.

Finally I've added a section to the settings in Sanity which will allow users to change the message when the cart is empty.

This is a fairly non-functioning cart although you should be able to move through to the checkout. The other functionality (remove items, update items etc.) will come in later PRs.

N/a

## 🚀 New behavior

- Integrates cart components with cart route
- Adds helper functions to get/set Shopify/Sanity product ids
- Adds helper function to add Sanity content to the cart line item
- Adds helper function to remove styles from wysiwyg content
- Adds cart settings to Sanity

## 📝 Additional Information

If you need to remove an item from your cart you can do this by navigating to http://localhost:2000/graphiql and adding these queries:
```gql
query Cart($Cart: ID!) {
  cart(id: $Cart) {
    id
    createdAt
    updatedAt
    lines(first: 10, reverse: true) {
      nodes {
        id
        merchandise {
          ... on ProductVariant {
            id
            title
          }
        }
      }
    }
  }
}

mutation cartLinesRemove($Cart: ID!, $lineIds: [ID!]!) {
  cartLinesRemove(cartId: $Cart, lineIds: $lineIds) {
    cart {
      id
    }
    userErrors {
      field
      message
    }
  }
}
```
And these variables:
```gql
{
  "Cart": "",
  "lineIds": []
}
```
First you will need to get the cart id, to do this you can log the `cart` object in the `CartLayout` component and add it to the `Cart` variable above. Run the cart query and this will give you the ids for the line items. Then just add those to the `lineIds` as array items, and run the mutation.

Test are planned in for a later sprint


mattwendzina (commented):
Looks great mate!
Couple of general comments, which may well be addressed in the proceeding PRs!
- Cart allows you to add multiple of the same item - there's no visual indicator to show quantity, you just see the price increase. Presuming this comes in one of the following PRs?
- When on the cart page, changing the method, e.g. study to online doesn't update anything. Again, guess this comes in a following PR?
- If they added multiple of the same, would we just want one card and an indicator on the quantity, rather than multiple of the same cards? May be addressed in further PR..!? 🙈

stevenbister:
Thanks man appreciate that :) 

> * Cart allows you to add multiple of the same item - there's no visual indicator to show quantity, you just see the price increase. Presuming this comes in one of the following PRs? / * If they added multiple of the same, would we just want one card and an indicator on the quantity, rather than multiple of the same cards? May be addressed in further PR..!? 🙈

Ah yeah good point, hadn't really thought about that... I was just going under the assumption that you likely wouldn't purchase multiple of the same course 😅 I think the current cart functionality doesn't allow you to do multiples so I guess we could mimic that functionality 🤔 

> * When on the cart page, changing the method, e.g. study to online doesn't update anything. Again, guess this comes in a following PR?

Yeah this one is in #935

mattwendzina:
> Thanks man appreciate that :)
> 
> > * Cart allows you to add multiple of the same item - there's no visual indicator to show quantity, you just see the price increase. Presuming this comes in one of the following PRs? / * If they added multiple of the same, would we just want one card and an indicator on the quantity, rather than multiple of the same cards? May be addressed in further PR..!? 🙈
> 
> Ah yeah good point, hadn't really thought about that... I was just going under the assumption that you likely wouldn't purchase multiple of the same course 😅 I think the current cart functionality doesn't allow you to do multiples so I guess we could mimic that functionality 🤔
> 
> > * When on the cart page, changing the method, e.g. study to online doesn't update anything. Again, guess this comes in a following PR?
> 
> Yeah this one is in #935

Yeah, I thought about it a bit and thought maybe it's possible that an educator themselves might buy multiple courses for students they're teaching, or say a parent homeschooling multiple kids. I guess if these were realistic scenarios they would need multiple in order to get the qualification - instead of just buying one and teaching multiples students with it. Maybe we don't actually have this kind of usage but we could check at least. I think we need something just incase people accidentally add a second and don't understand why the price is so high lol. 

Sorry, one other thing I forgot to mention... When the basket is empty there is a different message from desktop to mob view. In desktop there is the message that you have nothing in the card and a 'Browse our courses' button. On mob, it's the 'Add more courses' button. Guess this should also be 'Browse our courses'?

stevenbister:
> Yeah, I thought about it a bit and thought maybe it's possible that an educator themselves might buy multiple courses for students they're teaching, or say a parent homeschooling multiple kids. I guess if these were realistic scenarios they would need multiple in order to get the qualification - instead of just buying one and teaching multiples students with it. Maybe we don't actually have this kind of usage but we could check at least. I think we need something just incase people accidentally add a second and don't understand why the price is so high lol.

So I checked with Tom and it's by design that we only allow purchasing of courses as singles. 
> yeah we only allow purchasing of courses as singles
theres no scenario where 1 person needs 2x a level biology
if its a parent for 2 children they should create 2 orders
I'd suggest making sure there is the ability to buy multiple though for things like gift cards
(i don't think we actively sell them at the moment but we have in the past)

So based on this I've updated the functionality in the product form, so now we will check if the product is in the cart and if it is; then set the quantity we add to 0. I've also added a helper function to exclude gift vouchers from this. I've popped it in as part of this PR as it seems related but let me know if you want me to split it out.

> Sorry, one other thing I forgot to mention... When the basket is empty there is a different message from desktop to mob view. In desktop there is the message that you have nothing in the card and a 'Browse our courses' button. On mob, it's the 'Add more courses' button. Guess this should also be 'Browse our courses'?

Aha yeah my bad -- sorted!

mattwendzina:
> > Yeah, I thought about it a bit and thought maybe it's possible that an educator themselves might buy multiple courses for students they're teaching, or say a parent homeschooling multiple kids. I guess if these were realistic scenarios they would need multiple in order to get the qualification - instead of just buying one and teaching multiples students with it. Maybe we don't actually have this kind of usage but we could check at least. I think we need something just incase people accidentally add a second and don't understand why the price is so high lol.
> 
> So I checked with Tom and it's by design that we only allow purchasing of courses as singles.
> 
> > yeah we only allow purchasing of courses as singles
> > theres no scenario where 1 person needs 2x a level biology
> > if its a parent for 2 children they should create 2 orders
> > I'd suggest making sure there is the ability to buy multiple though for things like gift cards
> > (i don't think we actively sell them at the moment but we have in the past)
> 
> So based on this I've updated the functionality in the product form, so now we will check if the product is in the cart and if it is; then set the quantity we add to 0. I've also added a helper function to exclude gift vouchers from this. I've popped it in as part of this PR as it seems related but let me know if you want me to split it out.
> 
> > Sorry, one other thing I forgot to mention... When the basket is empty there is a different message from desktop to mob view. In desktop there is the message that you have nothing in the card and a 'Browse our courses' button. On mob, it's the 'Add more courses' button. Guess this should also be 'Browse our courses'?
> 
> Aha yeah my bad -- sorted!

Re Tom:
"theres no scenario where 1 person needs 2x a level biology... if it's a parent... create 2 orders"
Isn't that a scenario?! 😂 
Well fair enough, if that's how we already do it and that's how the want to keep it..!

Sounds good in terms stopping multiples being added. An alternative approach though, which I think might be better UI would be to simply pass `isAlreadyInCart` to `AddToCart.tsx` and change the message to `Already in Cart` or something, and also disable the button when `isAlreadyInCart` is true. Think this is better than being able to click it again. If a user does want multiple this might be confusing to them as it will just not update with another one. At least adding a message to the button makes it a bit clearer that they can only add one? 🤷

stevenbister:
> Sounds good in terms stopping multiples being added. An alternative approach though, which I think might be better UI would be to simply pass `isAlreadyInCart` to `AddToCart.tsx` and change the message to `Already in Cart` or something, and also disable the button when `isAlreadyInCart` is true. Think this is better than being able to click it again. If a user does want multiple this might be confusing to them as it will just not update with another one. At least adding a message to the button makes it a bit clearer that they can only add one? 🤷

Yeah that's a good shout, updated!

stevenbister:
Have just thought, (and added) a case into the `addLinesToCart` on the backend to also filter out multiple lines, just in case someone manages to change the code on the frontend to allow it

hi-im-si:
> > Sounds good in terms stopping multiples being added. An alternative approach though, which I think might be better UI would be to simply pass `isAlreadyInCart` to `AddToCart.tsx` and change the message to `Already in Cart` or something, and also disable the button when `isAlreadyInCart` is true. Think this is better than being able to click it again. If a user does want multiple this might be confusing to them as it will just not update with another one. At least adding a message to the button makes it a bit clearer that they can only add one? 🤷
> 
> Yeah that's a good shout, updated!

image

mattwendzina:
> > > Sounds good in terms stopping multiples being added. An alternative approach though, which I think might be better UI would be to simply pass `isAlreadyInCart` to `AddToCart.tsx` and change the message to `Already in Cart` or something, and also disable the button when `isAlreadyInCart` is true. Think this is better than being able to click it again. If a user does want multiple this might be confusing to them as it will just not update with another one. At least adding a message to the button makes it a bit clearer that they can only add one? 🤷
> > 
> > 
> > Yeah that's a good shout, updated!
> 
> Something i quickly screengrabbed on a competitors website @stevenbister @mattwendzina image

I personally really like this kind of idea. I think there are scenarios where people might want to buy multiples and could just be annoyed or confused that it only allows you to only add one. At least now, we are telling them the item is already in the cart, which implies there can only be one obviously, but having a message at least acknowledges that we think there may be scenarios where people may want more than one which I think a better UX, and gives them a next step action on the screen - contacting us. 

One consideration on how we'd implement this which I think kind of applies to the 'Already in Basket' message too... Atm we're taking them directly to checkout when they click Add To Basket, but in reality we won't be doing this. Having the message 'Already in Basket' as soon as they add it will probably look a bit strange. Feels like that would only make sense if they re-visit the page. Think it would need to be something more like 'Added' when they initially add it. Annoying to have to have logic for different messages though depending on the user journey.  Maybe 'Added' is just a better term that could cover all cases? 
But yeah, in terms of that additional messaging, we wouldn't want that unless they re-visit the page for similar reasons. We could have some logic that would show the message if they clicked on the Add To Basket button a second time maybe? Perhaps I'm overthinking this though now?! 😅

mattwendzina (commented):
Changes look nice.

Would be good to get your thoughts on the additional messaging that Si has suggested, and how to manage messaging on the button when we move to not sending them straight to checkout when adding.

stevenbister:
Oo yeah I like that messaging. I think if we have that then changing the label on the button probably isn't necessary, we could just disable it and have this notice underneath it. If you leave out the "product is already in the cart" messaging then you could just show it as soon as it's added

Kinda like this maybe?

image

mattwendzina:
> Oo yeah I like that messaging. I think if we have that then changing the label on the button probably isn't necessary, we could just disable it and have this notice underneath it. If you leave out the "product is already in the cart" messaging then you could just show it as soon as it's added
> 
> Kinda like this maybe?
> 
> image

Really like having that messaging underneath. I'd still argue that changing to something like "Added" is still nice from a UX perspective as it just makes it clear that it's been added. The fact that it still says 'Add to Cart', even if disabled, still kind of gives the impression that it can still be added to the cart again 🤷

mattwendzina (approved):
LGTM

packages/osc-ecommerce/app/components/Cart/AlreadyInCartMessage.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export const AlreadyInCartMessage = () => {
+    return (
+        <Alert status="info" displayIcon={false}>
+            <p>
+                If you are looking to enrol multiple people on the same course, please{' '}
+                <Link to="/contact" className="u-text-underline">
+                    contact us
+                </Link>{' '}
+                to discuss the best offers.
```
hi-im-si, line 12: Update to:

You already have this course in your basket. If you are lookin**g to enrol multiple people onto the same course, please give us a call on 0121 329 4666 to discuss packages.**

hi-im-si (changes requested):

packages/osc-ecommerce/app/components/Cart/AlreadyInCartMessage.tsx
stevenbister: Done! 
image

hi-im-si (approved):

PR #934, opened 2023-05-24
Feature/remove from cart

### LINKED ISSUE #949: Remove Cart Items
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/948

### Description

n/a

stevenbister:
- Closes #949 

## 📝 Description

Adds the remove from cart functionality with some pending UI animations

N/a

## 🚀 New behavior

- Adds a `CartActions` subdirectory into `Forms` 
- Adds remove from cart function and mutation
- Adds remove from cart fetcher form
- Adds pending states to cart components when remove is in flight
- Adds `data-anim="shimmer"`
- Refactors `ProductForm` to use `data-anim`
- Adds an error message for if mutation returns with an error

## 📝 Additional Information

If you want to test this a few times without having to actually remove and add things to your cart over and over you can simulate the action by commenting out the `removeLinesFromCart` function in `routes/cart.tsx` and adding this below it like this:
```tsx
// result = await removeLinesFromCart({
//    cartId,
//    lineIds,
//    storefront,
// });

// cartId = result.cart.id;

const wait = () => new Promise((resolve) => setTimeout(resolve, 2000));
await wait().then(
    () => (result = {})
);
```
If you then want to simulate an error you can do this
```tsx
const wait = () => new Promise((resolve) => setTimeout(resolve, 2000));
await wait().then(
    () =>
        (result = {
            errors: [
                {
                    message: 'something went wrong!',
                },
            ],
        })
);
```


packages/osc-ecommerce/app/routes/cart.tsx
```diff
@@ -106,6 +107,23 @@ export const action: ActionFunction = async ({ request, context }: ActionArgs) =
...
+
+            invariant(lineIds.length, 'No lines to remove');
+
+            result = await removeLinesFromCart({
+                cartId,
+                lineIds,
+                storefront,
+            });
+
+            cartId = result.cart.id;
```
mattwendzina, line 123: Do we need optional operators here if result fails to return a result?

packages/osc-ecommerce/app/utils/cart.helpers.ts
```diff
@@ -167,3 +172,38 @@ export async function addLinesToCart(args: AddLines) {
...
+    const { cartLinesRemove } = await storefront.mutate<{
+        cartLinesRemove: {
+            cart: Cart;
+            errors: UserError[];
+        };
+    }>(REMOVE_LINE_ITEMS_MUTATION, {
+        variables: { cartId, lineIds },
+    });
+
+    invariant(cartLinesRemove, 'No data returned from remove lines mutation');
```
mattwendzina, line 206: I tried simulating this but commenting out the cartLinesRemoves code and just seeing it to undefined. 

The invariant throws an error, but it doesn't look like it gets handled anywhere so it just crashes the app 🤔

mattwendzina (commented):
Hi mate, looks good.
Left a couple of comments on the code in regards to error handling. Just thinking around what would happen if there was a bad connection with the storefront for example. 
Aside from that a couple of things from UI perspective:
- When an item is removed - I was initially thinking it would be nice to give them a warning modal to confirm, still kinda like this, but noticed that a lot of apps don't do this. But they do some other nice things. Amazon tells you the item you removed with a clickable link back to it. Gymshark has an undo option. I mean perhaps these things are nice to haves and definitely subjective. I always just think though that people can accidentally delete an item and nice to have an easy way back. But yeah, no super strong feelings on this. Just a thought.
- I think the remove loading is too subtle personally. I like the shimmer but it's not that easy to see, especially I'd imagine for someone hard of sight or in very bright conditions. I think we should have something additional that is clearer, and I think we should be disabling the remove button when they press it, in the same way that is happening with the enrol button. If you click it a few times it throws the error banner (but does still delete the item!)
- An aside I guess, but while I'm thinking about it 😅 Have similar views about the add to cart in terms of it being subtle and think perhaps something clearer might be helpful, say on the button itself. Also I think this needs to be disabled to stop users clicking and adding multiples (though I think this is already in a ticket..?) Again, subjective of course. Not super strong feelings on it, just thinking it may be helpful for people with less good sight.
-

packages/osc-ecommerce/app/routes/cart.tsx
```diff
@@ -106,6 +107,23 @@ export const action: ActionFunction = async ({ request, context }: ActionArgs) =
...
             break;
 
+        case CartAction.REMOVE_FROM_CART:
+            const lineIds = formData.get('linesIds')
+                ? (JSON.parse(String(formData.get('linesIds'))) as CartType['id'][])
+                : ([] as CartType['id'][]);
+
+            invariant(lineIds.length, 'No lines to remove');
+
+            result = await removeLinesFromCart({
```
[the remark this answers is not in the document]
stevenbister: Ooops yeah good point! Invariant and Remix are throwing and catching respectively but appreciate that the error message could be nicer for the user! Will update

packages/osc-ecommerce/app/routes/cart.tsx
stevenbister: No I don't think so in this case. I'll move the cartId bit into the catch so that should cover that, and Shopify will always return the [cart and error objects in the payload](https://shopify.dev/docs/api/storefront/2022-10/mutations/cartLinesRemove#returns) so they're safe to destructure. They actually do it like that in their demo store

packages/osc-ecommerce/app/utils/cart.helpers.ts
stevenbister: Related to above answer I guess

stevenbister:
> * When an item is removed - I was initially thinking it would be nice to give them a warning modal to confirm, still kinda like this, but noticed that a lot of apps don't do this. But they do some other nice things. Amazon tells you the item you removed with a clickable link back to it. Gymshark has an undo option. I mean perhaps these things are nice to haves and definitely subjective. I always just think though that people can accidentally delete an item and nice to have an easy way back. But yeah, no super strong feelings on this. Just a thought.

Yeah that's a good shout, we get the payload back from Shopify so should be able to do something 🤔 

> I think the remove loading is too subtle personally. I like the shimmer but it's not that easy to see, especially I'd imagine for someone hard of sight or in very bright conditions. I think we should have something additional that is clearer...

Any ideas? I struggled with coming up with something for this one, a spinner on the enrol style button would be the most obvious but I think when you're removing or updating an item it would be strange to see that button spinning on it's own. 
This also leads onto how would you indicate when we update an item in #935? I feel like the shimmer works well as you can target specific areas with it, i.e. shimmer the card button when removing an item, shimmer the prices when you add a discount code which helps to indicate "what" is changing if you follow me. Perhaps it just needs darkening or something to make it more obvious? But I don't want it to be too intrusive you know 🤔 

Agree with disabling the buttons though!

packages/osc-ecommerce/app/routes/cart.tsx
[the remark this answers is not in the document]
stevenbister: Have updated! Shopify will return more specific errors from the GraphQl response if there is an issue on their side of things but I've handled our invariant now 👍

stevenbister:
> When an item is removed - I was initially thinking it would be nice to give them a warning modal to confirm, still kinda like this, but noticed that a lot of apps don't do this. But they do some other nice things. Amazon tells you the item you removed with a clickable link back to it. Gymshark has an undo option. I mean perhaps these things are nice to haves and definitely subjective. I always just think though that people can accidentally delete an item and nice to have an easy way back. But yeah, no super strong feelings on this. Just a thought.

I've added a bit of state to the cart which will hold the line items when the cart is created and comparing that with the original line items. When the remove action is run the original line items array is updated and the one is state stays the same, I am using this to compare the two arrays and displaying a message if they are different.
It's a bit too awkward to try and put the removed item into the place it used to be in the array so I've just popped them at the top.

mattwendzina:
> > * When an item is removed - I was initially thinking it would be nice to give them a warning modal to confirm, still kinda like this, but noticed that a lot of apps don't do this. But they do some other nice things. Amazon tells you the item you removed with a clickable link back to it. Gymshark has an undo option. I mean perhaps these things are nice to haves and definitely subjective. I always just think though that people can accidentally delete an item and nice to have an easy way back. But yeah, no super strong feelings on this. Just a thought.
> 
> Yeah that's a good shout, we get the payload back from Shopify so should be able to do something 🤔
> 
> > I think the remove loading is too subtle personally. I like the shimmer but it's not that easy to see, especially I'd imagine for someone hard of sight or in very bright conditions. I think we should have something additional that is clearer...
> 
> Any ideas? I struggled with coming up with something for this one, a spinner on the enrol style button would be the most obvious but I think when you're removing or updating an item it would be strange to see that button spinning on it's own. This also leads onto how would you indicate when we update an item in #935? I feel like the shimmer works well as you can target specific areas with it, i.e. shimmer the card button when removing an item, shimmer the prices when you add a discount code which helps to indicate "what" is changing if you follow me. Perhaps it just needs darkening or something to make it more obvious? But I don't want it to be too intrusive you know 🤔
> 
> Agree with disabling the buttons though!

I would think something like a spinner next the actual remove button. I think slightly dark shimmer may work too. Having the disabled remove button definitely helps though. I'm not totally against just keeping it how it is, but wanted to at least have the conversation. May be worth @hi-im-si weighing in with his thoughts on it.

mattwendzina (commented):
Changes look great mate 🙌 

I've left one additional comment about the loading UI for removing items. Like I've said, not really strongly opposed to leaving how is, but thought might be worth getting @hi-im-si view on this too.

mattwendzina (approved):
LGTM!

PR #939, opened 2023-05-31
Feature/discounts

stevenbister:
- Closes: #951 
- Closes: #955 
- Closes: #952 
- Closes: #953 
- Closes: #957 

## 📝 Description

Adds a `DiscountBox` component to the cart which will allow users to enter their discount code. This will update the values in the cart and persist through to the checkout.

I ended up refactoring the `Price` component to allow it to be a bit more flexible. This way we can pass different money objects through. In this case I needed it to update the value in the `CartCard` when the discount is applied.

I deviated from the design/current functionality a bit by making the input box permanently visible, rather than showing it when you click in the title. I felt that this UI wasn't super clear that you could do that as there's no real indicator to say you can click on it.

If you enter an invalid discount code or a combination of codes that aren't applicable to the products in the cart/existing discount codes then it will show an info alert with a generic message. Unfortunately Shopify doesn't return much detail on why the discount is invalid, the returned object only looks like this:
```ts
"discountCodes": [
  {
    "applicable": false,
    "code": "fakecode"
  }
],
```
If you add two codes that exist, but aren't applicable together e.g. `TESTORDER50` and `COMBODISCOUNT1` (see additional info on when this are applicable) then Shopify will automatically apply the better discount and set the other one to invalid. In this case `COMBODISCOUNT1` would be the better discount so `TESTORDER50` would be removed.

I did want to write some unit tests for this one but I can't get it to render the component correctly because of the `fetcher.Form`. So I'll have to include these tests as part of the e2e tests most likely.

## 🚀 New behavior

- Refactors `Price` component in ecommerce to take more specific props, this let's us pass pricing from objects other than just the selected variant
- Adds `className` prop to `TextInput`
- Adds a `DiscountBox` component and stylesheet to ecommerce
- Adds update discount mutation and helper function to the cart

## 📝 Additional Information

I've created three discount codes in Shopify:
- `TESTORDER50` which will apply a blanked discount across the cart - doesn't work with other discount codes
- `COMBODISCOUNT1` will apply a 20% discount to the A Level Accounting product when in the cart
- `COMBODISCOUNT2` will apply a 5% discount to the AAT Level 2 and 3 Accounting product when in the cart, can be used in conjunction with `COMBODISCOUNT1`

Couple of things to note regarding styles:

- The styles on the `Alert` are still in progress so the icon looks tiny in that space at the moment.
- There is a small bug in the Tertiary `TextInput` if it is controlled. When you blur the component when there is still a value then the label shrinks back down and covers the input


packages/osc-ecommerce/app/components/Price/Price.tsx
```diff
@@ -3,40 +3,32 @@ import type { ProductVariant } from '@shopify/hydrogen/storefront-api-types';
...
 interface PriceProps {
-    selectedVariant: ProductVariant;
+    price: ProductVariant['price'];
+    compareAtPrice: ProductVariant['compareAtPrice'];
+    sku?: ProductVariant['sku'];
     isLoading?: boolean;
 }
 
 export const Price = (props: PriceProps) => {
-    const { selectedVariant, isLoading } = props;
```
mattwendzina, line 11: Do we want to remove/update the comment below that says `If selectedVariant...` as this no longer exists?

mattwendzina (commented):
Looks great mate 🙌 
Left tiny bit of specific bit of feedback, and just a couple of other UX type comments:

- In mobile view - The 'Apply' button seems to get stuck - you have to click somewhere a second time for it to lift. This is just testing in browser, so don't know if this is actually the behaviour on a mobile...
- Is it worth disabling the Apply button until a code is added? If you click it with nothing it still makes a network request which is a bit redundant.
- I did initially think we needed some pending UI on the Apply button, then after a while noticed there is the shimmer effect. It is really subtle lol. I think I'd advocate something on the button itself like 'applying...'  or a spinner or something to make it clearer. 
- One other small thing - When using the blanket discount, it doesn’t show the old total and new total price. With the other discounts it crosses out the old and shows the new on the product. Just thinking if someone has a few items they may not notice the discount if it’s only a small amount. I just think it would be a nice touch but isn’t huge so happy either way.

packages/osc-ecommerce/app/components/Price/Price.tsx
stevenbister: Ahh yeah good spot!

stevenbister:
> Looks great mate 🙌 Left tiny bit of specific bit of feedback, and just a couple of other UX type comments:
> 
> * In mobile view - The 'Apply' button seems to get stuck - you have to click somewhere a second time for it to lift. This is just testing in browser, so don't know if this is actually the behaviour on a mobile...

I think this is just default behaviour across all buttons. I guess when on mobile when you tap it the hover/active state gets applied

> * Is it worth disabling the Apply button until a code is added? If you click it with nothing it still makes a network request which is a bit redundant.

Yes! Good shout

> * I did initially think we needed some pending UI on the Apply button, then after a while noticed there is the shimmer effect. It is really subtle lol. I think I'd advocate something on the button itself like 'applying...'  or a spinner or something to make it clearer.

Yeah that's a fair point. I might handle all of these pending/loading state things together in a separate PR if that's cool with you? There's a few mentions across each of these branches so would be easier than trying to track them all.

mattwendzina:
> > Hi mate, Changes look good. Noticed a couple more things though soz!
> > 
> > 1. If you add the same discount code but with different casing it adds another discount code to the list. It still only applies on discount though thankfully 😅
> 
> Should be sorted, appears that discount codes are case insensitive so I've added a `toUpperCase` to the map to handle that.
> 
> > 2. This is a bit of an annoying one possibly but kind important I'd say... If you enter a code and hit the enter button it actually presses the Enrol Now button rather than the Apply button. Presumably because it's in a form.
> 
> Looks like it's just the pending disabling that button. I've added that to the discount form button as well so it's clearer that's doing something. I think might be worth tackling the pending ui (spinner or whatever) for that button too when we go through all of that.
> 
> > One other small thing - When using the blanket discount, it doesn’t show the old total and new total price. With the other discounts it crosses out the old and shows the new on the product. Just thinking if someone has a few items they may not notice the discount if it’s only a small amount. I just think it would be a nice touch but isn’t huge so happy either way.
> 
> I've added the strikethrough to the total when a cart level discount is applied, Shopify will only return the discounted difference for the entire cart depending on the type of discount. If it's product only discount then that gets applied to the line item, if it's a cart discount then that gets applied to the cart as a whole so the UI will looks slightly different depending on which is used. Spoke with Si & Nat and we've agreed that a user won't be able to apply product only discounts along with cart discounts so they won't see the difference (and if they do then the UI should handle that anyway).
> 
> Another thing that came out of that discussion is to make the discount box less obvious, so I've removed the background colour

The changes look great mate!

PR #940, opened 2023-05-31
Bugfix/text input

stevenbister:
## 📝 Description

Adds a `value` prop and sets the the `TextInput` component to be controlled internally so that the `c-label--filled` class is correctly applied when controlling the input with external state.

Currently when you blur or try to clear the state of a controlled input (tertiary variant) then the label drops back down but the input value persists, which leads to this obscured looking text.

https://github.com/Open-Study-College/osc/assets/23461173/f0bca1f5-763e-4d7c-9fed-52c74944c48d

## 🚀 New behavior

By passing a `_value` prop through to the inner component we can use that in conjunction with the inner `value` prop to control whether the `filled` variant gets set or not while also being able to control the state outside of the component.

https://github.com/Open-Study-College/osc/assets/23461173/290d7924-c2dd-4840-a833-7dec9fbc77dc

### DIFF
--- a/packages/osc-ui/src/components/TextInput/TextInput.spec.tsx
+++ b/packages/osc-ui/src/components/TextInput/TextInput.spec.tsx
@@ -1,6 +1,7 @@
 import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import React, { useState } from 'react';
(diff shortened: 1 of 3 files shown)


mattwendzina (approved):
Looks good mate 🙌

PR #941, opened 2023-06-01
Feature/cart buyer

stevenbister:
## 📝 Description

Adds the mutation and functions for updating the buyer id in the cart. This doesn't do anything as of yet, but will be used with the currency selector. I've removed the customer name/email details from the cart query as well as we're not doing anything with those.

N/a

## 🚀 New behavior

- Adds mutation to update the users currency

### DIFF
--- a/packages/osc-ecommerce/app/routes/cart.tsx
+++ b/packages/osc-ecommerce/app/routes/cart.tsx
@@ -27,6 +27,7 @@ import {
     addLinesToCart,
     createCart,
     removeLinesFromCart,
+    updateCartBuyerIdentity,
     updateCartDiscounts,
     updateLinesInCart,
 } from '~/utils/cart.helpers';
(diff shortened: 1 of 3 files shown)


packages/osc-ecommerce/app/routes/cart.tsx
```diff
@@ -106,6 +98,34 @@ export const action: ActionFunction = async ({ request, context }: ActionArgs) =
 
             break;
 
+        case CartAction.UPDATE_BUYER_IDENTITY:
+            const buyerIdentity = formData.get('buyerIdentity')
+                ? (JSON.parse(String(formData.get('buyerIdentity'))) as CartBuyerIdentityInput)
+                : ({} as CartBuyerIdentityInput);
+
+            result = cartId
```
mattwendzina, line 106: Do we need to wrap this in a try/catch in the same way we're now doing with CartAction.ADD_TO_CART?

stevenbister: Ah yeah good catch! Done

mattwendzina: No pun intended 😄

mattwendzina (approved):
LGTM

PR #954, opened 2023-06-05
Feature/checkbox size sm

stevenbister:
## 📝 Description

As part of the cart, the addons section uses a smaller than usual checkbox. Here I've added a "small" modifier to the checkbox component so we can use it there and anywhere else it might pop up.

- Size modifiers are only in m and xl

## 🚀 New behavior

- Adds small modifier to checkbox and label


mattwendzina (approved):
LGTM!

PR #1034, opened 2023-06-12
Feature/product cart pending states

stevenbister:
## 📝 Description

Attempts to make the pending states of the cart clearer to the user when they are updating the contents.
I've added a spinner to the buttons within the cart, so when they are in a pending state they will show a spinner as well as the shimmer. The opacity of the shimmer has been reduced to make it less subtle.

I've not added a spinner to the button in the product form as when the minicart gets added the `AddToCart` form will become a fetcher and I plan to use some optimistic UI here to add the item to the minicart before the transition has finished. Having a spinner on here when the cart opens will likely lead to the form looking out of sync. If it turns out the optimistic UI isn't the best behaviour then I'll add the spinner as part of that work.

I noticed that the discounted text size was larger than the total on mobile devices so have added a small fix for this in here also.

N/a

## 🚀 New behavior

- Adds a condition to the Button to absolutely position the spinner when there is no loading text, this stops the button from changing size as the spinner is much smaller than the text
- Adds spinners to cart buttons when cart is in a pending state
- Makes shimmer animation less transparent
- Tidies up a left over console log and duplicated prop

### DIFF
--- a/packages/osc-ecommerce/app/components/Cart/DiscountBox/DiscountBox.tsx
+++ b/packages/osc-ecommerce/app/components/Cart/DiscountBox/DiscountBox.tsx
@@ -109,7 +109,6 @@ export const DiscountBox = (props: DiscountBoxProps) => {
                             variant="primary"
                             size="sm"
                             isLoading={pendingFetcher}
-                            loadingText=""
                             isDisabled={pendingFetcher || !inputValue ? true : false}
                             tabIndex={active ? 0 : -1}
                         >
--- a/packages/osc-ecommerce/app/components/Cart/DiscountBox/DiscountBox.tsx
+++ b/packages/osc-ecommerce/app/components/Cart/DiscountBox/DiscountBox.tsx
@@ -108,6 +108,8 @@ export const DiscountBox = (props: DiscountBoxProps) => {
                         <Button
                             variant="primary"
                             size="sm"
+                            isLoading={pendingFetcher}
+                            loadingText=""
                             isDisabled={pendingFetcher || !inputValue ? true : false}
                             tabIndex={active ? 0 : -1}
                         >
--- a/packages/osc-ecommerce/app/components/Cart/Layout.tsx
+++ b/packages/osc-ecommerce/app/components/Cart/Layout.tsx
@@ -188,7 +188,6 @@ export const CartLayout = () => {
                                         isFull
                                         isDisabled={linesArePending || discountCodeIsPending}
                                         isLoading={linesArePending || discountCodeIsPending}
-                                        loadingText=""
                                     >
                                         Enrol now
                                     </Button>
--- a/packages/osc-ecommerce/app/components/Cart/Layout.tsx
+++ b/packages/osc-ecommerce/app/components/Cart/Layout.tsx
@@ -187,6 +187,8 @@ export const CartLayout = () => {
                                         href={cart.checkoutUrl}
                                         isFull
                                         isDisabled={linesArePending || discountCodeIsPending}
+                                        isLoading={linesArePending || discountCodeIsPending}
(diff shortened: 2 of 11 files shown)


packages/osc-ecommerce/app/components/Cart/DiscountBox/DiscountBox.tsx
```diff
@@ -108,6 +108,8 @@ export const DiscountBox = (props: DiscountBoxProps) => {
                         <Button
                             variant="primary"
                             size="sm"
+                            isLoading={pendingFetcher}
+                            loadingText=""
```
mattwendzina, line 112: Is there a reason `loadingText` is just an empty string?

packages/osc-ecommerce/app/components/Cart/Layout.tsx
```diff
@@ -187,6 +187,8 @@ export const CartLayout = () => {
                                         href={cart.checkoutUrl}
                                         isFull
                                         isDisabled={linesArePending || discountCodeIsPending}
+                                        isLoading={linesArePending || discountCodeIsPending}
+                                        loadingText=""
```
mattwendzina, line 191: Same as comment above

packages/osc-ui/src/components/Button/Button.tsx
```diff
@@ -164,14 +163,17 @@ export const Button = forwardRef<typeof HTMLElement, ButtonProps>(
...
         const buttonInner = isLoading ? (
             <span className="c-btn__inner">
                 {loadingText && loadingText}{' '}
-                <span className="c-btn-loader">
+                <span className={`c-btn-loader ${!loadingText ? 'c-btn-loader--absolute' : ''}`}>
                     <span className="c-btn-loader__dot"></span>
                     <span className="c-btn-loader__dot"></span>
                     <span className="c-btn-loader__dot"></span>
                 </span>
+                {!loadingText && <span className="u-vis-hidden">{children}</span>}
```
mattwendzina, line 176: I was about to comment that `loadingText` has a default of `Loading` so when could it be false. Which was then answered with my last comment about passing an empty string. Doesn't seem like the most intuitive way to have no loadingText - ie having to pass in an empty string for it to coerce to falsy. Not sure I have an alternative to offer though lol, and maybe you've already thought about this and couldn't think of a more intuitive way to handle it 🤷

mattwendzina:
Sorry, this comment isn't directly related to this PR so doesn't need addressing here, but I noticed it when doing testing on the cart. The alert message has a different size X and icon in each of the removal messages - presumably related to the amount of text? Don't know how quick/easy this is to adjust and whether it's best to do in a separate ticket?

Screenshot 2023-06-13 at 18 35 00

mattwendzina (commented):
Really like the look of this now mate. Left a couple of small comments, but other than that all looks good to me 👌

packages/osc-ui/src/components/Button/Button.tsx
stevenbister: Yeah that's a fair point. I did think about and I was on the fence about changing it but decided to leave it. I think I agree it is unintuitive. I'll get it updated 👍

stevenbister:
> Sorry, this comment isn't directly related to this PR so doesn't need addressing here, but I noticed it when doing testing on the cart. The alert message has a different size X and icon in each of the removal messages - presumably related to the amount of text? Don't know how quick/easy this is to adjust and whether it's best to do in a separate ticket?

Yeah I noticed that too, and the i icon is as well it's a task (https://github.com/Open-Study-College/osc/issues/924) assigned to @hi-im-si to update the Alerts so I've left it as it in this case

mattwendzina (approved):
LGTM!

PR #1108, opened 2023-06-12
Feature/cart add ons

### LINKED ISSUE #958: [S67.1]: Add Ons
### Link to Epic

### Description

_No response_

### Linked Tasks

- [x] #959 
- [ ] #960 
- [ ] #961

### LINKED ISSUE #960: Integrate add ons with cart
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/958

### Description

n/a

### LINKED ISSUE #961: Pull Addon Content from CMS
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/958

### Description

n/a

stevenbister:
- Closes #958 
- Closes #960 
- Closes #961 

## 📝 Description

Adds the addons section to the cart.

Addons are a type of product within Shopify identified by the "Add on" product type.
image

These are intended to work slightly differently to regular products in Shopify and should only be accessible from the cart when applicable products are added. To achieve this I've added a condition in the courses pages which will throw a 404 if someone tries to access the add on product url.

Certain addons are not available to certain products, for example course replacement cover is not available for AAT courses. To handle this I've added a custom metafield to Shopify called `exclusion_list` which houses the SKU prefix for the type of courses that addon should be excluded from. Unlike the current site setup where this is handled and updated directly in the database, making this controllable within Shopify will let us change these without having to do any releases.
image

Similarly course replacement cover is only available for the "study pack" option, so that addon will be removed from the cart when that option is selected.

I've setup the update cart query to clear _all_ addons when it is run for simplicity, rather than trying to hold onto whichever is selected and decide in the query whether it should be removed or not. So if a user changes the variant in their cart then they will need to reselect their addon. I don't imagine people changing their variants in the cart regularly enough for this to be a massive issue. We will be adding this into the checkout as well later down the line so there will be another chance for them to select it again if they miss it.

N/a

## 🚀 New behavior

- Adds `CartAddOns` component
- Throw 404 if trying to access add on product page
- Updates `updateLinesInCart` function to remove addons from cart when fired
- Updates the `Checkbox` component so that it can be controlled
- Filter addons from `cartLineItems` state so we don't show the removed message when one is un-checked

## 📝 Additional Information

I think the design intends the info icon to show a tooltip when hovered over, however we don't have this component built yet so I've used a Popover for now which means you'll need to click on the icon instead.

One more thing to note, with regards to this, rather than house the description in Sanity I've left it in Shopify. To me as these aren't full fledged products content management won't really be a thing for them. To me it feels like it makes sense to keep all of the controls for it within Shopify (I also don't imagine they will be updated regularly). Happy to discuss this and potentially change it around however.


mattwendzina:
Looking good mate 👍  Couple of things:

When there are multiple addons, might be nice if the 'i' icon lined up on them all? And perhaps the checkbox be aligned to centre when it goes over multiple lines? (That's probably my bad though when I made the component! 🤦 ) Not strong opinions on these, just think it would look a bit neater.

Screenshot 2023-06-14 at 07 42 02

Looks like there is an issue with displaying all the correct add ons when there are multiple items in the cart. See screenshots below. When there is a CACHE course and an A-Level there is only an option for the Career kickstart package. If you remove the CACHE course you then get the options for the course replacement cover and replacement cover/career kickstart. Same behaviour if you have two of the same award, e.g. two A-Levels. In this case I would assume there would need to be the option to select Course replacement cover for each course, unless one set of cover covers all the courses..?

Screenshot 2023-06-14 at 07 46 32

Screenshot 2023-06-14 at 07 46 55

stevenbister:
After a few discussions with Si and the leadership team it's been decided that addons will work as stand alone products rather than going against the line item. This is because it works like insurance so you can take out x amount and that gives you x uses against any qualifying product.

As per request I've added a quantity picker to the addon which shows when it is selected. I've kept the original behaviour in that the addons will reset/clear when updating or removing a course from the cart for simplicity as it's difficult to track; what product has been removed and if it qualifies for the addon while also taking into account updating the quantity.

mattwendzina:
Looks good mate. Just one comment about the 'Course Replacement Cover' tooltip - The text probably needs updating to reflect the fact that you can purchase multiple. Guess this is probably content would be marketing controlled though anyway?
Screenshot 2023-07-11 at 18 14 57

mattwendzina (commented):
One other small comment mate, but other than that, looking good!

stevenbister:
Thanks mate,

> Looks good mate. Just one comment about the 'Course Replacement Cover' tooltip - The text probably needs updating to reflect the fact that you can purchase multiple. Guess this is probably content would be marketing controlled though anyway?

Yeah that would probably be a marketing thing to do I think

mattwendzina (approved):
LGTM

PR #1134, opened 2023-06-13
Feature/pagination component

### LINKED ISSUE #171: [EP4]: Build Components
### Description

N/A

mattwendzina:
Closes #171 

## 📝 Description

Adds Pagination component to OSC-UI library

N/A

## 🚀 New behavior

Adds a Pagination component

No


stevenbister:
Oh something else I just noticed, Chromatic has failed. Looks like the progress component is missing from the package-lock file?
image
Might just need to push that up

packages/osc-ui/src/components/Pagination/Pagination.spec.tsx
stevenbister, line 1: Do you think it's worth including a test for the `onPaginate` prop?

packages/osc-ui/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+const ProgressDetails = (props: ProgressDetailsProps) => {
+    const { numberLoaded, total, itemTypeDescription } = props;
+
+    return (
+        <div className="c-pagination__progress-details">
+            <span> You've viewed </span>
+            {''}
+            <span>
```
stevenbister, line 89: Do we need all of these `span`s? I think this would be okay
`<span> You've viewed {numberLoaded} of {total} {itemTypeDescription}</span>`

packages/osc-ui/src/components/Pagination/pagination.scss
```diff
@@ -0,0 +1,15 @@
+@import "../../styles/settings";
+@import "../../styles/tools";
+
+.c-pagination {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+
+    &__button {
```
stevenbister, line 10: I think this button looks like it needs more work to match the design (too much letter spacing, font weight too heavy for example)

Si and I discussed consolidating the buttons styles when the component was built so rather than trying to create a specific button element here we should just use the existing ones (tertiary variant might be a better fit than secondary 🤔) Appreciate the design doesn't show that.

packages/osc-ui/src/components/Pagination/pagination.scss
stevenbister, line 1: Missing a few styles I think, from looking at the design:

image
image

The font weight is too heavy and we need to include some spacing between the elements

stevenbister (changes requested):
Looks good! Left a couple of questions and I think the styling needs some tweaks but nothing major :)

packages/osc-ui/src/components/Pagination/pagination.scss
mattwendzina: Secondary and tertiary both have uniform border and font colours, so either way it seemed like I'd have to change the font or border colour, unless I'm missing something? Couldn't see a way to change text to lower case in the component which is why I added this in here, again maybe there is a way that I'm missing? 😅 
Tbh, didn't notice the font weight difference. I think it's marginal, but I can see it now you mention it. I was going with the lightest that we had in the system - `fw-reg`, so would you recommend just creating a lighter variety?

packages/osc-ui/src/components/Pagination/pagination.scss
mattwendzina: Ah yeah that's my bad on the spacing. It actually was there before lol, so I don't know what I did to remove it 😅 but that's fixed. Sorted out the font-weights too.

packages/osc-ui/src/components/Pagination/pagination.scss
mattwendzina: Scrap the above comment lol. Si and I chatted about it and now understand what you meant - ie just use one of our designs rather than their design in Figma! Done - using the `Tertiary` button.

stevenbister (commented):
Looks really good mate, only thought I've had how are we handling changing the colour of the progress bar within the component? It looks like the progress is abstracted away as an internal component so AFAICT there's no way to pass props through to it right now? Perhaps just needs a `colorVariant` prop here as well?

mattwendzina:
> Looks really good mate, only thought I've had how are we handling changing the colour of the progress bar within the component? It looks like the progress is abstracted away as an internal component so AFAICT there's no way to pass props through to it right now? Perhaps just needs a `colorVariant` prop here as well?

Ah yeah, good shout! Added ✅

stevenbister (approved):
🚀

PR #1136, opened 2023-06-13
Feature/add infinite hits to instantsearch

### LINKED ISSUE #1135: Add InfiniteHits to Search page
### Link task to story

Story: #196

### Description

Add InfiniteHits component to Search page and integrate Pagination component

mattwendzina:
Closes #1135 

## 📝 Description

Integrates the <InfiniteHits> component into the Search page and adds the Pagination component.

## 🚀 New behavior

Currently there is no pagination

No

## 📝 Additional Information

We are integrating the `useInfiniteHits` hook rather than the `usePagination` hook because we want to load results cumulatively rather than one a page by page basis. This requires integrating the <Hit> component into the <InfiniteHits> component and this effectively replaces the <Hits> component which has been removed.


packages/osc-ecommerce/app/components/InstantSearch/Widgets/InfiniteHits.tsx
```diff
@@ -0,0 +1,62 @@
...
+                ))}
+            </NoResultsBoundary>
+            {results?.nbHits !== 0 ? (
+                <OSCPagination
+                    itemTypeDescription="courses"
+                    total={results?.nbHits}
+                    numberLoaded={hits.length}
+                    onPaginate={onPaginateHandler}
+                    isLoading={isLoading}
+                    className="o-grid o-grid__col--12 o-container o-container--4xs"
```
stevenbister, line 57: I think this looks too wide comparing it to the design
image
image

I think rather than trying to control it with the grid and a container in this case you might be better off flexing it and setting a max width. This gets you pretty close
```tsx
<div className="o-flex o-flex--center o-grid__col--12">
    <OSCPagination
        itemTypeDescription="courses"
        total={results?.nbHits}
        numberLoaded={hits.length}
        onPaginate={onPaginateHandler}
        isLoading={isLoading}
    />
</div>
```

packages/osc-ecommerce/app/components/InstantSearch/instant-search.scss
```diff
@@ -31,4 +31,10 @@
         display: flex;
         max-width: 30ch;
     }
+
+    .c-pagination {
+        @include mq($mq-tab) {
+            padding-inline: $space-7xl;
```
stevenbister, line 37: Not sure if I'm looking in the wrong place but I don't think this is doing anything?

stevenbister:
Oh also you might want to add the `build:ecommerce` label to this to build the preview link 😎

packages/osc-ecommerce/app/components/InstantSearch/instant-search.scss
mattwendzina: This was being used but is no longer needed, so it's gone!

packages/osc-ecommerce/app/components/InstantSearch/Widgets/InfiniteHits.tsx
mattwendzina: Had a chat with Si on this. We ended up adding two more container sizes - `--xs5` and `--xs6`. So this is now using `--xs6`, which is a tiny bit wider than the design, but it means we keep the 'You've viewed...' text all on one line when we're into the 100's of courses.

mattwendzina:
> Hey man, pretty cool that! Couple of comments around styling hopefully nothing too major.
> 
> Only other comment I have is the component looks a bit too close to the top and bottom especially on desktop
> 
> image image

Ah yeah, good spot. Added some more spacing 👍

packages/osc-ecommerce/app/components/InstantSearch/Widgets/InfiniteHits.tsx
stevenbister: Ahh nice yeah that makes sense! You might need to take the padding utility off now though :S I would maybe look at adding the `o-container--flush` class just to trim off the container's padding as well

image

Just on this, I would take the `o-grid` class off here too. You don't really need it here and the `gap` that comes with it is adding some extra space between each item which is adding to the margin

packages/osc-ecommerce/app/components/InstantSearch/instant-search.scss
```diff
@@ -31,4 +31,11 @@
         display: flex;
         max-width: 30ch;
     }
+
+    /* stylelint-disable-next-line plugin/selector-bem-pattern */
+    .c-pagination {
+        gap: 0;
+        padding: 0;
```
stevenbister, line 38: Rather than adding padding and gap here I would take off the `o-grid` and `u-p` utility classes you have on the component

stevenbister (changes requested):
Looks good! Couple of little comments but should be gtg once those are sorted 👍

packages/osc-ecommerce/app/components/InstantSearch/instant-search.scss
mattwendzina: Updated

packages/osc-ecommerce/app/components/InstantSearch/Widgets/InfiniteHits.tsx
mattwendzina: Updated 👍

stevenbister (approved):
🚢

PR #1138, opened 2023-06-13
Feature/algolia instantsearch e2e tests v2

### LINKED ISSUE #1137: Add E2E testing for Search page
### Link task to story

Story: #196

### Description

Adds e2e test for the /search route.

mattwendzina:
Closes #1137 

## 📝 Description

Adds e2e tests for the /search route

N/A

## 🚀 New behavior

E2E tests for /search route

No

## 📝 Additional Information

Had to do a couple of things to get this all working:
1) Add `--disable-web-security` in the `launchOptions` config, to bypass CORS issue when testing 3rd party API
2) When Searching for a course the Hits don't automatically update. I found that doing a sort forced an update which then enabled me to test what needed to be tested. E.g. that the correct number of courses are shown.

One other note:
- These have the potential to be quite brittle unless we have a test index which is used solely for the purpose of testing. This is because the tests test the actual data in the index and if this got changed then it would cause the tests to fail.

### DIFF
--- a/packages/osc-ecommerce/e2e/search.spec.ts
+++ b/packages/osc-ecommerce/e2e/search.spec.ts
@@ -1,0 +1,101 @@
+import { expect, test } from '@playwright/test';
+
+test.beforeEach(async ({ page }) => {
+    await page.goto(`/search`);
+});
+test('should return filtered hits based on query in searchbox ', async ({ page }) => {
+    await page.fill('input[type="search"]', 'Biology');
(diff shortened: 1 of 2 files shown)


stevenbister (approved):
Looks good! Yeah I agree I think we should look at setting up a "test" index that doesn't change. Could we create a one off based on the current staging Shopify store? Might then need to update the CI workflows slightly 🤔 But happy for this to go in as is now 👍

PR #1140, opened 2023-06-15
Feature/modal footer

stevenbister:
## 📝 Description

Adds a footer component to the `Modal` which we can use to position and style for the mini cart. I've updated the `PositionedTemplate` story to closer match the design of the mini cart which includes a sticky footer and header.

Something to note, is that you can't scroll all the way to the bottom of the modal when on screens shorted than about 570px. I'm not too fussed about this as on the website I'll swap this out for a link directly to the cart instead when on mobile as this modal seems a bit too small to interact with nicely on mobile devices.

N/a

## 🚀 New behavior

- `ModalFooter` component added
- Adds `isSticky` prop to `ModalHeader` and `ModalFooter`


mattwendzina (approved):
LGTM

PR #1141, opened 2023-06-15
Feature/mini cart

### LINKED ISSUE #966: Update bag icon in header with number of items when item is added
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/965

### Description

n/a

### LINKED ISSUE #967: Add mini cart modal component
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/965

### Description

n/a

### LINKED ISSUE #968: Add components to minicart modal
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/965

### Description

Line items
Total
View bag button
Enrol now button

### LINKED ISSUE #969: Update Product form to push to and open minicart
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/965

### Description

n/a

stevenbister:
- Closes #966 
- Closes #967 
- Closes #968 
- Closes #969 

## 📝 Description

Adds the `MiniCart` component to the ecommerce store. 

https://github.com/Open-Study-College/osc/assets/23461173/22755fe4-08d9-4970-8324-0b0994a49bdc

The `MiniCart` uses a slight variation on the logic in the `Cart`. We need to make more use of `useState` and `useEffect`s to handle a few parts of the behaviour as we aren't able to rely on the browser quite as much to set or clear state when the page is loaded. But is otherwise a trimmed down version.

I've opted to keep the modal closed when adding an item, rather than opening it like the Hydrogen demo store does. Instead I've added a spinner to the product form to indicate that something is happening. I think this could be a bit more whimsical, something like this could be nice: https://codepen.io/aaroniker/pen/pogGgbx.
Might do this as a separate PR but let me know what you guys think?

The main reason for this is because I don't really like the way the demo store works by opening the modal then "popping in" the cart content, to me this feels a bit janky. 

https://github.com/Open-Study-College/osc/assets/23461173/b4096302-52c3-4abb-8d7a-9746168cd0b4

Ours would either end up working in the same way as we need to wait for Shopify to generate the cart and that takes a second or so for the data to come back (hence the pop in). Or we wait for the data to return and then open the modal, but the timing would feel off and make the site feel slow.

I've also opted to not display it when on mobile sized devices, instead swapping it out for a link to the `/cart` route. I think the UX for this minicart on mobile isn't as good; it seems a bit small and there's too much opportunity to accidentally close it if part of your hand touches outside of the modal. Similarly if there are a number of items in the cart there is the chance that it will overflow the screen and is difficult to scroll to the cart/enrol buttons.

N/a

## 🚀 New behavior

- Updates `ProductForm` to `useFetcher` form
- Adds a `size` prop to the `CartCardItem` to adjust the size and spacing
- Adds ability to pass classnames through to the `CartTotal`
- Adds `MiniCart` component
- Adds `MiniCart` to header
- Adds `CartBadge` component

## 📝 Additional Information

~This doesn't include the badge on the cart icon in the header. I want to do that separately from this as I want to look at including some animations and felt that this already had a fair bit in it.~

I do have a task for tests as well but will tackle this separately


packages/osc-ecommerce/app/components/Cart/CartCardItem.tsx
```diff
@@ -66,34 +71,38 @@ export const CartCardItem = (props: CartCardItemProps) => {
...
+                                        </AccordionPanel>
+                                    </AccordionItem>
+                                </Accordion>
+                            )}
+                        </CardBody>
+                    ) : null}
+
+                    <CardFooter
+                        className={`${size === 'lg' ? 'u-pt-xl' : 'u-pt-0 o-flex--v-center'}`}
+                    >
```
mattwendzina, line 105: Can't comment on the line because it already exists, but just noticed it. Line 116 - Price component that has been commented out. Does this still need to be here?

packages/osc-ecommerce/app/components/Cart/CartBadge/CartBadge.tsx
```diff
@@ -0,0 +1,11 @@
+interface CartBadgeProps {
+    totalItems: number;
+}
+
+export const CartBadge = (props: CartBadgeProps) => {
```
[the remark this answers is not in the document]
stevenbister: Ahh! No forgot to update the PR, I played around with a few things but ended up not really liking anything because of the time it takes for the cart to update. So decided to keep this in here instead and leave the interaction with the add to cart form.

packages/osc-ecommerce/app/components/Cart/CartCardItem.tsx
stevenbister: Ooops no doesn't look like that needs to be in here

packages/osc-ecommerce/app/components/MiniCart/MiniCart.tsx
```diff
@@ -0,0 +1,269 @@
...
+                                as="link"
+                                to={`/${PATHS.COLLECTIONS}`}
+                                variant="secondary"
+                                isFull
+                            >
+                                Browse our courses
+                            </Button>
+                        )}
+
+                        {linesCount ? (
```
mattwendzina, line 232: Is this needed, or can it be consolidated into the above `linesCount` check on line 217?

packages/osc-ecommerce/app/components/MiniCart/MiniCart.tsx
```diff
@@ -0,0 +1,269 @@
...
+    const linesArePending = pendingLineIds.length > 0;
+
+    const fetchersWithErrors = allFetchers.filter((f) => fetcherHasError(f));
+
+    const cartLines = useMemo(
+        () => (linesCount > 0 && cart?.lines ? flattenConnection(cart?.lines) : []),
+        [cart?.lines, linesCount]
+    );
+    // Store the cart lines in state so we can compare them to the cart lines in the cart object when the component updates
+    const [cartLineItems, setCartLineItems] = useState(cartLines);
```
mattwendzina, line 74: As far as I can tell `cartLine` and `cartLineItems` are the same other than the latter is what is in local state and gets synced with `lineItems` when things change. It's hard to know the difference between them based on the naming here. Wonder if something like `locallyStoredLineItems` would make it a bit more clear what the difference is between them. Obviously there is the comment so it's not completely vague.

mattwendzina: One other thought, again from a consistency perspective. You currently have:
- cartLines, cartLineItems, removedItems, removedItemsState
May be nice if these all used the same naming convention?
e.g.
- cartLines, localCartLines, removedCartLines, removedLocalCartLines

Not a massive deal but just thoughts in terms of making it a bit easier to digest.

mattwendzina (commented):
This is looking great mate! Left a few comments. 

One other thing I've noticed which is a bit weird but not related to this PR so feel free to ignore. If you go to the cart and add a global discount, e.g. `TESTORDER50` it just changes the subtotal. If you hit Enrol Now and then go back to the checkout or just back to the website in general it then adds the individual level of discount to each line item which looks a bit messy. Not really a big deal as won't be a common scenario but if it's an easy fix then may be worth looking at at some point. Before/after screenshots below:

BEFORE
Screenshot 2023-06-19 at 15 42 41

AFTER
Screenshot 2023-06-19 at 15 42 54

mattwendzina:
In terms of behaviour when adding item to the cart. I like how you've done it personally. The more whimsical approach I'd be open to for sure. That one you showed is very cool. Only thing is that there isn't really anything 'whimsical' anywhere else on the website so would be wary it could feel out of place if it's too out there..!

mattwendzina:
Sorry one more thing, and apologies again as it isn't directly related so wouldn't need to be addressed here I guess. We added in the behaviour to have the 'You already have this course...' banner in once an item has been added. It feels weird that this comes up as soon as you add an item. It kind of makes it feel like you had already added the item. I think ideally this wouldn't show up when you initially add the course but only if you revisit the page.

packages/osc-ecommerce/app/components/MiniCart/MiniCart.tsx
stevenbister: Ooo yeah good catch, that defo makes sense. Will get those updated

stevenbister:
> One other thing I've noticed which is a bit weird but not related to this PR so feel free to ignore. If you go to the cart and add a global discount, e.g. `TESTORDER50` it just changes the subtotal. If you hit Enrol Now and then go back to the checkout or just back to the website in general it then adds the individual level of discount to each line item which looks a bit messy. Not really a big deal as won't be a common scenario but if it's an easy fix then may be worth looking at at some point. Before/after screenshots below:

Ooo that's quirky, I'll have a look but not sure if that's something we can do anything about from memory 🤔 The storefront API is giving us all of those values so should be updating them all when in the cart I would have thought! But will double check, will probs do it as a separate PR

stevenbister:
> Sorry one more thing, and apologies again as it isn't directly related so wouldn't need to be addressed here I guess. We added in the behaviour to have the 'You already have this course...' banner in once an item has been added. It feels weird that this comes up as soon as you add an item. It kind of makes it feel like you had already added the item. I think ideally this wouldn't show up when you initially add the course but only if you revisit the page.

Yes I was thinking about this! I think it would actually be best as a tooltip, so you only see it when you hover over the button again or something like that 🤔 

I get what you're saying, having it only show up when you re-visit the page, but I think UX wise it might seem odd that the button stays disabled after adding it to the cart without some indication as to why.

mattwendzina:
> > Sorry one more thing, and apologies again as it isn't directly related so wouldn't need to be addressed here I guess. We added in the behaviour to have the 'You already have this course...' banner in once an item has been added. It feels weird that this comes up as soon as you add an item. It kind of makes it feel like you had already added the item. I think ideally this wouldn't show up when you initially add the course but only if you revisit the page.
> 
> Yes I was thinking about this! I think it would actually be best as a tooltip, so you only see it when you hover over the button again or something like that 🤔
> 
> I get what you're saying, having it only show up when you re-visit the page, but I think UX wise it might seem odd that the button stays disabled after adding it to the cart without some indication as to why.

I think the tooltip approach would be good tbh. A couple more thoughts though. I just had a look at Udemy and they actually update the button to 'Go to Cart' which might actually be a nicer approach. Makes it quicker/simpler for someone to then go directly to the cart if they only want one course, which I'm guessing is probably a lot of people?
The other thought is that we could remove 'You already have this course in your basket' from the alert as this is obvious, certainly when you first click it is. At that point they just need to know that they can talk to us if they want multiple.

mattwendzina:
Changes look great mate 🙌 

Couple more things I've noticed, sorry! See screen video below. When on the checkout the 'View my bag' button is still visible, but doesn't do anything when you click it as it's already on the cart page. Do you think we should hide this button when you're on the cart?

Secondly, maybe being over picky, but not sure about interchanging between 'Cart' and 'Bag'..? When adding it's 'Add to Cart', when added it's 'View my Bag'. Not particularly precious about this but thought I'd see what you think.


https://github.com/Open-Study-College/osc/assets/36510083/c2426b27-b426-4792-bfa5-300db7eea588

stevenbister:
> Couple more things I've noticed, sorry! See screen video below. When on the checkout the 'View my bag' button is still visible, but doesn't do anything when you click it as it's already on the cart page. Do you think we should hide this button when you're on the cart?

Ahh yeah I bet that's because the path isn't changing so the modal isn't closing! Probs a good idea to hide it on the cart route

> Secondly, maybe being over picky, but not sure about interchanging between 'Cart' and 'Bag'..? When adding it's 'Add to Cart', when added it's 'View my Bag'. Not particularly precious about this but thought I'd see what you think.

Ah yeah that's my bad, it should be "Bag" across everything not "Cart"

mattwendzina (approved):
LGTM! 🚢

hi-im-si:
> I've opted to keep the modal closed when adding an item, rather than opening it like the Hydrogen demo store does. Instead I've added a spinner to the product form to indicate that something is happening. I think this could be a bit more whimsical, something like this could be nice: https://codepen.io/aaroniker/pen/pogGgbx.
> Might do this as a separate PR but let me know what you guys think?

I really like this and probably ties in with a future PR of adding animations to the store. definitely keep this one for future reference though. I also think by adding the animation it gives the users attention to something whilst the content updates the mini cart allowing for it to be revealed in its completed state rather than the jankiness of the hydrogen demo store.

PR #1142, opened 2023-06-15
Refactor/Close nav on navigation change

stevenbister:
## 📝 Description

Not an urgent one / related to any sprint work but more of a quality of life change. I discovered `useLocation` hook which updates the pathname when a navigation occurs. We can use this to change the state of the navigation menu rather than watching for click events, which always felt like a bit of a workaround to me.

Should make the UX a bit more reliable and more like how a site should actually handle closing navigation menus.

- Nav closes when a click occurs on a link

## 🚀 New behavior

- Nav closes when a navigation occurs


mattwendzina (approved):
Noice! LGTM.

PR #1165, opened 2023-06-19
Feature/cart e2e tests

### LINKED ISSUE #971: [S70.1]: Tests
### Link to Epic

### Description

n/a

### Linked Tasks

- [ ] #972 
- [ ] 
- [ ]

### LINKED ISSUE #972: End to End testing
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/971

### Description

Add e2e tests for cart
- [x] Should be able to adjust product variants
- [x] Should be able remove products
- [x] Should be able to add and remove discount codes
- [ ] Should be able to add addons
- [x] Should take user though to checkout

stevenbister:
- Closes #971 
- Closes #972 

## 📝 Description

Updates the product e2e tests to account for new minicart behaviour and adds the e2e tests for the cart 

N/a

## 🚀 New behavior

- Updates product e2e tests to account for fetcher.Form
- Adds cart e2e tests
- Adds accessible label to the remove discount button (both for being able to target in test and general good practice)
- Adds id to cart total component (for being able to better target in tests)

## 📝 Additional Information

Note this doesn't include the minicart tests, that will come separately from this. Also the addons aren't here yet so the tests for those will need to be included at a later date once the behaviour for that PR is locked down.


mattwendzina (commented):
Looks good mate. It did fail one test for me when trying to add to cart. Not sure what the configuration looks like with Shopify for the testing but are we mocking this out or calling the real endpoints? I wonder, if it's the latter, whether we should mock this functions/responses given that we just want to test the all the triggers like moving to cart, removing an item work etc... Just thinking it might make the tests a bit flakey and reliant on network requests otherwise? 
I say this knowing full well that the Algolia tests are doing exactly this 😅 I think with those, the entire behaviour is api driven so it's probably more difficult 🤷

stevenbister:
> Looks good mate. It did fail one test for me when trying to add to cart. Not sure what the configuration looks like with Shopify for the testing but are we mocking this out or calling the real endpoints? I wonder, if it's the latter, whether we should mock this functions/responses given that we just want to test the all the triggers like moving to cart, removing an item work etc... Just thinking it might make the tests a bit flakey and reliant on network requests otherwise?
I say this knowing full well that the Algolia tests are doing exactly this 😅 I think with those, the entire behaviour is api driven so it's probably more difficult 🤷

Okay so spent probably far too long on this now 😅 We probably can mock out the responses but we'd need to use MSW as the requests happen on the server rather than in the browser where you would normally use Playwright to do it. I managed to get it sort of working, however because of the way MSW works this means that we can't change the mock on a per test basis. This is fine until you get to the tests where we're firing mutations (e.g. updating the product option in the cart), the original mock query ends up getting fired again after the mutation so the test fails the values don't get updated.

I'm conscious of time on the project so I suggest we live with the flakey tests for the time being and perhaps readdress this when we've got more time to invest into researching / setting this up.

mattwendzina:
> > Looks good mate. It did fail one test for me when trying to add to cart. Not sure what the configuration looks like with Shopify for the testing but are we mocking this out or calling the real endpoints? I wonder, if it's the latter, whether we should mock this functions/responses given that we just want to test the all the triggers like moving to cart, removing an item work etc... Just thinking it might make the tests a bit flakey and reliant on network requests otherwise?
> > I say this knowing full well that the Algolia tests are doing exactly this 😅 I think with those, the entire behaviour is api driven so it's probably more difficult 🤷
> 
> Okay so spent probably far too long on this now 😅 We probably can mock out the responses but we'd need to use MSW as the requests happen on the server rather than in the browser where you would normally use Playwright to do it. I managed to get it sort of working, however because of the way MSW works this means that we can't change the mock on a per test basis. This is fine until you get to the tests where we're firing mutations (e.g. updating the product option in the cart), the original mock query ends up getting fired again after the mutation so the test fails the values don't get updated.
> 
> I'm conscious of time on the project so I suggest we live with the flakey tests for the time being and perhaps readdress this when we've got more time to invest into researching / setting this up.

Well done for looking into mate. Shame it isn't quite working but hopefully it's not wasted time and will serve as a platform for when we can come back and look at it further down the line. 

I'm happy to leave it as is for now given, like you said, the time pressures on the project. Might be worth checking @hi-im-si  is okay with this for the moment before signing it off.

stevenbister:
> Well done for looking into mate. Shame it isn't quite working but hopefully it's not wasted time and will serve as a platform for when we can come back and look at it further down the line.

Haha thanks, yeah defo know what doesn't work now at lease 😂 And hopefully the platforms will have matured a bit too and might be some more info out there 🤞

mattwendzina (approved):
LGTM based on chat with @hi-im-si  and coming back to look at this/raise it on some forums if possible.

PR #1166, opened 2023-06-21
Feature/tutor create

### LINKED ISSUE #891: [EP4]: As a tutor I would like to register
### Description

n/a

### LINKED ISSUE #892: [S1.6]: A LS user sends a magic link email and Tutor clicks link in email
### Link to Epic

Epic: https://github.com/Open-Study-College/osc/issues/891

### Description

_No response_

### Linked Tasks

#893 
#894 

_No response_

### LINKED ISSUE #893: Tutor course creation/read/update
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/892

### Description

n/a

### LINKED ISSUE #894: Mark user as IV
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/892

### Description

n/a

hall2521:
- Closes #893 
- Closes #894 
- Closes #892 
- Closes #891 

## 📝 Description

Allows a LS user to:
- create a tutor
- assign a tutor to courses
- mark tutor as an IV
- Select which courses a tutor is an IV for

Allows handling os SSO login/creation in the API

## 🚀 New behavior
New mutation:
- createtutor - create a new tutor
- validateTutor - validates a magic link token and sends a list of courses assigned to a tutor for them to approve
- completetutorcreate - send a list of approved courses and create a password for the tutor
- markUserAsIV - Allows marking a tutor as an IV (happens in create tutor but separate endpoint to assign a tutor as an IV post creation)
- socialLogin - login using the SSO ID - returns an auth token
- socialLoginCreate - create an SSO login

Changes the DB schema to allow for IV users.
- Merge the changes into the schema file 
- Run npx prisma db push


packages/osc-api-auth/src/utils/account.ts
```diff
@@ -328,3 +333,119 @@ export const lmsTokens: LmsTokensFn = async (userId) => {
...
+        magicKey: input.magicKey,
+        email: input.email,
+        password: input.password,
+    };
+
+    const registration = await completeRegistration(complete);
+    const user = await getUserByEmail(input.email);
+    if (user) {
+        for (var i = 0; i < input.courses.length; i++) {
+            if (input.courses[i].accept == false) {
```
Tom-Pearce, line 423: We may want to consider here that "accept" may be a checkbox on the front-end, and so if "unticked" it won't actually be submitted in the form data.

May be worth looping through the list from the database, checking if it is included in the submitted "accepted" list and if not pushing it to an array, and then see if you can do a delete where `userId = X` and `courseId IN array`

packages/osc-api-auth/src/utils/social.ts
```diff
@@ -0,0 +1,43 @@
...
+        },
+    });
+    if (!userSocial) {
+        return new Error('SSO Login failed');
+    }
+    const accessToken = token.access(userSocial.userId);
+    const refreshToken = token.refresh(userSocial.userId);
+
+    return { accessToken, refreshToken };
+};
```
Tom-Pearce, line 43: This doesn't seem right to me.

We should have a list of Socials, so 1 row for each platform. Which should then link to the `UserSocial` table which will contain the `userId` to identify the user in our platform, an `ssoId` to identify the platform used, and an `ssoRef` which should identify the user on the chosen platform as well as some `created` etc fields.

When logging the user in we probably want to receive the ID for the platform (e.g. 1 may be Facebook, 2 Twitter etc), as well as the `ssoRef` for the user which we then match both in `UserSocial` to get the user ID. If we just match based on `ssoRef` theres a chance it could match the wrong user, only likely if 2 platforms both return an ID for the user and it's an auto incrementing value, but rather be safe than sorry.

We probably want to infer the `userId` from the `user` in `AuthContext` here again as well when creating the SSO link

packages/osc-api-auth/src/utils/account.ts
```diff
@@ -328,3 +333,119 @@ export const lmsTokens: LmsTokensFn = async (userId) => {
...
+        },
+    });
+
+    for (var i = 0; i < input.course.length; i++) {
+        await prisma.courseTutor.create({
+            data: {
+                tutorId: user.id,
+                courseId: input.course[i].courseId,
+                createdBy: createdBy,
+                iv: input.course[i].iv,
```
Tom-Pearce, line 392: Want to consider here that `.iv` may not be submitted if it's a checkbox, so want to type it as optional and then put add a null check

```suggestion
                iv: input.course[i].iv ?? false,
```

packages/osc-api-auth/src/utils/account.ts
```diff
@@ -328,3 +333,119 @@ export const lmsTokens: LmsTokensFn = async (userId) => {
...
+    });
+    if (input.IVUser == true) {
+        await markAsIV(user.id, createdBy);
+    }
+
+    //Assign tutor user role
+    const tutorRole = await getRoleByTitle('Tutor');
+    if (!tutorRole) {
+        return new Error('Unable to verify role');
+    }
```
Tom-Pearce, line 377: As a precaution can we move the validation and error responses to the top - again don't want to cause clunky UX by half making a tutor

packages/osc-api-auth/src/utils/account.ts
```diff
@@ -328,3 +333,119 @@ export const lmsTokens: LmsTokensFn = async (userId) => {
...
+    const user = await prisma.user.create({
+        data: {
+            firstName: input.firstName,
+            lastName: input.lastName,
+            email: input.email,
+        },
+    });
+    if (input.IVUser == true) {
+        await markAsIV(user.id, createdBy);
+    }
```
Tom-Pearce, line 371: Just so the logic flows better can be move this to after the point where we assign the tutor role

PR #1204, opened 2023-06-26
feat: add sanityAlgolia route and set up webhook

### LINKED ISSUE #1004: [S78.1]: Indexing Sanity Pages and Blog Posts into Algolia
### Link to Epic

### Description

_No response_

### Linked Tasks

- [x] #1005 
- [ ] 
- [ ]

### LINKED ISSUE #1005: Create Web Hooks
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/1004

### Description

n/a

mattwendzina:
Closes #1004 
Closes #1005 

## 📝 Description

Adds a route to the Remix actions folder that will receive incoming requests from a Sanity webhook which in turn will synchronise data between Sanity and Algolia.

Initially I created two new indexes - `sanity-posts` and `sanity-pages` but this approach only lent itself to multi indexes and pagination on the front end. Because we want to use InfiniteSearch we need to merge the data into one index. Therefore we merge the data into the `shopify_products` index. This is also required because we want to be able to refine on the `post_type`. In the projections (line 109 and 148) I have manually the `post_types` - 'Other' and 'Blog'. The post_type 'Courses' will need to be added to the Shopify Products. (This is how it works in our current live Algolia index). I have manually added this to one shopify product just to help with testing it on the front end.

N/A

## 🚀 New behavior

- New Route in the actions folder
- Sanity/webhook and sanity-algolia npm packages
- New webhook in Sanity - https://www.sanity.io/organizations/ogYSfeeva/project/v6lebos6/api/webhooks/g9eAjeiBIWa0wbtm 

No

## 📝 Additional Information

Firstly you will need to add the following env-var, which is set in the Sanity Webhook.

`SANITY_WEBHOOK_SECRET="w3bh00ks3cret"`

You will also need the following Algolia env_vars if you don't already:
`ALGOLIA_APP_ID`, `ALGOLIA_API_KEY_ADMIN` 

In order to test this you will need to use something like ngrok or LocalTunnel. I used LocalTunnel which was very easy to install and run. Once you have a url up and running you will need to update the URL in the Sanity webhook in order for it to work:

Once this is set it should trigger everytime a change is published in Algolia (it won't run on draft content changes). You should see the changes pull across in Algolia and you can see the log attempts by clicking 'Show attempt log' on the Sanity API page:

Screenshot 2023-06-26 at 16 34 09


packages/osc-ecommerce/app/routes/actions/sanityAlgolia.tsx
```diff
@@ -0,0 +1,198 @@
...
+    const algoliaShopifyProductsIndex = searchClient.initIndex('shopify_products');
+
+    const sanityAlgolia = indexer(
+        {
+            post: {
+                index: algoliaShopifyProductsIndex,
+                projection: `{
+                    _id,
+                    "content": modules[_type match 'module.content'].body,
+                    "description": seo.description,
```
stevenbister, line 88: This is for displaying this excerpt right?
image

I'm not sure if the seo description is the best thing to use here. I know marketing will want to tailor the seo description to be optimised for search engines whereas the excerpt here is more of a preview of the content on the page. I think this might be better as the first bit of content on the page or we add an excerpt field to Sanity for them to use and we can then truncate that on the front end if needs be

stevenbister (commented):
Nice man! 

I've left one comment for you re the description. I'm also wondering if it's worth us using more modules? From my understanding algolia will use the content it takes here to help with searches? We have other content based modules such as accordions, tabs, text grids etc. which marketing could end up using and I'm guessing we'll want that content indexed and searchable?

stevenbister:
Another thought, actually. Do you think it's worth pushing products into Algolia here as well? I know we have Shopify linked up already but we're not managing any of the content in there so I wonder if that could mean we miss out on indexing content for the products 🤔

packages/osc-ecommerce/app/routes/actions/sanityAlgolia.tsx
mattwendzina: Yeah, makes sense. I'll just remove description, and pull in the preview from the content and truncate it 👍

mattwendzina:
I've added in the additional content types, but as discussed, we're leaving products out for now. Because I'm leaving description out, I will update the `Feature/add-new-cards-to-search-route` branch to pull the content from `content` and truncate it.

stevenbister (approved):
LGTM!

PR #1206, opened 2023-06-27
Feature/sen requirements

### LINKED ISSUE #884: Save SEN data
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/880

### Description

Save SEN data

presets of common SEN requiremtns like ADHD, dyslexia, dyspraxia, sight impairment etc
Optional text field for details

SYNC to hubspot

Tom-Pearce:
- Closes #884 

## 📝 Description

Adds Prisma model for SEN data.
Adds Prisma model for `UserAdditional` data - currently includes bio and SEN additional
Adds queries and mutations to manage SEN data for a student/user

### Integrations
Hubspot client is exposed in the Apollo context and all behaviour assumes we are using Hubspot.
There is also an assumption that the _first_ `CrmUserLink` is the one to use.

### Integrations
CRMs are now properly interfaced with a series of required functions. When integrating new CRMs we will have the ability to essentially create an integration library and interface it with `CrmBase` and then simply plug them into the integration root `index` - essentially creating a "plug and play" scenario. 

`CrmUserLink` Prisma model now also has an `active` field allowing us to call the correct integration for the given user and also perform phased rollouts in the event we migrate our CRM away from Hubspot. In the short term we should work on the basis a user will have **only 1** `active` CRM.

By doing all of these we are creating a flexible and robust CRM API that in future should allow us to introduce new and multiple CRM integrations in a simple and concise manner.

We now query and pull relevant details for the active CRM and sync using the specified library. We now have a `crm` object on the `CRMContext` which will contain all our CRM integrations as and when we need to add functionality for other platforms.

### Queries
- `studentSen` pulls a list of student SEN presets (e.g. ADHD, Dyslexia, etc)

### Mutations
- `saveSen` save SEN presets for a student or the logged in user
- `saveSenAdditional` save SEN additional data against `UserAdditional` - free text 512 char limit


prisma/schema.prisma
```diff
@@ -490,3 +492,44 @@ model UserSocial {
   @@index([userId])
   @@index([socialId])
 }
+
+model Sen {
+  id          Int       @id @default(autoincrement())
+  key         String    @unique @db.VarChar(32)
+  title       String    @db.VarChar(64)
+  shorthand   String?   @db.VarChar(32)
```
hall2521, line 500: Short hand here is optional but required in schema file. Is this intentional?

prisma/schema.prisma
```diff
@@ -490,3 +492,44 @@ model UserSocial {
...
+  user    User @relation("userSenUser", fields: [userId], references: [id])
+
+  @@index([userId])
+  @@index([senId])
+  @@map("users.sen")
+}
+
+model UserAdditional {
+  id            Int       @id @default(autoincrement())
+  userId        Int       @unique
```
hall2521, line 526: will this be unique? Is a user only allowed 1 entry into this model?

Tom-Pearce: Yes, it's one record per user for storing additional data like their bio

prisma/schema.prisma
Tom-Pearce: Will fix, think this was just a codeium autocomplete I missed

hall2521 (approved):
LGTM

hi-im-si (changes requested):
As per Tom's continued work, moving this away from 'approved'

packages/osc-api-crm/src/integrations/hubspot/index.ts
```diff
@@ -1,4 +1,163 @@
...
+            lastName: response.properties.lastname,
+            email: response.properties.email,
+            phone: response.properties.phone,
+            mobile: response.properties.mobilephone,
+            addressLine1: response.properties.address,
+            addressLine2: response.properties.address_2,
+            city: response.properties.city,
+            county: response.properties.state,
+            postcode: response.properties.zip,
+            country: response.properties.country,
```
hall2521, line 66: Are all these fields required or will they sometimes be null?

packages/osc-api-crm/src/integrations/hubspot/index.ts
```diff
@@ -1,4 +1,163 @@
...
+        return e.message === 'HTTP request failed'
+            ? new Error(JSON.stringify(e.response))
+            : new Error('Uknown API error: crm.getContact');
+    }
+};
+
+/**
+ * Saves a contact to Hubspot CRM using the provided input data and user ID.
+ *
+ * @param id - The ID of the user whose contact info is being updated.
```
hall2521, line 78: fix this to externalId

packages/osc-api-crm/src/integrations/hubspot/index.ts
```diff
@@ -1,4 +1,163 @@
...
+                firstname: input.firstName,
+                lastname: input.lastName,
+                phone: input.phone,
+                mobilephone: input.mobile,
+                address: input.addressLine1,
+                address_2: input.addressLine2,
+                city: input.city,
+                state: input.county,
+                zip: input.postcode,
+                country: input.country,
```
hall2521, line 95: can a users email not be updated at this stage?

packages/osc-api-crm/src/integrations/hubspot/index.ts
```diff
@@ -1,4 +1,163 @@
...
+            properties: {
+                special_educational_needs: crmIdsArr.join(LIST_SEPERATOR),
+            },
+        });
+
+        return true;
+    } catch (e: any) {
+        return e.message === 'HTTP request failed'
+            ? new Error(JSON.stringify(e.response))
+            : new Error('Uknown API error: crm.syncSen');
```
hall2521, line 145: minor spelling mistake Unknown

prisma/schema.prisma
```diff
@@ -490,3 +495,75 @@ model UserSocial {
...
+
+  @@map("settings.sen")
+}
+
+model UserSen {
+  id        Int      @id @default(autoincrement())
+  userId    Int
+  senId     Int
+  createdAt DateTime @default(now())
+  createdBy Int
```
hall2521, line 518: might be worth linking this to user id as well

hall2521 (changes requested):

packages/osc-api-crm/src/integrations/hubspot/index.ts
Tom-Pearce: They may be empty but the field will always be there from Hubspot, so we're always going to want to return it regardless

packages/osc-api-crm/src/integrations/hubspot/index.ts
Tom-Pearce: Correct, as we probably want to perform some extra auth around it as it will be changing their login details

packages/osc-api-crm/src/integrations/hubspot/index.ts
Tom-Pearce: Good spot, I also typo that word somehow

prisma/schema.prisma
Tom-Pearce: It already is?

hall2521 (approved):
LGTM

hi-im-si (approved):

PR #1207, opened 2023-06-29
Feature/add collections page

### LINKED ISSUE #985: [S74.1]: Build Collections pages
### Link to Epic

Epic: https://github.com/Open-Study-College/osc/issues/984

### Description

_No response_

### Linked Tasks

- [ ] #986 
- [ ] #991 
- [ ] #992 
- [ ]

### LINKED ISSUE #986: Layout above the fold area
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/985

### Description

Hero must use the collection name from Shopify and not be handled by the modules in CMS. Having said that CMS users should be able to change the image, using the settings section in the CMS

### LINKED ISSUE #991: Integrate Algolia
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/985

### Description

n/a

### LINKED ISSUE #992: Collections Card
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/985

### Description

Need updating to pull in the image and description

mattwendzina:
Closes #985 
Closes #986 
Closes #991  
Closes #992  
 
## 📝 Description

Creates the Collections $slug route page. 

Currently the route is just a shell.

## 🚀 New behavior

- Adds a custom Hero that combines data from Sanity and Shopify
- Adds a description property to Collections page in Sanity
- Adds a SelectBar component 
- Updates the Configuration component so we can refine the search results
- Refactors the Search component into its own file so that it is reusable
- Updates the /search route to use the resuable Search component, as well as tweaks for Configuration and Refinement data

No

## 📝 Additional Information

It's possible that the CSS isn't pulling through for the 'Select Your Course' in the SelectBar, sometimes it appears to not be centralised and not in bold. Not sure why this is happening. Also creating an opacity on the Select has a knock on effect of making the dropdown itself use this opacity setting.


packages/osc-ecommerce/app/components/SelectBar/select-bar.scss
```diff
@@ -0,0 +1,49 @@
...
+            font-size: $fs-m;
+
+            @include mq($mq-tab) {
+                justify-content: flex-end;
+            }
+        }
+
+        &__select {
+            display: flex;
+            flex: 2;
```
stevenbister, line 28: These flex values feel a bit magic. And they expand quite far on really large screens (not sure how good that screenshot will be)

image

Is there a reason you went for this over using the grid utility classes to position and size things?

I'd also recommend putting a container class on the div with `c-select-bar`

packages/osc-ecommerce/app/components/SelectBar/select-bar.scss
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        /* stylelint-disable-next-line selector-class-pattern */
+        .c-select__trigger {
+            display: flex;
+            justify-content: center;
+            font-weight: $fw-bold;
+        }
+
+        @include mq($mq-tab) {
```
stevenbister, line 45: Can you move this up to the top above line 11? Just to group it with the `c-select-bar` styles, might get missed down here if we ever add/refactor anything

packages/osc-ecommerce/app/routes/collections/$slug.tsx
```diff
@@ -97,30 +163,98 @@ export const meta: MetaFunction = ({ data, parentsData }) => {
...
+            <Hero data={{ slides: slide, carouselSettings: { carouselName: 'Collections Title' } }}>
+                {isGreaterThanTab ? (
+                    <Breadcrumb
+                        matches={breadcrumbData}
+                        separator={<Icon id="chevron-right" />}
+                        className="o-container"
+                    />
+                ) : null}
+            </Hero>
+            <Row
```
stevenbister, line 228: Rather than adding this `Row` and `Content` could you reuse the `ContentModule` component we're using in the `Modules` loop?
You've missed a couple of props on the `Content` component I think that would potentially get passed from the CMS (buttons, alignment etc.)

packages/osc-ecommerce/app/routes/collections/$slug.tsx
```diff
@@ -97,30 +163,98 @@ export const meta: MetaFunction = ({ data, parentsData }) => {
...
+                configureAttributes={{
+                    facetsRefinements: { collections: [shopifyCollection.handle] },
+                    hitsPerPage: ENV.ALGOLIA_HITS_PER_PAGE,
+                }}
+                env={ENV}
+                refinementData={COLLECTIONS_REFINEMENT_DATA}
+                serverState={serverState}
+            />
             {isPreviewMode ? (
                 <PreviewSuspense fallback={<PageContent {...page} />}>
```
stevenbister, line 253: Might need to update the `{...page}` here as well?

packages/osc-studio/schemas/documents/collection.tsx
```diff
@@ -61,6 +61,13 @@ export default defineType({
             type: 'proxyString',
             options: { field: 'store.slug.current' },
         }),
+        // Description
+        defineField({
+            name: 'description',
+            title: 'Description',
+            type: 'module.content',
```
stevenbister, line 68: I wonder rather than giving them the full contend module we just use `'body'` instead (just the text box). I imagine this part being fairly controlled and not really changing much page to page, would save us having to handle all of the other components that come bundled in 🤔

packages/osc-ui/src/components/Select/Select.tsx
```diff
@@ -151,7 +151,7 @@ export const Select = forwardRef<ElementRef<typeof SelectPrimitive.Trigger>, Pro
                         ref={forwardedRef}
                         aria-label={name}
                     >
-                        <SelectPrimitive.Value placeholder={placeholder} />
+                        <SelectPrimitive.Value>{placeholder}</SelectPrimitive.Value>
```
stevenbister, line 154: Is there a reason we've changed this? The Radix docs look like they suggest the previous way?
image

stevenbister (changes requested):
Looks really good mate! Left a few comments mostly around css. 

Only other thought I had, think it's worth increasing the vertical padding inside the select trigger here:
image
So that you click on most (if not all) of the box, rather than having to select the text specifically?

mattwendzina:
> Looks really good mate! Left a few comments mostly around css.
> 
> Only other thought I had, think it's worth increasing the vertical padding inside the select trigger here: image So that you click on most (if not all) of the box, rather than having to select the text specifically?

Good idea! Done ✅

packages/osc-ecommerce/app/components/Hero/Hero.tsx
```diff
@@ -69,30 +79,40 @@ export const Hero = (props: { data: heroModule }) => {
         return (
             <>
                 {slides?.map((slide) => (
-                    <Slide {...slide} initialSlide={true} key={slide?._key} />
+                    <Slide {...slide} initialSlide={true} key={slide?._key} children={children} />
```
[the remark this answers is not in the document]
mattwendzina: Yup, done 👍

packages/osc-ecommerce/app/components/SelectBar/SelectBar.tsx
```diff
@@ -0,0 +1,36 @@
...
+    const classes = classNames('c-select-bar', className ? className : '');
+
+    return (
+        <div className={classes}>
+            <div className="c-select-bar__title o-container o-container--flush">{title}</div>
+            <div className="c-select-bar__select">
+                <Select
+                    defaultValue="listview"
+                    groupVariants={['tertiary']}
+                    name={'view_select'}
```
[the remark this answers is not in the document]
mattwendzina: LOL. Maybe if you kept a money jar that would help 😂

packages/osc-ecommerce/app/components/SelectBar/SelectBar.tsx
```diff
@@ -0,0 +1,36 @@
...
+                    placeholder={placeholder}
+                    setExternalValue={selectFunction}
+                >
+                    {selectItems.map((item, index) => (
+                        <SelectItem key={index} value={item.handle} to={item.handle}>
```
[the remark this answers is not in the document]
mattwendzina: Yeah good point... So originally I tried wrapping the SelectItem in a <Link/> component and this didn't work. 
I've tried doing this direct on the component itself using an <a/> tag with a conditional but this also doesn't work. 
What about if we still wrapped it in the Link but still used the navigate function? This adds the semantics in at least. Or is this a ridiculous suggestion? 😅 
I'll reach out to the Radix team to see if they have any suggestions.

packages/osc-ecommerce/app/components/SelectBar/select-bar.scss
```diff
@@ -0,0 +1,49 @@
...
+        &__select {
+            display: flex;
+            flex: 2;
+            align-items: center;
+            justify-content: center;
+            width: 100%;
+            margin-right: $space-l;
+            padding: $space-l;
+            background: var(--color-tertiary);
+            opacity: 0.8;
```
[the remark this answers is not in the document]
mattwendzina: Sorted

packages/osc-ecommerce/app/components/SelectBar/select-bar.scss
```diff
@@ -0,0 +1,49 @@
...
+        flex-direction: column;
+        background: var(--color-gradient-primary-270);
+
+        &__title {
+            display: flex;
+            flex: 5;
+            align-items: center;
+            justify-content: center;
+            padding: $space-s 0;
+            padding-right: $space-3xl;
```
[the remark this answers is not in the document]
mattwendzina: Done 👍

packages/osc-ecommerce/app/components/SelectBar/select-bar.scss
mattwendzina: Ahhh, old habits! Always forget when best time to use Grid is... Much easier though lol. Updated. 
Don't see the need for a container on the `c-select-bar` div though?

packages/osc-ecommerce/app/components/SelectBar/select-bar.scss
mattwendzina: It's no longer needed now it's using grid so it's been removed instead = )

packages/osc-studio/schemas/documents/collection.tsx
mattwendzina: Yeah good shout. Updated.

packages/osc-ecommerce/app/routes/collections/$slug.tsx
mattwendzina: I've updated to the `ContentModule` and added the horizontalAlignment and button props in. I've also removed rowSettings as now just using the body (as per below comment).

packages/osc-ui/src/components/Select/Select.tsx
mattwendzina: Yeah, so this one flummoxed me and I faffed about for a while trying to figure out what was going on. Basically the way it was set up (passing in as prop) just stopped working... I found that passing it through to the component was the only way to fix this. See image below:
 
Screenshot 2023-07-04 at 18 43 00

packages/osc-ecommerce/app/components/SelectBar/SelectBar.tsx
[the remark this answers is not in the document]
stevenbister: Cool, I wouldn't spend too long on it. I don't think `select` or `option` takes an `a` tag but I guess this is technically a `button` 🤔 But that might not be valid html either... I was thinking use something else like a popover maybe 🤔 But appreciate that would probably require some effort to match it to the design.
Happy to leave as is.

packages/osc-ecommerce/app/components/SelectBar/select-bar.scss
stevenbister: Noice

packages/osc-ecommerce/app/components/Hero/Hero.tsx
```diff
@@ -61,6 +70,7 @@ export const Hero = (props: { data: heroModule }) => {
                                 : 0 === index
                         }
                         key={slide?._key}
+                        children={children}
```
stevenbister, line 73: Sorry, missed this one last time, you can change this to match the one below?

packages/osc-ecommerce/app/routes/collections/$slug.tsx
```diff
@@ -69,21 +121,41 @@ export const loader = async ({ request, params, context }: LoaderArgs) => {
...
         params: isPreview ? params : null,
+        serverState,
+        ENV: algoliaEnvVars,
     });
 };
 
 // https://github.com/sergiodxa/remix-utils#dynamiclinks
 const dynamicLinks: DynamicLinksFunction = ({ data }) => {
-    return getComponentStyles(data.collection);
+    const stylesheets = getComponentStyles(data.page);
```
stevenbister, line 137: Is there a reason for using the `DynamicLinksFunction` over the regular `LinksFunction`? 
We don't really need to use it on route pages unless you're expecting the styles to be dynamic? But AFAICT we need the styles as a minimum for the `Search` component, so could we use the same `LinksFunction` that we have on the `/search` route?

stevenbister (changes requested):
Looking good man, couple of extra little things I've noticed, and I think a comment from the previous review might have been missed. Nearly there thought!

packages/osc-ecommerce/app/components/SelectBar/SelectBar.tsx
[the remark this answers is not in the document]
mattwendzina: Yeah, so I reached out to Radix and they suggested that we could try the DropDownMenu instead, but weren't 100% sure whether this would support a/link tags.  But yeah either way this would probably take a bit of work to get it to match the design. Could create a ticket to come back and optimise this at a future date?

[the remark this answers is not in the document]
stevenbister: Yeah I think leave it as is for now, don't want to take up too much time here. Yeah I think a ticket or even a discussion actually might be a good idea

packages/osc-ecommerce/app/routes/collections/$slug.tsx
mattwendzina: This is update now.

packages/osc-ecommerce/app/components/Hero/Hero.tsx
mattwendzina: Updated

packages/osc-ecommerce/app/routes/collections/$slug.tsx
mattwendzina: I thought this was basically needed for pages where Marketing can pull in modules - as in it's a way to ensure that styles for the page data gets correctly pulled in?

stevenbister: Yeah, so this already pulls through the styles from the modules in Sanity which is what the `getComponentStyles` is doing, you don't need to change that, from what it was. This is pushed through the `handle` and picked up by the `DynamicLinks` component in the root.

The other stylesheets you've added (e.g. `stylesheets.push({ rel: 'stylesheet', href: selectStyles });`) should ideally sit in the regular `LinksFunction` as they're required by the page and will load sooner.

Admittedly it probably doesn't matter but I think it's clearer to differentiate them 🤔

mattwendzina: Ah! Yeah, that makes total sense = ) Updated 👍

packages/osc-ecommerce/app/components/SelectBar/select-bar.scss
mattwendzina: Ah right. Got it. Added the o-container and o-container--flush, but also needed to add an outer container to house the background colour for when it extends beyond the container itself.

mattwendzina:
As per @hi-im-si suggestion - I've added `select` and `trigger` to the postcss-purgecss safelist.

packages/osc-ecommerce/app/components/SelectBar/SelectBar.tsx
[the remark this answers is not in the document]
mattwendzina: Cool. I've started a discussion here: 
https://github.com/Open-Study-College/osc/discussions/1250 

I've also raised/linked a task which can be found here:
https://github.com/Open-Study-College/osc/issues/1249

PR #1240, opened 2023-07-05
Feature/page theming

### LINKED ISSUE #980: Add theme controls to Sanity
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/979

### Description

Theme controller should be on at least product and collections pages.

This value should allow us to pass a css variable to certain themeable components such as cards/product form

### LINKED ISSUE #981: Pass theme value from Sanity to ecommerce pages
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/979

### Description

n/a

stevenbister:
- Closes #979 
- Closes #980 
- Closes #981 

## 📝 Description

Adds the ability to pass the collection theme settings from Sanity through to the product page.
This works by running a collection query within the Sanity product query, and filtering it by either the matching product tag or the collection slug as a fallback.
By using the product tag `Parent_col:*` we can add some specificity to the collection we want to target. For example this means that any product with `Parent_col:A Levels` will take the theme from the A Levels collection even if it is in both A Levels and Science. e.g

https://github.com/Open-Study-College/osc/assets/23461173/4992dd09-4f0f-42fe-ad25-0ab4c9be92ce

If the product doesn't have a matching `Parent_col` tag then it will fallback to the collection colour e.g.
image

- N/a

## 🚀 New behavior

- Adds themable css variables to the radio and checkboxes
- Adds a utility function to pair gradients and colours
  - This is purely based on my interpretation of the design and which colours are most readable - there is one (nonary I think) which is very low contrast but they are the colours so it will be up to Mkt to choose whether to use this or not
- Adds utility function to get the value of an array based on a regex
- Adds utility function to slugify a string
- Updates product queries
- Adds hook to get the theme from the loader data
- Adds theme colours to product components

## 📝 Additional Information

The `Parent_col` is a tag that is handled within Shopify and will be used to generate collections. Not all of the collections are linked to this tag in the dev store so some products might end up taking the colour of the collection instead. If you updated the condition in the collection in Shopify then this should sync with Sanity and give you the expected outcome.

We will probably need to consider theming for the collections pages, the data is available so should hopefully be fairly simple to implement 🤞 

Something else we'll need to note is that currently product cards will only be styled when they have access to the


packages/osc-ecommerce/app/utils/slugify.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Slugify a string
+ *
+ * @param str string to slugify
+ * @returns slugified string e.g. 'hello-world'
+ * @example slugify('Hello World') // 'hello-world'
+ */
+export const slugify = (str: string) =>
+    str
+        .toLowerCase()
```
mattwendzina, line 10: Is it worth adding .trim() to stop a `-` being added in any case that there is space at the end of the string?

mattwendzina (commented):
Looks good mate 👍  Just a couple of small comments.

packages/osc-ecommerce/app/routes/courses/$slug.tsx
```diff
@@ -151,6 +165,8 @@ export default function Index() {
     const intersectionRef = useRef<HTMLDivElement>(null);
     const isPreviewMode = isPreview && query && params;
     const pageThemeColor = page?.theme ? page.theme.color : 'gradient-primary';
+    // console.log('page', page);
+    // console.log('pageThemeColor', pageThemeColor);
```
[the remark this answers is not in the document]
stevenbister: Ooo nice catch, yep!

packages/osc-ecommerce/app/utils/slugify.ts
stevenbister: Ahh yeah good thinking! I've updated that and updated the test to check for it

mattwendzina (approved):
Nice mate! LGTM 🚀

PR #1241, opened 2023-07-05
Feature/configure document types

### LINKED ISSUE #1226: Configure Document Types
### Link task to story
Story: 

### Description
Configure Document Types:

- Home Page.
- Pages.
- Jobs.
- Job Categories.
- Testimonials.
- Navigation.
- Global Settings

jkellerman:
Closes #1226 

## 📝 Description

Configure document types on Sanity for OSC Recruitment

N/A

## 🚀 New behavior

N/A

No


packages/osc-studio/schemas/singletons/settingsRecruitment.ts
```diff
@@ -0,0 +1,218 @@
...
+            type: 'string',
+            group: 'contact',
+            validation: (Rule) =>
+                // regex to match email address: https://ihateregex.io/expr/email/
+                Rule.regex(/[^@ \t\r\n]+@[^@ \t\r\n]+\.[^@ \t\r\n]+/, {
+                    name: 'email',
+                }),
+        }),
+        // Cart
+        defineField({
```
stevenbister, line 183: Might be worth taking this out, no need to have the cart on here

stevenbister (changes requested):
Hey man looks really good, left a few comments here and there, nothing major :)

I was also thinking, rather than having a bunch of files that are `filenameRecruitment` it might be a bit tidier if we move those ones into their own recruitment folder. So the directory structure would be something like

`documents/recruitment/filename.ts`
`objects/recruitment/filename.ts` etc.

What do you think?

stevenbister (changes requested):
Looking really good, just a couple of files missed when moving into the recruitment directories

- `annotations/linkInteranRecruitment.tsx`
- `objects/module/buttonRecruitment.tsx`
- `objects/module/cardsRecruitment.tsx`
- `objects/module/tabsRecruitment.tsx`

If we can move those into `recruitment` folders as well this will be good to go!

stevenbister (approved):
LGTM!

PR #1243, opened 2023-07-05
Feature/configure GitHub actions

### LINKED ISSUE #1227: Configure GitHub Actions
### Link task to story
Story: #1215

### Description
Configure GitHub Actions by duplicating the existing configuration to work with this package

jkellerman:
Closes #1227 

## 📝 Description

Configure gitHub actions for OSC Recruitment

N/A

## 🚀 New behavior

N/A

No


stevenbister (changes requested):
I think this might be a little out of date so not looked over the sanity stuff as fairly confident they've been sorted in another branch? Might be worth checking these over? (if easier start a new branch and just move the GH actions stuff across?)

With regards to the GH actions, the workflow file looks good! You'll need to add a `Dockerfile` and `Dockerfile.pr` to the `osc-recruitment` package similar to `osc-ecommerce`, you'll just need to adjust the `ARG`/`ENV` variables to remove things like the Shopify variables and include the correct Sanity ones. Does that make sense? Give me a shout if not :)

packages/osc-recruitment/Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+ENV ALGOLIA_PRIMARY_INDEX_GROUPED=${ALGOLIA_PRIMARY_INDEX_GROUPED}
+ENV ALGOLIA_PRIMARY_INDEX_QUERY_SUGGESTIONS=${ALGOLIA_PRIMARY_INDEX_QUERY_SUGGESTIONS}
+ARG PUBLIC_STOREFRONT_API_TOKEN
+ARG PRIVATE_STOREFRONT_API_TOKEN
+ARG PUBLIC_STOREFRONT_API_VERSION
+ARG PUBLIC_STORE_DOMAIN
+ENV PUBLIC_STOREFRONT_API_TOKEN=${PUBLIC_STOREFRONT_API_TOKEN}
+ENV PRIVATE_STOREFRONT_API_TOKEN=${PRIVATE_STOREFRONT_API_TOKEN}
+ENV PUBLIC_STOREFRONT_API_VERSION=${PUBLIC_STOREFRONT_API_VERSION}
+ENV PUBLIC_STORE_DOMAIN=${PUBLIC_STORE_DOMAIN}
```
stevenbister, line 61: There's a lot of env vars in these dockerfiles that we probably don't need in osc-recruitment (algolia, shopify, planetscale etc.) I would have a look at what envs you're using. Probably only `SANITY_STUDIO_RECRUITMENT_PROJECT_ID`? Remove the ones you're not using and add these in.
You'll probably also need to add these into the `deploy-preview-app` and `deploy-production-app`.

The best thing to do would be to create the label in GitHub add it to this pr and see how your app builds.

stevenbister (changes requested):
Hiya mate, left a comment on the dockerfile. That'll need applying to both dockerfiles.

Could you also rebase this PR, there's a lot of outdated stuff in here so probably won't be a good representation when you test deployments.
Shout me if you need a hand!

stevenbister (changes requested):
Okay think I've worked out why the styles aren't getting generated on build.
It looks like purge css isn't picking up the classes on the component as they're being imported from the osc-ui library, rather than being part of osc-recruitment.

Have a look at https://purgecss.com/safelisting.html I think you'll need to update your `postcss.config` to allow the override classes. e.g
```js
safelist: {
    standard: [/select$/, /trigger$/],
}
```
is currently allowing the select and trigger styles to get built. So maybe just a case of adding the other classes to that list? 🤔

PR #1244, opened 2023-07-06
Feature/store and read notifications

### LINKED ISSUE #1124: [EP22] Notification system

### LINKED ISSUE #1125: [S27.1] Create notification library
- Epic #1124

### LINKED ISSUE #1126: Store a notification
- Story #1125 

Creation of DB tables:
- notifications
- notifications types

Fields
- User id
- message
- url to another page
- type
- date
- read
- read date

Function to store a notification

### LINKED ISSUE #1127: Read notifications
- Story #1125 

Queries for listing/counting notifications

Filters:
- Read/Unread/All
- Type

hall2521:
- Closes #1126 
- Closes #1127 
- Closes #1125 
- Closes #1124  

## 📝 Description

Adds a notification system to the backend with the ability to store notifications as well as access notifications via status + type
## 🚀 New behavior
2 new queries and 1 new mutation:
- getAllNotifications - allows you to get all your own users notifications. can filter on this list by status (Unread, Read, All)
- getAllNotificationsByType - allows you to get all your own users notifications, can filter on this list by type and status

- storeNotifications - create a new notification

Makes an update to the prisma schema so changes will need to implemented via npx prisma db push


prisma/schema.prisma
```diff
@@ -490,3 +491,27 @@ model UserSocial {
   @@index([userId])
   @@index([socialId])
 }
+
+model Notifications {
+  id        Int       @id @default(autoincrement())
+  userId    Int
+  message   String
+  url       String
```
Tom-Pearce, line 499: Thinking about the URL generation. If we hardcode it in the way we have it, then if we were to change the routing on the front end the URLs would then be wrong.

What do you think about storing the `url` as a template (e.g. `ah.osc.com/assignments/submissions/{subId} against the `notificationType` and then just storing the relevant ID here as `targetId` or something. Then on the front end they just find/replace the template to generate the final URL. This way the URL can be changed inline with any front end routing changes.

We may also need to consider the need for multiple `targetId` though. Let me know your thoughts on this?

prisma/schema.prisma
```diff
@@ -490,3 +491,27 @@ model UserSocial {
...
+  Type NotificationType @relation(fields: [typeId], references: [id])
+
+  @@index([userId])
+  @@index([typeId])
+  @@map("settings.notifications")
+}
+
+model NotificationType {
+  id            Int             @id @default(autoincrement())
+  type          String
```
Tom-Pearce, line 515: Can we rename this to `title`?

packages/osc-api-settings/src/resolvers/resolver.ts
```diff
@@ -64,5 +82,8 @@ export const resolvers = {
         ) => {
             return await preferences.saveByKey(user.id, input.key, input.value);
         },
+        storeNotification: async (_: undefined, { input }: saveNotificationArgs) => {
+            return await notifications.storeNotification(input);
```
Tom-Pearce, line 86: I think somewhere we may want to perform a check against the `userId` and their `systemPreferences` for the notification type - Or maybe more appropriately we should add a ToDo and a new issue for adding support for `notificationTypeSubscriptions` which we can then check against before creating the notification?

Tom-Pearce (changes requested):

packages/osc-api-settings/src/schemas/schema.ts
```diff
@@ -25,6 +25,13 @@ export const typeDefs = gql`
     type Notifications {
         id: Int!
         message: String!
+        targetId: Int!
+        Type: NotificationType
```
Tom-Pearce, line 29: Being pedantic but can we make the `T` lowercase in `Type` Guessing it's coming through from the prisma schema?

Tom-Pearce (changes requested):
Want to keep the properties as camelCase and keep PascalCase for the model/type name

packages/osc-api-settings/src/schemas/schema.ts
hall2521: yup this comes from the prisma schema. it happens in a number of relations so may be worth tagging this as a separate issue?

Tom-Pearce: Sounds good to me we'll do that instead

Tom-Pearce (approved):
LGTM 🚀

PR #1259, opened 2023-07-12
Bugfix/product page errors

stevenbister:
## 📝 Description

Fixes a couple of issues with errors occurring on the product pages.

- If a product isn't in a collection then the product page throws e.g.
image
- If the page doesn't exist, rather than 404-ing then it will throw e.g
image

## 🚀 New behavior

- Adds optional chaining to prevent error being thrown and fallback into the ternary instead
image
- Adds an extra redirect/404 step just after the product query to make sure 404 is fired if the product doesn't exist in Shopify
image


mattwendzina (approved):
LGTM

PR #1263, opened 2023-07-12
Feature/setup design tokens

### LINKED ISSUE #1229: Set up design tokens
### Link task to story
Story: 

### Description
Set up design tokens
- Colours

jkellerman:
Closes #1229 

## 📝 Description

Adds fonts, colours and logo for recruitment website

N/A

N/A


packages/osc-design-tokens/tokens/colors.js
```diff
@@ -109,6 +109,10 @@ for (const direction of directions) {
     gradients[`gradient-nonary${dirLabel}`] = `linear-gradient(${direction}, ${gradientLemonCyan})`;
 }
 
+// Bloom colours
+const bloomPrimary = '#ffed00';
+const bloomSecondary = '#1c4a50';
```
stevenbister, line 114: Do we need to include the other secondary colours?
image

packages/osc-ui/src/components/Logo/Logo.tsx
```diff
@@ -25,16 +25,22 @@ export interface LogoProps extends HTMLAttributes<HTMLElement> {
      * Custom class
      */
     className?: string;
+    /**
+     * Imports logo as prop
+     * @default logo
+     */
+    logoId?: 'logo' | 'bloom-logo';
```
stevenbister, line 32: I think `id` rather than `logoId` would work better here 🤔 You're already on the `<Logo>` component so feels unnecessary/repetative to include the word "logo" in the prop.

What do you think?

packages/osc-ui/src/components/Logo/Logo.stories.tsx
```diff
@@ -19,3 +19,5 @@ const Template: Story<LogoProps> = (args) => <Logo {...args} />;
 
 export const Primary = Template.bind({});
 Primary.args = {};
+export const Bloom = Template.bind({});
+Bloom.args = {};
```
stevenbister, line 23: I would add the `logoId` as a default prop here otherwise this story is the same as the primary one

packages/osc-ui/src/styles/settings/_tokens.scss
```diff
@@ -544,6 +544,24 @@ $colors: (
         neutral-500: hsl(0deg 0% 96%),
         neutral-600: hsl(0deg 0% 98%),
         neutral-700: hsl(0deg 0% 100%)
+    ),
+    bloom: (
+        primary: hsl(56, 100%, 50%),
+        secondary: hsl(187, 48%, 21%),
+        tertiary: hsl(42, 72%, 81%),
```
stevenbister, line 551: Did you manually add these? Not sure how these values would have been generated?

This file is generated by the `tokens:generate` command so will be replaced everytime it runs. If you want to include these colours here you'll need to add them to the `colors.js` file in `osc-design-tokens`

stevenbister (changes requested):
Good work! Couple of comments but otherwise looks pretty solid

packages/osc-design-tokens/tokens/colors.js
jkellerman: Secondary colours have been added in commit [c208707](https://github.com/Open-Study-College/osc/pull/1263/commits/c208707fb6d638e10b18241b5f587cbf6d5f23c3)

packages/osc-ui/src/components/Logo/Logo.tsx
jkellerman: Yup, agreed

packages/osc-design-tokens/tokens/colors.js
```diff
@@ -109,6 +109,24 @@ for (const direction of directions) {
...
 
+// Bloom colours
+const bloomPrimary = 'hsl(56, 100%, 50%)';
+const bloomSecondary = 'hsl(187, 48%, 21%)';
+const bloomTertiary = 'hsl(42, 72%, 81%)';
+const bloomQuaternary = 'hsl(40, 56%, 95%)';
+const bloomQuinary = 'hsl(3, 71%, 57%)';
+const bloomSenary = 'hsl(33, 100%, 56%)';
+const bloomSeptanary = 'hsl(340, 77%, 90%)';
+const bloomOctonary = 'hsl(166, 39%, 65%)';
```
stevenbister, line 120: Can you update these values to include the deg format, just so they're consistent with the other values
e.g
```suggestion
const bloomOctonary = 'hsl(166deg 39% 65%)';
```

stevenbister (changes requested):
Nice man! One little minor comment but otherwise we're there :)

packages/osc-design-tokens/tokens/colors.js
```diff
@@ -109,6 +109,24 @@ for (const direction of directions) {
     gradients[`gradient-nonary${dirLabel}`] = `linear-gradient(${direction}, ${gradientLemonCyan})`;
 }
 
+// Bloom colours
+const bloomPrimary = 'hsl(56, 100%, 50%)';
+const bloomSecondary = 'hsl(187, 48%, 21%)';
+const bloomTertiary = 'hsl(42, 72%, 81%)';
```
stevenbister, line 115: Not sure how I feel about how some of the colours are mapping across, for example in the testimonials PR I don't think that browney/gold colour looks that nice when used in the cards 🤔

jkellerman: The browney/gold colour and the red that is currently the background of one of the content blocks are not intentional. For example, the browney/gold colour is supposed to be white but for some reason sanity is mapping the colours incorrectly.

Also, we discussed in the brief with marketing that each section will probably need a different background colour instead of the whole page being yellow, meaning we need the theme colours for Bloom in sanity to select different colour backgrounds for each content block. I believe this is the best way to go about this because these content blocks are imported as modules (hope that makes sense, please correct me if I am wrong).

I'm planning on changing the colours of the site so it should look a lot different to what it does now, I was also going to do this in a separate PR, so maybe we should accept these changes and then change the colours in different PR, what do you think?

packages/osc-design-tokens/tokens/colors.js
```diff
@@ -109,6 +109,24 @@ for (const direction of directions) {
...
 
+// Bloom colours
+const bloomPrimary = 'hsl(56deg, 100%, 50%)';
+const bloomSecondary = 'hsl(187deg, 48%, 21%)';
+const bloomTertiary = 'hsl(42deg, 72%, 81%)';
+const bloomQuaternary = 'hsl(40deg, 56%, 95%)';
+const bloomQuinary = 'hsl(3deg, 71%, 57%)';
+const bloomSenary = 'hsl(33deg, 100%, 56%)';
+const bloomSeptanary = 'hsl(340deg, 77%, 90%)';
+const bloomOctonary = 'hsl(166deg, 39%, 65%)';
```
stevenbister, line 120: Ah sorry one more! When using the "deg" syntax you don't need the commas. I _think_ they won't work correctly if you use commas as well as the "deg". You okay to remove them, and also re-run the `tokens:generate` command to test and update new scss file?

stevenbister (changes requested):

packages/osc-design-tokens/tokens/colors.js
stevenbister: Yeah that's all good! Let's tackle that once the colours in Sanity are sorted!

stevenbister (approved):

PR #1268, opened 2023-07-13
Refactor/osc UI

### LINKED ISSUE #775: Improve osc-ui setup to reduce bundle size and make it available for treeshaking
### Description

Improve the osc-ui setup by either expanding on what we have (Vite & Rollup) or removing Rollup and leaning into Vite fully. Either way we will need to change the setup to allow for better code splitting and bundling.

At the moment osc-ui is bundling _everything_ into one file so Remix isn't able to treeshake it leaving us with a huge bundle

https://pr-773-osc-ecommerce.fly.dev/build/_shared/chunk-2CUZYLHL.js

https://googlechrome.github.io/lighthouse/treemap/?gzip=1#H4sIAAAAAAAAE62V32/TMBDH/5UqTyA1SZ04v/bG2omCOiFaOsYQQm7sJO4c27Kdsmrq/47bVWik5mFlUh6SO9/3c3c5248ea5R38ehVlCO2VMy78BpjpL4IQ6n8LIt9oUuflKJtiSpJULFtgMkm9IYe6jA1eh+sS0Wl8Y0ipEXSx8igvZliq+byDT1DDSPWuzh4B1+evIPJkxeTpygquF2z1AQPKqEGR40BktIu0qVQVoJ3jB0/JlRLhrbXAu+lKbcxLTJ0Qw6SBlF2yNZs5d7fy4jbKOv+/uhx1JIXtEERLTpru9yavQDIswzshi+TWXWU4bBFnFZEGz+HACQAjoO1PgXAJCqGXsc725ajaXQej3CjtkHJqH3xp9O7cby8WTiZIEniHhPk0EH9RXFNTGBUp42kTJjAssOVEEYbhWS4SUIjg+OqP+agpdzNLaIi6oFhAvOhVza2AkX48z/mUv6Hatxv4V71WT1vOm5nQxL89jQ8iXc/zmv4T90gRXBYNh2/96Px8u7bbDpz5pjDKMrTXpIphCBNXoN9e/35+iodJ+4Jy5Mc9NBRkWXpa5Bnk0+z6e3sxknOUgD7c5aCMytWdgj8MZzPsw/x3D0IIDrhRXl05vb9u86Pi6v55Hbu3sPRKE2zHjg7G6xEZxVCyjF58C/ffX0PZ5dTJ7cYJf32wuwZtGyUaIlPHgzh2p6+ll9LKXjdNut7WfFVg2rZrqt7xHm1Ygy1dbjW9nHPETgpEhQj+P88LFonMMmLvAeMRzGw23W3s0eG4BWtF8QYyuvDTcBEiQ7XEOH+cuHZVb8BGqpSURAHAAA=
image

This is contributing to a long blocking time and likely contributing to some flickering as components are loaded.

### Proposed Solution or API

Not sure on a full implementation suggestion yet but somethings to bear in mind is that by including things like react in the dependencies array is causing them to be bundled as part of the lib.

From a bit of reading around it would be better to include those as peer deps like the following ui libraries:

- https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/accordion/package.json
- https://github.com/radix-ui/primitives/blob/main/packages/react/accordion/package.json

### Alternatives

_No response_

### Additional Information

_No response_

stevenbister:
- Closes #775 

## 📝 Description

Updates the vite.config in osc-ui to enable treeshaking, reduce bundle size and improve css importing in Remix apps.
I've removed the rollup config and moved all build/config into the vite.config instead. This has allowed us to lean into Vite's built in treeshaking and library capabilities, as well as the built in css processing, meaning we can remove the transpilecss stuff that was generating the long filenames.

I have a few outstanding tasks that I want to tackle as part of this ticket, but felt like this was a good point to get this PR'd.
- Sort out the issue with duplicates of the css variables being generated per component css file and add the postcss stuff back in as a config file (#1270)
- Fix the ts errors and enable typescript in vite config, I can set this up to fail if typescript fails on build which should help us catch these ts errors that have crept in
- Move dependencies such as remix/Link into peer dependencies or find another way to remove them from the bundle

- When I started this work the bundle size of osc ui on a production built local site was 1.2mb: https://github.com/Open-Study-College/osc/issues/775#issuecomment-1613333285

- Import css files using the file-path generated name e.g. 
```tsx
import button from 'osc-ui/dist/src-components-Button-button.css'
```

- Import the spritesheet by directly accessing the dist folder in osc-ui
```tsx
import spritesheet from 'osc-ui/dist/spritesheet.svg';
```

## 🚀 New behavior

- In the same environment the bundle size is now 673kb: https://github.com/Open-Study-College/osc/issues/775#issuecomment-1634131970
This is still a little on the larger side but I'm unsure how to reduce this unless perhaps we look into using react lazy or something like [Loadable components](https://loadable-components.com/docs/getting-started/) however this might mean we need to adjust out components to use default exports as react lazy currently doesn't support named exports, which felt out of scope for this bit of work right now.

- Import css files using the original name of the file
```tsx
import css from 'osc-ui/my-css.css'
```

- Import the spritesheet directly from osc-ui
```tsx
import spritesheet from 'osc-ui/spritesheet.svg';
```

I've also updated the `nx.json` config to make sure that osc-ui is built before any dev commands are run.

Component prop interfaces **must** be unique (e.g. `ButtonProps`). I've changed the way we export components out of osc-ui in the `src/index.ts` file so now the interfaces get exported too. If multiple components have an interface with the name of `Props` then the build will fail.

When importing css files now you can import them by doing
```tsx
import css from 'osc-ui/my-css.css'
```
rather than the previous long css path. Similarly the spritesheet can now be accessed from
```tsx
import spritesheet from 'osc-ui/spritesheet.svg'
```

## 📝 Additional Information

There are a lot of file in this PR but a majority of them are updates to css paths and changes to prop names so that we don't get conflicts when importing components into the entry `index.ts` file, so it's actually not as bad as it seems ;)

Once you've pulled this down I would recommend deleting your `dist/` folder in `osc-ui` before running any commands, just so you have a clean slate.

To test make sure to run `npm run setup` and then you can run `npm run dev:osc-ecommerce`


mattwendzina (commented):
Nice work mate, a lot of updates you've had to do here 😅 

Seems to be working well for me. Is it worth updating the AH file paths too? This was causing a failure to do a full build.

stevenbister:
> Seems to be working well for me. Is it worth updating the AH file paths too? This was causing a failure to do a full build.

Yeah good shout! Think I was being lazy 😂 

I also got that error you mentioned where the paths were failing when I started dev. I noticed that trying to build osc-ui, then also run dev mode on it in parallel with the remix app commands would cause it to build (because of NX recognising it being a dependency of the remix app), then (because of how vite works) remove the files from the dist dir and re-build them in parallel with remix spinning up, meaning at times the files wouldn't exist when remix started building :|

To get around this I've updated the dev commands to only run dev on whichever app you've targeted. So now NX will build osc-ui then run dev on remix. If we need/want to edit osc-ui at the same time we'll need to spin up a separate terminal to run dev on osc-ui.

mattwendzina:
As discussed, getting issues with running `osc-ecommerce` and `osc-ui` independently in parallel. When making changes to a component osc-ui rebuilds and triggers ecommerce however it seems to be working intermittently and sometimes `e-commerce` throws:
```
💿 Rebuilding...
✘ [ERROR] The build was canceled
```
And the changes don't pull through.

As suggested, I've tried adding `emptyOutDir: false,` to the vite config in `osc-ui` but this hasn't fixed it.

mattwendzina (approved):
LGTM! 🎉 Although it does look like you now have some conflicts that need resolving.

Issue #1203, opened 2023-06-26
[S96.1]: [As a developer, I want to improve website performance so that users will have a great experience]


hi-im-si:
### Link to Epic

Epic: 

### Description

Improve the osc-ui setup by decreasing bundle sizes for a better performing application

### Linked Tasks

- [ ] #775

PR #1270, opened 2023-07-13
Refactor/osc UI css

stevenbister:
## 📝 Description

Related to #1203 this pr moves the `:root` declarations with the css variables out of the settings and into the elements group. This will prevent us from loading in the css variables multiple times (once in each component stylesheet) causing unnecessary bloat. While still giving us access to the properties.

The space, font size and shadow css variables live within the related settings partial stylesheet.

## 🚀 New behavior

The space, font size and shadow css variables now live within the `_root.scss` partial which is imported before all of the other elements.


hi-im-si (approved):
Working well 🚀

PR #1271, opened 2023-07-14
Fix/ecommerce header nav

jkellerman:
Closes #

## 📝 Description

Fixed header for ecommerce site due to creation of new HeaderInner sub component

Screenshot 2023-07-14 at 11 35 22

## 🚀 New behavior

Screenshot 2023-07-14 at 11 33 40

No


packages/osc-ui/src/components/Header/Header.tsx
```diff
@@ -84,7 +84,7 @@ export const HeaderInner = (props: HeaderInnerProps) => {
     const classes = classNames('o-container c-header__inner', className);
 
     return (
-        <div className="u-bg-color-primary">
```
stevenbister, line 87: I see you've passed it in as a class on the parent `<Header>`. Is there a reason for moving this out of here and putting it there?

stevenbister (commented):
Looks good! I do have a question which I've left against the file

stevenbister:
Oh also don't forget to include the `build:osc-ecommerce` label here as you'll want this branch to build a preview link and deploy when we merge

packages/osc-ui/src/components/Header/Header.tsx
jkellerman: Yes, firstly, the ecommerce website uses tertiary for the background colour, recruitment uses primary, so the class needs to go on the component for flexibility. Also, the class needs to be on the Header component to prevent only the inner header having the styles (see below). 
Screenshot 2023-07-17 at 17 26 25

stevenbister: Okay cool yeah that makes sense. In that case can we remove this `div`? Keeps the markup tidy

stevenbister (changes requested):

stevenbister (approved):
LGTM!

PR #1277, opened 2023-07-17
Feature/build autocomplete as sanity component

### LINKED ISSUE #1159: Autocomplete - Build, style and add to Sanity modules
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/1158

### Description

Build Autocomplete component into sanity as a module to be added on any page.

- potentially adding a colour variant on the input to ensure that the inputs work on different background colours

### LINKED ISSUE #1160: Make dynamic with sanity
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/1158

### Description

n/a

mattwendzina:
Closes #1159 
Closes #1160

## 📝 Description

Enables CMS users to add a search module in Sanity and adds it to osc-ecommerce

No way for CMS to add a Search Module

## 🚀 New behavior

- Search Module in Sanity with a customisable title, placeholder, background and input/text colours
- Enables the Search Module to be added to pages
- Tidies up the /search and collections /$slug routes and adds checks for envVars
- Creates a helpers file to keep better track of Algolia envVars 
- Updates names of Algolia envVars in Autocomplete component
- Adds onStateChange to Search component to enable state to be updated manually

No

## 📝 Additional Information

There was an issue with transitioning from a page that uses the Autocomplete to the search page that meant the URL got updated twice. E.g. If you type 'Biology' into an Autocomplete instance and hit enter it would navigate to '/search' route initially with the correct query param - ie - /search/?query='biology' but then do a second update to '/search'. 

I managed, by luck, to discover that is we use `onStateChange` in the InstantSearch instance then this enables us to manually update the state. In doing so it appears that it stops Algolia from doing this second state change where it updates the query to `/search`. However, I found that this still didn't quite fix everything... The results were correct but the searchbox itself was empty. This is because for some reason the `query` doesn't get set in the `useSearchBox` hook in the `SearchBox` component. I have found a way around this by adding the `useHits` hook into this component. This hook has the correct query when the page loads and so I've added a `useEffect` in which will only run on initial load and look to see if there is a query in the `useHits` hook - if there is it sets it in the SearchBox using the refine function.


packages/osc-ecommerce/app/utils/algolia.helpers.ts
stevenbister, line 1: Do you think these should sit in our `constants.ts` file or maybe the `types/algolia.ts`? These aren't really utility functions so feels like they shouldn't sit in here? What do you think?

packages/osc-studio/schemas/objects/module/search.ts
```diff
@@ -0,0 +1,61 @@
...
+            title: 'Row',
+        },
+        {
+            name: 'content',
+            title: 'Content',
+        },
+    ],
+    fields: [
+        defineField({
+            name: 'title',
```
stevenbister, line 22: I feel like this should either be a text input or named something other than title? Using the body type means all kinds of content can be added here so title feels a bit misleading when reading the code in the components in ecommerce.

packages/osc-ui/src/components/Autocomplete/Autocomplete.tsx
stevenbister, line 1: The button on line 287 doesn't do anything, should it point to the /search page?

packages/osc-ecommerce/app/routes/collections/$slug.tsx
```diff
@@ -115,13 +115,15 @@ export const loader = async ({ request, params, context }: LoaderArgs) => {
     );
 
     const hubspotFormData = await getHubspotForms(page);
+    const algoliaSearchModule = page.modules.some((module) => module._type === 'module.search');
```
stevenbister, line 118: Getting an error on the collections pages, I _think_ it's related to this line

image

```
TypeError: Cannot read properties of null (reading 'some')
    at loader5 (/Users/stevebister/SITES/osc/packages/osc-ecommerce/build/index.js:9632:88)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.callRouteLoaderRR (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/server-runtime/dist/data.js:52:16)
    at async callLoaderOrAction (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/router/router.ts:3626:16)
    at async Promise.all (index 1)
    at async loadRouteData (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/router/router.ts:2954:19)
    at async queryImpl (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/router/router.ts:2735:20)
    at async Object.query (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/router/router.ts:2611:18)
    at async handleDocumentRequestRR (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/server-runtime/dist/server.js:163:15)
    at async requestHandler (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/server-runtime/dist/server.js:61:18)
```

stevenbister (changes requested):
Looks good! Got a few comments/suggestions also getting some interesting behaviour with the url. I've added a recording for you below:

https://github.com/Open-Study-College/osc/assets/23461173/36d9ff60-fa06-49e6-b962-dfe74a265ff3

Navigating back changes the search params, which changes the number of results and then when you go all the way back to the home page it keeps you on the homepage but changes the url to /search/

mattwendzina:
> Looks good! Got a few comments/suggestions also getting some interesting behaviour with the url. I've added a recording for you below:
> 
>  Screen.Recording.2023-07-19.at.09.28.55.mov 
> Navigating back changes the search params, which changes the number of results and then when you go all the way back to the home page it keeps you on the homepage but changes the url to /search/

Now fixed by moving away from programmatic navigation to using anchor tags and the native form submission. The url does still change - pretty sure this is the url syncing from Algolia, but it doesn't add a step in the history so pressing back still takes you back to the previous page, so don't think this should be an issue..!

packages/osc-studio/schemas/objects/module/search.ts
```diff
@@ -0,0 +1,61 @@
...
+            group: 'content',
+        }),
+        defineField({
+            name: 'searchText',
+            title: 'Search text',
+            type: 'string',
+            description: 'Placeholder text for the search box',
+        }),
+        defineField({
+            name: 'inputColor',
```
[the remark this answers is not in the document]
mattwendzina: Ah yeah good point. I've added them to the "Content" group. Not sure what you mean by - "should maybe be the default?"

packages/osc-studio/schemas/objects/module/search.ts
```diff
@@ -0,0 +1,61 @@
...
+            title: 'Text and Input colour',
+            type: 'string',
+            components: {
+                input: ColorPickerRestricted,
```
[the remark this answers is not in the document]
mattwendzina: I thought it didn't make sense to give them so many colour choices for the text and textInput colour. I think ideally it would just be solid input colours but didn't want to faff about with creating another ColorPicker file so thought that at least restricted reduced the options. That said, having the option to have different greys is pretty important so think I'll just change it. I guess if they want a wild gradient for the text and input then they will at least have the option 😂

packages/osc-ecommerce/app/components/InstantSearch/Components/SearchBox.tsx
```diff
@@ -1,19 +1,31 @@
+import { useEffect } from 'react';
```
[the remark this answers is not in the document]
mattwendzina: Sorted

packages/osc-studio/schemas/objects/module/search.ts
mattwendzina: Yeah good point. Well ideally we need them to be able to adjust the colour and size so thought it was easiest just to use Content as it has the colour picker. Is there a type that allows just text and those kinds of adjustments?

packages/osc-ui/src/components/Autocomplete/Autocomplete.tsx
mattwendzina: Updated 👍

packages/osc-ecommerce/app/routes/collections/$slug.tsx
mattwendzina: Ah good spot. I've just added optional chaining to all of these cases.

packages/osc-studio/schemas/objects/module/search.ts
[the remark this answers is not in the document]
stevenbister: If you look at any of the other modules (e.g. tabs.ts) there is a `groups` array and one of the items has a `default: true` property on it
image

That makes Sanity open that group up rather than the `all fields` tab when an editor is editing the module

packages/osc-ecommerce/app/utils/algolia.helpers.ts
stevenbister: I don't seem to be getting an issue putting them into `constants`, and I can't see why that would be the case? What's the error you're getting? 

Tbf types probably isn't the best pace for them, but I do think `constants` is. In my mind that `utils` folder should only really be for holding functions, rather than data

packages/osc-studio/schemas/objects/module/search.ts
stevenbister: Yeah the type is fine, I just think the name "title" isn't the clearest when reading the code as it can be more than just the title ya know?

packages/osc-studio/schemas/objects/module/search.ts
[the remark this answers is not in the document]
mattwendzina: Ah, nice. Done 👌

packages/osc-ecommerce/app/utils/algolia.helpers.ts
mattwendzina: Now sorted this. As per discussion, I used the `typeof document` in that file. They have been set to null as we pull them through the loader but it's not happy if you don't deal with them on the client. 
I've also managed to move the searchClient outside of the components by setting up a getClient and config file to and pulling in the app id and api key through the root, similar to what you've set up with Sanity.

packages/osc-studio/schemas/objects/module/search.ts
mattwendzina: Updated to 'Description' 👍

mattwendzina:
I've temporarily removed the Algolia Insights Plugin as it was causing and issue and isn't actually being used atm anyway. I've added a ticket to ensure this gets implemented when I do the insights work:
https://github.com/Open-Study-College/osc/issues/1336

packages/osc-ecommerce/app/components/Autocomplete/AutocompleteInlineSearch.tsx
```diff
@@ -0,0 +1,51 @@
...
+        >
+            <div
+                className="o-container o-container--4xs"
+                style={{
+                    ['--input-color' as string]: `var(--color-${module.inputColor})`,
+                    ['--autocomplete-input-font-size' as string]: `var(--font-scale-l)`,
+                    ['--autocomplete-input-placeholder-size' as string]: `var(--font-scale-l)`,
+                }}
+            >
+                <Content value={module.title} />
```
stevenbister, line 36: Think this need to be `module.description` now?

packages/osc-ecommerce/app/queries/sanity/fragments/modules/search.ts
```diff
@@ -0,0 +1,11 @@
+import groq from 'groq';
+import { ROW_SETTINGS } from './rowSettings';
+
+export const MODULE_SEARCH = groq`
+    _key,
+    _type,
+    inputColor,
+    searchText,
+    title,
```
stevenbister, line 9: Think this needs to be updated to description now? To match the new schema name

stevenbister (changes requested):
Looking good man! Couple of (hopefully minor) things. There is an issue with the search route on the deployed link however. I've left a comment a little further down about this

packages/osc-ecommerce/app/root.tsx
```diff
@@ -160,6 +162,8 @@ export const loader: LoaderFunction = async ({ request, context }) => {
     const cart = cartId ? await getCart(context, cartId) : undefined;
 
     return json<LoaderData>({
+        ALGOLIA_API_KEY_SEARCH: process.env.ALGOLIA_API_KEY_SEARCH,
```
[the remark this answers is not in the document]
mattwendzina: Sorted

packages/osc-ecommerce/app/routes/search.tsx
```diff
@@ -38,23 +39,20 @@ export const links: LinksFunction = () => {
...
-        ALGOLIA_PRIMARY_COLLECTIONS_INDEX: process.env.ALGOLIA_PRIMARY_COLLECTIONS_INDEX!,
-        ALGOLIA_PRODUCTS_INDEX_GROUPED_BY_ID: process.env.ALGOLIA_PRODUCTS_INDEX_GROUPED_BY_ID!,
-        ALGOLIA_HITS_PER_PAGE: +process.env.ALGOLIA_HITS_PER_PAGE!,
-        SANITY_STUDIO_API_TOKEN: process.env.SANITY_STUDIO_API_TOKEN!,
-    };
+    // If any algolia envVars are undefined throw error
+    for (const envVar of Object.keys(algoliaSearchPageEnvVars)) {
+        if (!algoliaSearchPageEnvVars[envVar as keyof typeof algoliaSearchPageEnvVars]) {
+            console.error(`${envVar} is not defined!`);
+            throw new Response('Not found', { status: 404 });
```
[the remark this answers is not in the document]
mattwendzina: Ah yeah, looks like I'd forgotten to add the Collections env var! Updated 👍

packages/osc-ecommerce/app/components/Autocomplete/AutocompleteInlineSearch.tsx
mattwendzina: Oops 😬 Good spot. Forgot to update the query too 😅  All sorted now and should be pulling the description through..!

packages/osc-ecommerce/app/queries/sanity/fragments/modules/search.ts
mattwendzina: 👍

packages/osc-ecommerce/app/routes/search.tsx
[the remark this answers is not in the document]
stevenbister: Everything looks good man! I'm still getting a 404 on this page though :/
https://pr-1277-osc-ecommerce.fly.dev/search

mattwendzina:
Finally... this now appears to all be working in https://pr-1277-osc-ecomgitmerce.fly.dev 😅 

Turns out I had double curly brackets for the docker env vars, and the collections env var needed adding to the secrets.

stevenbister (approved):
LGTM! 🚢

Issue #1011, opened 2023-06-08
Scheduled Sales 


EnaamAlum:
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/1009

### Description

n/a

stevenbister:
Apps suggested to Marketing -- now waiting for feedback

stevenbister:
Sharnie has approved NA Bulk Editor app
![image](https://github.com/Open-Study-College/osc/assets/23461173/26435bd4-4167-4b9a-8424-e7bc3cea6274)

PR #1284, opened 2023-07-19
Feature/integrate badge component

stevenbister:
## 📝 Description

Doesn't fully close the issue but is part of #1011. This integrates badge component based on the schedules sales process marketing will be using.

When a sale is set up on the store marketing will bulk tag the products on that sale with the tag `sale:[name of sale]` for example `sale:10% off` or `sale:black friday`. 

image
(this is using the NA Bulk Price editor app but may change depending on Marketing's preference)

Either before or after they will be able to go into Sanity to style the badge if the sale is a special event (default badge style is pink)

image

We can then match the name of the tag against this setting in Sanity and apply the badge styles
image

N/a

## 🚀 New behavior

- Adds `SalesBadge` component into osc-ecommerce
- Adds `sale` document type to Sanity
- Refactors the product/collection type interfaces from Sanity


packages/osc-ecommerce/app/components/UrgencyComponents/SaleBadge.tsx
```diff
@@ -8,6 +9,20 @@ interface SaleBadgeBadgeProps {
...
+
+    const classes = classNames(
+        'c-sale-badge',
+        matchingTag
+            ? `u-bg-color-${matchingTag?.backgroundColor} u-color-${matchingTag?.textColor}`
+            : ''
+    );
+
+    return (
+        <Badge variant={matchingTag ? matchingTag?.variant : 'primary'} className={classes}>
```
jkellerman, line 24: Just a small thing, not sure if you are aware but we are getting a type error here, any idea why?

jkellerman (changes requested):

packages/osc-ecommerce/app/components/UrgencyComponents/SaleBadge.tsx
stevenbister: If you run `build:osc-ui` and restart your ts server does it go away? Until I did that it was trying to pick up the old variants type which was `'subtle' | 'outline'` rather than `'primary' | 'secondary'`

jkellerman (approved):
LGTM

PR #1286, opened 2023-07-20
Feature/third party scripts

### LINKED ISSUE #1015: [S82.1]: Live Chat Widget
### Link to Epic

Epic: https://github.com/Open-Study-College/osc/issues/669

### Description

_No response_

### Linked Tasks

- [ ] #1016 
- [ ] 
- [ ]

### LINKED ISSUE #1016: Live Chat Widget Testing
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/1015

### Description

n/a

stevenbister:
- Closes #1015 
- Closes #1016 

## 📝 Description

Adds the ability for cms users to add third party scripts to the ecommerce site. This includes things like the hubspot chatbot. 
Have confirmed that the scripts are firing with the Hubspot tracking
image

N/a

## 🚀 New behavior

- Adds script input area to Sanity
- Adds `RawScript` component to render script tags entered through Sanity

## 📝 Additional Information

The chatbot won't show up on localhost or the preview urls AFAICT the domains need to be linked to Hubspot? Might be wrong though!


packages/osc-ecommerce/app/root.tsx
```diff
@@ -221,12 +225,17 @@ const Document = ({ children }: DocumentProps) => {
...
+                {headerThirdPartyScripts ? <RawScript html={headerThirdPartyScripts} /> : null}
             </head>
             <body>
                 {children}
                 <ScrollRestoration />
                 <Scripts />
                 <LiveReload />
+                {/* Scripts from Sanity */}
+                {footerThirdPartyScripts ? <RawScript html={footerThirdPartyScripts} /> : null}
+                {/* {footerThirdPartyScripts ? parse(footerThirdPartyScripts) : null} */}
```
mattwendzina, line 238: Is this commented out line meant to be there?

mattwendzina (commented):
Hey mate, 
Looking good. Couple of comments:
I tried adding a couple of simple console log scripts to test:
First one worked fine:
```
 <script>
        console.log("hello");
    </script>
 ```
Adding a second worked seemed not to work at first, I tried putting a semi colon after the first incase it needed some kind of break, this worked but did something weird. It put the second script at the top of the body tag:
Screenshot 2023-07-28 at 07 55 42
Don't know if this could lead to bugs but it could but might be quite difficult to debug if it did? I actually then removed the semi-colon and then it seemed to work fine 🤷‍♂️ 
Looks like anything other than a script tag causes this issue, e.g. I put a comment after the first and then it caused the second to drop into the <body> tag again. Maybe it's common sense to literally just have the script tag and nothing else, but maybe worth putting a comment to say that it shouldn't have anything outside of the script tags? Especially if marketing themselves might be using this? 😅 

I'm getting a server/client hydration issue = /
Screenshot 2023-07-28 at 08 06 40

Wondering if there is a better UI to have rather than just one text box for all the scripts? Just thinking it could get a bit unwieldy and difficult to follow if we ended up needing to inject a few, especially if they have a lot of content. Would there be a way of potentially having one script per text box? Would be easier/cleaner to follow this way I would think.
Appreciate time is against us, so this might not be practical at this stage..!

mattwendzina:
I still had to add in the bugfix although it's been merged, so looks like it needs a rebase now.

mattwendzina (changes requested):
Sorry meant to hit request changes, but hit comment instead!

packages/osc-ecommerce/app/root.tsx
stevenbister: Lol no 😅

stevenbister:
Ugh, don't think this is gonna work very well with React and SSR :/ I think we're probs best off inserting ourselves and using something like GTM to handle adding scripts instead 🤔 Gonna update this PR a bit to do that instead I think

stevenbister:
Okay, changed how this will work. I think we're better off installing the scripts and passing in the ID's. These IDs are in the env but I think they're better off in this case coming out of Sanity, so marketing can change them if needs be, and we can easily change whether we point at sandbox or live without having to set up bunch of logic or new Docker/GH action nonsense

mattwendzina (commented):
Yeah, this seems like a good approach. 

LGTM!

mattwendzina (approved):
🚢

PR #1308, opened 2023-07-24
Bugfix/remix app fonts

stevenbister:
## 📝 Description

Fixes an issue where fonts aren't being pulled into the `pubilc/build/_assets` folder correctly in the ecommerce & academic hub apps.
This seems to be related to the fact that our current dev command runs the build command first and then follows up with dev in parallel. I think this was causing the fonts to get added by one of the commands but then removed again when the other one starts. I've removed the build step from the dev command, as I don't see why we need to do this, and it _seems_ to have sorted the issue.

Fonts not loading and fallback font being displayed
image

## 🚀 New behavior

Correct font files loading
image

## 📝 Additional Information

This is a dev only update so probably no need to build the app


mattwendzina (commented):
Yay, fonts working again 😄 
LGTM!

mattwendzina (approved):

PR #1309, opened 2023-07-24
Bugfix/sanity invalid preview

### LINKED ISSUE #792: 'Invalid Preview Config' when body text in Content is left empty
### Link task to story

Story: 

### Description

If you add a Content item in Sanity but don't set the body text then you get the error - 'Invalid Preview Config':
Screenshot 2023-03-15 at 11 23 55

The error in the console is:
`TypeError: Cannot read properties of undefined (reading '0')`

Presumably it is expecting there to be a body text. So if this is required then we need a validation error if it isn't set, otherwise need to work out how to make it happy if no text is entered.

stevenbister:
- Closes #792 

## 📝 Description

Fixes issue where the module in Sanity would display 'Invalid Preview Config' if the body content was empty in certain modules. I've added a ternary to show some fallback text instead which should make for a nicer ux.

If the content body is empty this message gets displayed
image

## 🚀 New behavior

If the content body is empty fallback text will be displayed
image


mattwendzina (commented):
Looks good. Do you think `textGrid.ts` could do with this too? It looks like it follows the same pattern as the other two.

mattwendzina (changes requested):
Sorry same as last one, hit comment rather than request changes!

stevenbister:
> Looks good. Do you think `textGrid.ts` could do with this too? It looks like it follows the same pattern as the other two.

Sorry mate I think I'm missing something? The text grid will display "Untitled" if the title is left empty and I've updated the `textGridItem` to do the same as the `content`. Are you seeing something else?

image
image

mattwendzina:
> > Looks good. Do you think `textGrid.ts` could do with this too? It looks like it follows the same pattern as the other two.
> 
> Sorry mate I think I'm missing something? The text grid will display "Untitled" if the title is left empty and I've updated the `textGridItem` to do the same as the `content`. Are you seeing something else?
> 
> image image

Sorry mate, I meant `detailsGrid.ts` 🤦‍♂️

stevenbister:
> Sorry mate, I meant `detailsGrid.ts` 🤦‍♂️

Aha! No worries! Good spot! I missed that one

mattwendzina (commented):
LGTM!

mattwendzina (approved):
🚢 🚢

PR #1333, opened 2023-07-25
Feature/seo structured data

### LINKED ISSUE #1293: Check/add structured data
### Link task to story

Story: #1290

### Description

Check/add structured data on page templates (products, collections, blog posts etc.)

stevenbister:
- Closes #1293 

## 📝 Description

Adds the json+ld data to the routes in ecommerce. I noticed that although the hydrogen docs say you need to use the `handle` to pass data, the demo store actually straight up passes it out of the loader, I quite like this pattern as it lets us move most of the SEO settings out of the routes and keep them a bit tidier, especially for pages like collections & products where there's a bit more to them. 
So I've moved most of the settings into the `seo.server.ts` file and we are now building the meta tags etc. there. I've left the cart and search pages as they are as we're not setting a lot in there other than the page title.

- Seo settings for all routes are handled by the `SeoHandleFunction`
- No structured data (json+ld) data is present

## 🚀 New behavior

- Seo settings for all routes except for cart and search are now generated by the `seoPayload` exported from `seo.server.ts`
- Adds the json+ld data to the routes in ecommerce.


packages/osc-ecommerce/app/utils/server/seo.server.ts
```diff
@@ -0,0 +1,270 @@
...
+        twitter && twitter?.socialProfile.substring(twitter?.socialProfile.lastIndexOf('/') + 1);
+
+    return {
+        titleTemplate: `%s ${seoSettings?.titleSeparator} ${seoSettings.siteTile}`,
+        handle: `@${twitterHandle}`,
+        robots: seoSettings.robots,
+        jsonLd: {
+            '@context': 'https://schema.org',
+            '@type': 'Organization',
+            name: seoSettings.schema.organizationName,
```
mattwendzina, line 41: A lot of this file is using optional chaining but there are few places where it isn't. Are these where you forgotten to add some or are you confident that they just won't be needed where you haven't added them?

packages/osc-ecommerce/app/routes/courses/$slug.tsx
```diff
@@ -145,6 +145,11 @@ export const loader = async ({ request, params, context }: LoaderArgs) => {
         request,
     });
 
+    const seo = seoPayload.product({
```
mattwendzina, line 148: Seem to be getting a TS error here.

mattwendzina (changes requested):
Looking good mate. Left a couple of comments, and for some reason the courses page seems to be crashing 🤔 Getting the below error:

TypeError: Cannot read properties of undefined (reading 'variants')

packages/osc-ecommerce/app/utils/server/seo.server.ts
stevenbister: Think I just forgot to add them! Optional chaining is one of those things that doesn't come very naturally lol

mattwendzina: Lol, I get it! 
So I'm guessing the ones now left are those you're confident about not needing?
e.g. line 30, 32, 35, 37, 43
Sorry if it wasn't obvious I was meaning these ones too. I just left the comment on one example 😅

mattwendzina (changes requested):
Just added a final comment on the optional chaining just to make sure we're completely good there.

packages/osc-ecommerce/app/utils/server/seo.server.ts
stevenbister: Oops missed 35! But yeah the others should always return something, and 30 & 32 should get caught by that `&&`. Probably couldn't hurt to include them there as well though so I'll add them in

packages/osc-ecommerce/app/utils/server/seo.server.ts
```diff
@@ -27,20 +27,20 @@ const root = ({
     canonicalUrl: Request['url'];
 }): SeoConfig<Organization> => {
     const socials: SanitySocial[] = seoSettings?.socials;
-    const twitter = socials && socials.find((social) => social.socialProfile.includes('twitter'));
+    const twitter = socials && socials?.find((social) => social?.socialProfile.includes('twitter'));
```
mattwendzina, line 30: Sorry, at risk of starting to sound anal about this mate 😅 Not sure optional chaining would do much there as that is just the item on the array. I would think if anything we would want to check that socialProfile was a property on the array to avoid an error on the `includes`?

stevenbister: Haha nah you're alright I'd probably be the same 😂 Think I was trying to do too many things at once and put the `?` in the wrong place 😅 I've moved it over to check before the `.includes` which should catch in the right place now. I've also added a ternary to line 43 just to be on the safe side there. The data should exist in Sanity now so probably won't be an issue but it's there _just_ incase

mattwendzina: Cool cool

mattwendzina (approved):
LGTM!

PR #1335, opened 2023-07-25
Feature/jobs pages

### LINKED ISSUE #1233: Page: Jobs
### Link task to story
Story: #1230

### Description
- Create a Jobs page view with the following components: Header, Navigation, Hero, Content, Course Cards (adapted for this app), Filtering Grid (with category filtering, pagination/progress), and Footer.

- Implement a Title component for the Jobs page.
- Implement a Description component for the Jobs page.
- Add an Apply Button component (with email link) to the Jobs page.

jkellerman:
Closes #1233 

## 📝 Description

Created jobs page with job filtering with links to individual job page

N/A

## 🚀 New behavior

N/A


packages/osc-ui/src/components/Card/card.scss
```diff
@@ -216,6 +223,17 @@ $card-fs-3xl: $fs-3xl;
             }
         }
 
+        &__align {
```
stevenbister, line 226: Couple of things on this one.
Something like this should be a modifier rather then an element (so `--align` instead) as it's changing some behaviour a bit like your `--row` modifier above.

Some of this could be extracted out as there's a few different behaviours happening here (flex/spacing/font-size), where you'd want a modifier to ideally only be doing one thing. Could these be handled by utility classes instead?

e.g. (you might have to check these classes actually exist lol)
```tsx
 <CardBodyInner className="o-flex o-flex--stack t-font-m u-mt-m">
```

stevenbister (changes requested):
Covered a lot off in here, looks good! Left a few comments/suggestions for you, hopefully nothing too major

packages/osc-recruitment/app/routes/jobs/$slug.tsx
```diff
@@ -70,7 +87,35 @@ export default function Job() {
                     <PagePreview query={query} params={params} Component={PageContent} />
                 </PreviewSuspense>
             ) : ( */}
-            <PageContent {...page} />
+
+            <div className="u-bg-color-septenary u-pb-2xl u-pt-2xl">
+                <div className="o-container">
+                    <h1>{jobDetails.page.title}</h1>
+                    <ul>
+                        <li>{jobDetails.page.jobDetails.positionType}</li>
```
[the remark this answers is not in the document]
jkellerman: As discussed, we're going to leave this as it is because JobDetails is an object, not an array, which will require us to use a regex pattern to get the key of each key:value pair. Also looping through an object gives us no control of the order of items in list, which we would get when looping through an array. We also don't anticipate anything being added to this list in future.

packages/osc-recruitment/app/routes/jobs/index.tsx
```diff
@@ -67,8 +125,32 @@ export default function Index() {
...
+
             <PageContent {...jobs} />
             {/* )} */}
+
+            <Form
+                className="o-container o-row--2xl o-flex o-flex--center"
+                onChange={handleOptionSelect}
+            >
+                <Select
+                    className={` u-pt-xs c-select--width o-flex--center`}
```
stevenbister, line 137: Should probably try and find an eslint rule for this 😅 But if you're not interpolating any values can you try and avoid setting props like this. Ideally it should just be a string
```suggestion
                    className="u-pt-xs c-select--width o-flex--center"
```

packages/osc-recruitment/app/routes/jobs/index.tsx
```diff
@@ -67,8 +125,32 @@ export default function Index() {
                     <PagePreview query={query} params={params} Component={PageContent} />
                 </PreviewSuspense>
             ) : ( */}
+
             <PageContent {...jobs} />
             {/* )} */}
+
+            <Form
+                className="o-container o-row--2xl o-flex o-flex--center"
```
stevenbister, line 133: If you're using an `o-row` modifier you should also be including the `o-row` class alongside it. Could this not be a utility class instead maybe?

packages/osc-recruitment/app/routes/jobs/index.tsx
stevenbister, line 137: Also on this one, I'm not super keen on the `--width` modifier class. Feels a bit like it opens doors to loads of adhoc classes, what if we need one that's it to be 250px for example 🤔 

Rather than setting things up like this you could use the grid classes to lay things out a bit like this

```tsx
<Form
    className="o-container o-row--2xl o-flex o-flex--center o-grid"
    onChange={handleOptionSelect}
>
    <Select
          className="u-pt-xs o-grid__col o-grid__col--2 o-grid__col--start-6"
          name="category"
          groupVariants={['secondary', 'inline-wrap']}
          defaultValue={categories[0]?.title.toLowerCase()}
      >
          etc...
```

You might need to adjust slightly so it doesn't get too small on mobile but I think you get the idea

stevenbister (changes requested):

packages/osc-recruitment/app/routes/jobs/index.tsx
jkellerman: When I use grid, it doesn't centre well in the viewport. Also it needs some sort of width on the select element otherwise the select element will shrink with grid in mobile and the text will overflow the element. The default styles for the select also inherits the width of its parent. We could do something like this instead:
```js
            <Form
                className="o-container o-row o-row--2xl o-flex o-flex--center"
                onChange={handleOptionSelect}
            >
                <div style={{ width: '200px' }}>
                    <Select
                        className="u-pt-xs"
                        name="category"
                        groupVariants={['secondary', 'inline-wrap']}
                        defaultValue={categories[0]?.title.toLowerCase()}
                    >
                        {categories.map((option: categoryModule) => (
                            <SelectItem value={option.title.toLowerCase()} key={option._key}>
                                {option.title}
                            </SelectItem>
                        ))}
                    </Select>
                </div>
            </Form>
```

What you think?

stevenbister: So i guess I envision the full markup to look like this
```tsx
            <Form
                className="o-container o-row o-row--2xl o-flex o-flex--center"
                onChange={handleOptionSelect}
            >
                    <Select
                        className="u-pt-xs"
                        size="sm"
                        name="category"
                        groupVariants={['secondary', 'inline-wrap']}
                        defaultValue={categories[0]?.title.toLowerCase()}
                    >
                        {categories.map((option: categoryModule) => (
                            <SelectItem value={option.title.toLowerCase()} key={option._key}>
                                {option.title}
                            </SelectItem>
                        ))}
                    </Select>
            </Form>
```

packages/osc-recruitment/app/routes/jobs/index.tsx
```diff
@@ -67,8 +125,32 @@ export default function Index() {
...
+            >
+                <Select
+                    className="u-pt-xs o-flex--center"
+                    size="sm"
+                    name="category"
+                    groupVariants={['secondary', 'inline-wrap']}
+                    defaultValue={categories[0]?.title.toLowerCase()}
+                >
+                    {categories.map((option: categoryModule) => (
+                        <SelectItem value={option.title.toLowerCase()} key={option._key}>
```
stevenbister, line 144: Think you need to check this `option._key` is returning a value. Getting an error in my console saying this is missing a `key` prop

packages/osc-ui/src/components/Select/select.scss
```diff
@@ -95,6 +101,10 @@
             }
         }
 
+        &--width {
```
stevenbister, line 104: Can we remove this now we're not using it?

stevenbister (changes requested):
Nice work! Last couple of things I've noticed and I think we're good to go

stevenbister (approved):
LGTM!

PR #1339, opened 2023-07-26
Feature/career kickstart package

### LINKED ISSUE #916: [EP17]: Basket Page
### Description

n/a

### LINKED ISSUE #962: [S68.1]: Featured Products
### Link to Epic

### Description

_No response_

### Linked Tasks

- [ ] #963 
- [ ] #964
- [ ]

### LINKED ISSUE #963: Integrate the callout
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/962

### Description

n/a

### LINKED ISSUE #964: Add career kickstart as featured product in cart
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/962

### Description

n/a

stevenbister:
- Closes #916 
- Closes #962
- Closes #963 
- Closes #964 

## 📝 Description

Adds the career kickstarter callout item to the cart. This is specifically for career kickstart so doesn't handle other addons.
This should link up with the career kickstart addon item in the sidebar (on desktop) and each one will update when it is added to the cart, (once the fetcher has completed). This will always be the second item in the basket as per the design file.

https://github.com/Open-Study-College/osc/assets/23461173/5441402d-b30a-4154-80a6-36a48f07bb55

N/a

## 🚀 New behavior

- Career Kickstart addon component


mattwendzina (changes requested):
Looks great mate. I think it would be a nicer UX if both checkboxes changes at the same time. It seems a bit strange that one ticks and you have to wait for the other - as in I get why, but to a user it might seem a bit odd. Presumably there isn't local state on these components that can easily be passed around, and we waiting on a network request to update the other button? 
Wouldn't block it, but just checking to see whether it is easily achievable/worth having the checkboxes update at the same time?

stevenbister:
> Looks great mate. I think it would be a nicer UX if both checkboxes changes at the same time. It seems a bit strange that one ticks and you have to wait for the other - as in I get why, but to a user it might seem a bit odd. Presumably there isn't local state on these components that can easily be passed around, and we waiting on a network request to update the other button? Wouldn't block it, but just checking to see whether it is easily achievable/worth having the checkboxes update at the same time?

No, that's a good point! I've added a provider and `useContext` so we can access the state throughout the cart. It's pretty specific but I think that's okay in this case, we can always refactor later if we need more flexibility. 
Only thing with this I noticed is that if you have the career kickstart selected and then refresh the browser it fires off another POST to the cart. It won't add another item due to how it's setup but it does set off the loading state. 

It's something to do with keeping the sync between the provider state and what's actually in the cart. I dunno if it's last thing on Friday brain but I can't figure out a way around this. I don't think it's the end of the world, I imagine it's unlikely people will be adding this to their cart and going away adding another product and coming back 🤔 A bit on the fence about it. What do you think?

mattwendzina:
Hmmm... had a look to try and understand. Looks like yeah because `kickStartAddOn` state always starts as false, so when it gets updated it triggers the onChange method. 
It seems like the useCartContext needs to somehow get updated before it hits `CartAddOns.tsx` or `CareerKickstartCallout.tsx` so that when it reaches these component it is already set to true if it has been added.

Annoyingly there is another weird bug off the back of it which isn't great lol. If you actually view on mobile, and refresh the page, it doesn't actually tick the box at all. The price will be correct but the checkbox will be unchecked... Guessing this must be something to do with the ternary in the onValueChange in `CartAddOns.tsx`.

I think if it's too much faff for now then maybe we should leave it and perhaps add a ticket to optimise this further down the line..?

stevenbister:
Thanks mate, that comment actually helped! I realised that I can just lift a bunch of the logic in the `CartAddons` into the provider and pass that into the state, which seems to have done the trick 🎉  

I've tested on my iPhone and it seems to have fixed the mobile issue you mentioned, would you mind checking on yours too?

mattwendzina:
> Thanks mate, that comment actually helped! I realised that I can just lift a bunch of the logic in the `CartAddons` into the provider and pass that into the state, which seems to have done the trick 🎉
> 
> I've tested on my iPhone and it seems to have fixed the mobile issue you mentioned, would you mind checking on yours too?

I'm still getting the same behaviour unfortunately - loading occurs when you refresh the page, and it still disappears if you're in mobile view 😢

Issue #1344, opened 2023-07-28
[EP23]: Implement Etika Embeddable


hi-im-si:
### Description

This will involve adding an Etika Financial Calculator to the ecommerce store.

PR #1351, opened 2023-07-30
Feature/etika finance calculator button

### LINKED ISSUE #1342: Control Etika Financial Calculator embeddable behind Button Type
### Link task to story

Story: 

### Description

To allow full flexibility on presenting etika's financial calculator embeddable throughout the ecommerce store we would want to extend our current Button component to support a "finance calculator" button type. This can then be used anywhere throughout ecommerce. This will of course need to be controllable from within the Sanity CMS

stevenbister:
- Closes #1342 
- Closes: #1344 

## 📝 Description

WIP!
Placeholder PR to generate preview URL for etika to add token

## 🚀 New behavior


packages/osc-ecommerce/app/components/Content/Content.tsx
```diff
@@ -1,24 +1,49 @@
...
+const Content = (props: { module: contentModule }) => {
+    const { module } = props;
+
+    return module.body ? (
+        <>
+            <OSCUIContent align={module.horizontalAlignment} value={module.body} />
+
+            {module.buttons && module.buttons.length > 0 ? (
+                <ButtonGroup>
+                    {module?.buttons?.map((button) => (
```
mattwendzina, line 42: Don't think this optional chaining is necessary as you've just done a check to make sure buttons exists and has a length that is greater than 1.

packages/osc-ecommerce/app/components/Button/Button.tsx
```diff
@@ -118,6 +119,9 @@ export const Button = (props: Props) => {
                 </CopyButton>
             ) : null;
 
+        case 'finance calculator':
+            return <EtikaFinanceCalculator />;
```
mattwendzina, line 123: From what I can understand here, we're not passing any props through from the Sanity data. In Sanity, when you select the finance calculator it asks for an embed code - I can't see where this gets used?
And presumably all of those the other button properties don't apply, e.g. inverse/variant etc... - is it worth hiding them when the finance calculator button is selected so that it's not misleading?

mattwendzina (changes requested):
Hey mate, left a couple of comments on here for ya.

packages/osc-ecommerce/app/components/Button/Button.tsx
stevenbister: Ooops forgot that was in there! And yeah good shout. Done

PR #1353, opened 2023-07-30
Bugfix/product page responsive

stevenbister:
## 📝 Description

Found a bug on the product page where the tabs weren't responding to the width of the viewport correctly, causing some overflow issues.

e.g. 
image

I narrowed this down to the `display: grid` property on the `o-page` class. Not entirely sure why but that seemed to be what was causing the issue. I have replaced this with `display: flex` instead and while this is a little more css, as we will need to target the footer specifically, the behaviour seems to be a bit more reliable.

- `o-page` is using `display: grid` with `grid-template-rows` to ensure the footer is always at the end of the page, regardless of how much content is there.

## 🚀 New behavior

- `o-page` is using `display: flex` with a `o-page__footer` element to push the footer down.


jkellerman (approved):
LGTM

PR #1354, opened 2023-07-31
Bugfix/header button styles

stevenbister:
## 📝 Description

Fixes an issue where the badge on the minicart floats away from the button container when on a page that doesn't have any button styles imported. e.g.
image

The buttons styles we're only being imported on certain pages/when certain modules were present, this meant that empty pages or pages without buttons had this behaviour. I've imported the button styles as a global import on the `root.tsx` file and removed any route/dynamic imports to prevent duplication.

image

## 🚀 New behavior

image


jkellerman (approved):
LGTM

PR #1371, opened 2023-07-31
feature/auth DX

Tom-Pearce:
Closes # 

## 📝 Description

Series of Developer Experience improvements for the Auth API

## 🚀 New behavior

Alphabetised GQL Schema/Resolver for easier dev navigation.
Consolidated registration token verification queries into a single `verifyRegistrationToken` query.
Renamed `createUserSetup` to `createStudent` as it's student specific.
Made password optional for `createUser` so we can now use this as the internal admin user creation.


packages/osc-api-auth/src/utils/account.ts
```diff
@@ -117,29 +116,24 @@ export const assignRole: assignRoleFn = async (userId, roleId) => {
...
         return new Error('No valid login link');
     }
-    const user = await getUserByEmail(input.email);
 
-    if (!user) {
-        return new Error('No matching active user found');
-    }
-
-    const userUpdate = UpdatePassword({ userId: user.id, password: input.password });
+    const userUpdate = UpdatePassword({ userId: userId, password: input.password });
```
hall2521, line 136: probably worth adding an await here

packages/osc-api-auth/src/utils/account.ts
```diff
@@ -117,29 +116,24 @@ export const assignRole: assignRoleFn = async (userId, roleId) => {
...
     return userUpdate;
```
hall2521, line 137: you can probably remove this line and return the function in the line above

hall2521 (changes requested):

hall2521 (approved):
LGTM

PR #1384, opened 2023-08-03
Feature/bloom connections styling

jkellerman:

- Closes #

## 📝 Description
This branch:

- adds all the styling for the recruitment website
- fixes the issue in the hero component where the hero title does not remain centred when the hero body is removed in Sanity CMS
- Changes the utility class prefixes for border to enable consistency of naming

This is the bug we have in the hero component. When the hero body is removed in the CMS, the title will push to the top of the grid.
Screenshot 2023-08-04 at 14 13 26

Screenshot 2023-08-04 at 14 08 29

## 🚀 New behaviour

When there is no body content selected in the CMS, it should now be in the  centre of the grid as shown below.
Screenshot 2023-08-04 at 14 03 47

**Breaking change**

- References to the utility class u-b-bottom or u-b-none in ecommerce will need to be changed to u-bb-3 or u-bb-none.
 

## 📝 Additional Information

- For the hero bug fixes to applied in ecommerce, the HeroContentWrapper component needs to be implemented the same way it is in recruitment, see below.

```
 <HeroInner
                flourishColor={copyFlourishes?.color}
                flourishPattern={copyFlourishes?.pattern}
                flourishVariant={copyFlourishes?.variant}
            >
                {variant === 'tertiary' ? (
                    <HeroContentWrapper>
                        <HeroTitleGroup>
                            <Title />
                        </HeroTitleGroup>
                        {content?.body ? (
                            <HeroContent>
                                <Content
                                    value={content?.body}
                                    align={content?.horizontalAlignment}
                                    {...content}
                                />
                            </HeroContent>
                        ) : null}
                    </HeroContentWrapper>
                ) : (
                    <>
                        <Title />
                        {content?.body ? (
                            <HeroContent>
                                <Content
                                    value={content?.body}
                                    align={content?.horizontalAlignment}
                                    {...content}
                                />
                            </HeroContent>
                        ) : null}
                    </>
```


packages/osc-ui/src/styles/utilities/_util.scss
```diff
@@ -96,6 +96,12 @@ $osc-config: (
             scales-map()
         )
     ),
+    border: (
+        prefix: "bd",
```
stevenbister, line 100: I would probably have this as just "b" rather than bd so it's consistent with the other utils

packages/osc-recruitment/app/components/Hero/Hero.tsx
```diff
@@ -118,24 +118,25 @@ const Slide = (props: SlideProps) => {
                 flourishPattern={copyFlourishes?.pattern}
                 flourishVariant={copyFlourishes?.variant}
             >
-                {variant === 'tertiary' ? (
-                    <HeroTitleGroup>
+                <div className="o-grid__col o-grid__col--12 o-grid__col--5@desk ">
```
stevenbister, line 121: Could we move the fixes you've put into the hero into osc-ui? This is an issue on the ecommerce site as well so would be great if we could fix it across both sites

packages/osc-recruitment/app/routes/jobs/index.tsx
```diff
@@ -30,6 +31,7 @@ export const links: LinksFunction = () => {
         { rel: 'stylesheet', href: contentStyles },
         { rel: 'stylesheet', href: buttonStyles },
         { rel: 'stylesheet', href: selectStyles },
+        { rel: 'stylesheet', href: recruitmentStyles },
```
stevenbister, line 34: I don't think you need this in here. You've already declared it in your `root.tsx` so it will filter through

packages/osc-recruitment/app/styles/objects/_page.scss
```diff
@@ -0,0 +1,13 @@
+@import "../../../../osc-ui/src/styles/settings";
+
+.o-page {
+    padding: $space-xl;
```
stevenbister, line 4: I wonder if this needs a media query? Feels a bit narrow on my phone and when the menu opens it doesn't align with the edges of the header
image

Perhaps on smaller devices it fills the width of the viewport and only when we get to the larger screens does it get the padding? More like this
image

stevenbister (changes requested):
Looking good! Left a few comments, nothing major. 

Just check the PR I think we're missing things like assignee, project status, milestone etc. and it would be good if you could fill out the description with some more details. For example a brief explanation as to why have we added the overrides or the fact you're fixing the tertiary hero.
It's good make a record of these things so if we need to go back to this PR later it would be easier to see what and why things have changed.

Could you also add the `build:osc-studio` label so we can deploy these Sanity changes.

packages/osc-recruitment/app/styles/objects/_page.scss
jkellerman: Yeah was thinking this myself, works better with no padding in mobile

jkellerman:
This feature branch hasn't been assigned to a sprint so it should be left as no milestone right?

packages/osc-recruitment/app/styles/objects/_page.scss
stevenbister: Ah yeah loads better!

packages/osc-ui/src/components/Hero/hero.scss
```diff
@@ -49,7 +49,7 @@
             row-gap: 0;
 
             @include mq($mq-mob-l) {
-                grid-template-rows: $hero-grid-outer repeat($hero-inner-grid-rows, auto) $hero-grid-outer;
+                grid-template-rows: 1fr;
```
stevenbister, line 52: Could you check how the hero looks in storybook? These grid rows and the other changes are needed for the other variants.
I think you'll need to change this value further down so it' nested within the `&--tertiary` modifier, same with the `ttl__group` class maybe?

stevenbister (changes requested):
Nice! Thanks for those changes :) Last one. Just make sure to double check the component in Storybook. If you're making edits to osc-ui components I would recommend doing it in Storybook so you can see how the changes effect each story for that component

stevenbister:
> This feature branch hasn't been assigned to a sprint so it should be left as no milestone right?

It still part of the work in this sprint though, isn't it? Even though it doesn't have a task

stevenbister (changes requested):
Looks really good! Sorry I missed this on the last one (so this should actually be the last change now 😅 )
Could you update any of the `u-b-bottom` or `u-b-none` in osc-ecommerce to the new classes? Don't want to push up a breaking change if we can help it!

stevenbister (approved):
LGTM!

PR #1398, opened 2023-08-10
Feature/ah login functionality

stevenbister:

- [ ] Close issues
- [x] Request reviewers
- [x] Assign yourself
- [x] Add project
- [x] Set project status
- [ ] Add milestone

---

- Closes: https://github.com/Open-Study-College/osc/issues/1427
- Closes: https://github.com/Open-Study-College/osc/issues/1428

## 📝 Description

Adds the login functionality to the login form on academic hub. I've added the [remix-auth](https://www.npmjs.com/package/remix-auth) package to handle user authentication, using the form strategy. This should give us the flexibility to use other strategies such as login with Google/Facebook/Microsoft etc.
This basically works by setting the user access token in a cookie which can then be checked using the `authenticator.isAuthenticated` method. We are then using this to redirect people to the dashboard when they're logged, or direct them to the login page if they are not.

I've added a codegen to the package which should help us generate types based on the queries, rather than us having to manually match them up. We can run `npm run generate:osc-academic-hub` to generate the types which will get stored in `app/types/__generated__/`. I've committed these but as they're autogenerated don't worry about looking at them too hard.

Finally I've added a few unit tests, testing Remix hooks/components isn't possible without mocking them currently as they require context that we don't have access to. So the login form test only really mocks and tests the validation.
Similarly mocking responses in Remix is pretty difficult as the queries will run on the server and currently there's no easy way to handle these without them being globally mocked (you can't change the mock on a test by test basis see: https://github.com/Open-Study-College/osc/pull/1165#issuecomment-1600532864), meaning e2e tests would be the way to go. However until we have a test database and some seeding scripts e2e tests could become flakey as the data in the databases currently are all manually added. I've put a pin in this for now and perhaps it could be something we introduce later down the line.

N/a

## 🚀 New behaviour

- Adds login & logout functionality
- Adds `useUser` hook to display user data on front end
- Adds Zod for form validation
- Adds redirect protection to certain routes if user if not logged in
- Adds codegen and codegen script to generate types
- 

## 🧾 Final Checklist

- [x] I've added tests to cover my changes (Unit, E2E, Integration, Contract, Functional, Acceptance etc)

## 📝 Additional Information

In order to test you will need the new planetscale `.env`s as well as all of the `.env`s for each api package. You will also need a user setting up if you don't have one already.
In order to run the app you will need to run all of the api services as well as the api-gateway (`npm run dev:api`) which should run on `localhost:4000/graphql` you can then run the academic hub app (`npm run dev:osc-academic-hub`).

EDIT: If you already have `API_GATEWAY_URL` in your `.env` you might need to update it from `localhost:4000` to `localhost:4000/graphql`


stevenbister:
Have updated this branch to include the refresh tokens

Tom-Pearce:
Cant seem to get it working. Managed to get past the `darwin` error, but now whenever I attempt to login I get a 404 error.

The error names Graphql/Apollo, making it look like thats the source, but I cannot get the gateway or microservices to log anything when submitting the form making it seem like the request is no even reaching the Gateway.

Error received:
```
{"name":"ApolloError","graphQLErrors":[],"clientErrors":[],"networkError":{"name":"ServerError","response":{"size":0},"statusCode":404,"result":"<!DOCTYPE html>\\n<html lang=\\"en\\">\\n<head>\\n<meta charset=\\"utf-8\\">\\n<title>Error</title>\\n</head>\\n<body>\\n<pre>Cannot POST /</pre>\\n</body>\\n</html>\\n"},"message":"Response not successful: Received status code 404"}
```

stevenbister:
Have updated the description to include details on `API_GATEWAY_URL` in .env will push up an update to the sample env and README shortly

packages/osc-academic-hub/app/components/Auth/LoginForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+                        </Link>
+                    </Alert>
+                </div>
+            ) : null}
+
+            <div className="u-mb-s">
+                <TextInput
+                    id="email"
+                    name="email"
+                    label="Username"
```
Tom-Pearce, line 60: Without looking, I'm guessing it says `Username` in the design, just caught my eye as it's surrounded by multiple `="email"`s

Tom-Pearce (commented):
Can't comment super indepth on the frontend. but what I do understand LGTM

packages/osc-academic-hub/app/components/Auth/LoginForm.tsx
stevenbister: Aha yeah good spot, Email would def make more sense in here. Although there was talk at one point about being able to pass both email and order id (or something similar). Is that still the case?

stevenbister: Have updated :)

Issue #1003, opened 2023-06-07
Article filter 


EnaamAlum:
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/1000

### Description

- Utilise button group
- Algolia powered filters (RefinementList attached to a button group)
- Factor in the Pagination / Load more.

PR #1403, opened 2023-08-11
Feature/blog buttons group

jkellerman:

- Closes #

## 📝 Description

This is part of [Article filter #1003](https://github.com/Open-Study-College/osc/issues/1003) Task. This PR adds the buttons group for the articles filter.

Screenshot 2023-08-11 at 18 22 35

N/A

## 🚀 New behaviour

N/A

## 🧾 Final Checklist

## 📝 Additional Information


packages/osc-ecommerce/app/queries/sanity/categories.ts
```diff
@@ -0,0 +1,11 @@
+import groq from 'groq';
+
+export const CATEGORIES_QUERY = groq`
+    *[ _type == "category" && !(_id in path("drafts.**")) ] {
+        _id,
+        _rev,
```
stevenbister, line 6: Do you plan on using the `_rev` property? I don't think you need to return it?

packages/osc-ui/src/components/Button/Button.stories.tsx
```diff
@@ -131,6 +131,17 @@ Senary.parameters = {
         },
     },
 };
+export const Septenary = Template.bind({});
+Septenary.args = {
+    children: 'button',
+    variant: 'septenary',
+};
+export const SeptenaryDecorator = (Story) => (
```
stevenbister, line 139: Do we need this? It's causing an error in Storybook
image

If you need to add a background colour or some other settings to a story you're better off creating a template like we do above, so something like

```tsx
const SeptenaryTemplate: Story<ButtonProps> = ({ children, ...args }) => (
<div className="...">
    <Button {...args}>{children}</Button>
</div>
);
```
Then replacing the template on line 134
```tsx
export const Septenary = SeptenaryTemplate.bind({});
Septenary.args = {
    children: 'button',
    variant: 'septenary',
};
```

packages/osc-ui/src/components/Button/button.scss
```diff
@@ -187,6 +187,21 @@
                 )
             )
         ),
+        septenary: (
+            bg-color: transparent,
+            bg-hover-color: var(--btn-hover-color-tertiary),
```
stevenbister, line 192: I wonder if it's worth making the hover colours match the active state in the design, so that semi-transparent background
image
I think it would look odd if it was navy and then became pink

packages/osc-ui/src/components/Button/button.scss
```diff
@@ -187,6 +187,21 @@
...
+            fg-color: var(--color-tertiary),
+            fg-hover-color: var(--btn-fg-hover-color-tertiary),
+            font-weight: $fw-reg,
+            text-transform: capitalize,
+            border-bottom: 1px solid var(--color-tertiary),
+            sizes: (
+                lg: (
```
stevenbister, line 199: As there's only one instance of this the only size should be `md` as that's the default `size` for the buttons. Saves you having to remember to pass is as a prop when using the variant then.

packages/osc-ui/src/components/Button/button.scss
```diff
@@ -187,6 +187,21 @@
...
+                    padding: $space-l $space-3xl,
```
stevenbister, line 200: I would reduce the size of the vertical padding to `$space-m`. It'll bring the height of the buttons down to closer match the design. They feel a bit big at the moment.
```suggestion
                    padding: $space-m $space-3xl,
```

packages/osc-ui/src/components/Button/button.scss
```diff
@@ -187,6 +187,21 @@
...
+                    font-size: $fs-m
```
stevenbister, line 201: I think the font is too big as well, looking at the design it should be `16px`, so you can use `--font-scale-base` in this case to bring it to `1rem`
```suggestion
       font-size: var(--font-scale-base);
```

packages/osc-ui/src/components/Button/button.scss
```diff
@@ -351,6 +390,23 @@
...
+                    &:nth-of-type(1) {
+                        @include mq($mq-desk) {
+                            border-left: 1px solid var(--color-tertiary);
+                        }
+                    }
+
+                    @include mq($mq-desk) {
+                        width: unset;
+                        border-right: 1px solid var(--color-tertiary);
+                        border-bottom: unset;
```
stevenbister, line 406: I would prefer to use `none` here rather than `unset`, I think. More of a personal preference as I find it a bit more explicit. `unset` could inherit from somewhere so might end up with some weird behaviour 🤔

packages/osc-ui/src/components/Button/button.scss
```diff
@@ -547,18 +603,32 @@
...
+            flex-direction: row;
+        }
+
         &--column {
             flex-direction: column;
             align-items: flex-start;
         }
 
+        &--none {
+            gap: none;
```
stevenbister, line 616: `none` doesn't exist as a property for `gap` should be `0` instead
```suggestion
            gap: 0;
```

packages/osc-ui/src/components/Button/button.scss
```diff
@@ -547,18 +603,32 @@
         flex-wrap: wrap;
         align-items: center;
 
+        &--row {
```
stevenbister, line 606: Do we need this explicitly being added? `display: flex` defaults to `row` anyway so not sure what benefit we gain by adding this modifier?

packages/osc-ui/src/components/Button/button.scss
```diff
@@ -547,18 +603,32 @@
...
+
         &--sm {
             gap: $space-xs;
         }
 
         &--md {
             gap: $space-m;
         }
+
+        &__articles {
```
stevenbister, line 627: The way this is being used in the markup suggests this should be a modifier rather than an element. Elements generally go inside of a block so should be used like this
```html
<div class="c-btn-group c-btn-group--row c-btn-group--none">
  <div class="c-btn-group__articles">
    <button>...</button>
  </div>
</div>
```

As it is modifying behaviour I would also name it something more descriptive as to what it's doing `--center` perhaps? So your markup would end up looking like
```html
<div class="c-btn-group c-btn-group--row c-btn-group--none c-btn-group--center">
    <button>...</button>
</div>
```

stevenbister (changes requested):

packages/osc-ecommerce/app/routes/blog/index.tsx
```diff
@@ -92,7 +111,23 @@ export default function Index() {
                     <PagePreview query={query} params={params} Component={PageContent} />
                 </PreviewSuspense>
             ) : (
-                <PageContent {...blog} />
+                <>
+                    <PageContent {...blog} />
+                    <div className="u-bg-color-gradient-primary-270">
```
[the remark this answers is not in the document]
jkellerman: I have updated this in the blog-posts PR, it can be reviewed there

packages/osc-ui/src/components/Button/button.scss
stevenbister: For the border?

jkellerman: For the button. In mobile the width of the button needs to be 100% so it fits the whole viewport. In desktop, if i set width to none, the button has no padding and shrinks. The only solution I know is to use unset here.

stevenbister: I'm not talking about the width. I'm talking about the `border-bottom`. Also does `width: 100%` not work?

jkellerman: No 100% doesn't work, the buttons stack in a column when doing this.

jkellerman: I've changed border-bottom to none

packages/osc-ecommerce/app/routes/blog/index.tsx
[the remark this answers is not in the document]
stevenbister: Okay that's fine. Just for future it would be better to do adjustments like that in PR it's related to as it makes reviewing difficult. This is the button group PR so creating the button group component in here would keep the PR's more modular. If that makes sense?

stevenbister (approved):

packages/osc-ecommerce/app/routes/blog/index.tsx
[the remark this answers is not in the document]
jkellerman: Yeah definitely, I know it must seem confusing and I'm aware that I could have handled this better. Reason why I did it was because I made a number of changes to how I structured the components in the next PR including the button group so I did all there instead. Will get in the habit of committing more often as well so things like this don't happen, I've noticed things can pile up real fast 😅.

PR #1404, opened 2023-08-14
Feature/ah forgot password

stevenbister:

- [ ] Close issues
- [x] Request reviewers
- [x] Assign yourself
- [x] Add project
- [x] Set project status
- [ ] Add milestone

---

N/a

## 📝 Description

Adds the forgotten password form and functionality.

N/a

## 🚀 New behaviour

- Updates password reset api to return access and refresh tokens
- Adds mutations to request and change password
- Adds `RequestResetPasswordForm` and `ChangePasswordForm`
- Adds action to `_auth.reset-password` route to request or change the password
- Adds loader to get token from url for use in main function
- Extracts auth form errors into `useAuthErrors` hook
- Extracts object from Zod to make the schema more reusable

## 🧾 Final Checklist

- [x] I've added tests to cover my changes (Unit, E2E, Integration, Contract, Functional, Acceptance etc)

## 📝 Additional Information

In order to test you will need the new planetscale `.env`s as well as all of the `.env`s for each api package. You will also need a user setting up if you don't have one already.
In order to run the app you will need to run all of the api services as well as the api-gateway which should run on `localhost:4000/graphql` you can then run the academic hub app.


mattwendzina (changes requested):
Looks good mate 👍 One issue I noticed when though - when you type an incorrect password you get a message 'No matching user found', rather than an incorrect password message:

Screenshot 2023-09-04 at 15 00 29

stevenbister:
> Looks good mate 👍 One issue I noticed when though - when you type an incorrect password you get a message 'No matching user found', rather than an incorrect password message:

Yeah that's expected, that message comes from the api rather than something on the front end. Perhaps something @Tom-Pearce or @hall2521 could advise on whether we can return a more specific error? 
Either way would be separate from this PR I think.

mattwendzina:
> > Looks good mate 👍 One issue I noticed when though - when you type an incorrect password you get a message 'No matching user found', rather than an incorrect password message:
> 
> Yeah that's expected, that message comes from the api rather than something on the front end. Perhaps something @Tom-Pearce or @hall2521 could advise on whether we can return a more specific error? Either way would be separate from this PR I think.

Okay fair enough. Doesn't seem like a great message if the user does exist... but like you said outside of scope here.

PR #1405, opened 2023-08-14
fix(product form): visually hide options rather than remove them if t…

stevenbister:
…here is only 1

- [x] Close issues
- [x] Request reviewers
- [x] Assign yourself
- [x] Add project
- [x] Set project status
- [ ] Add milestone

---

- Closes https://www.notion.so/openstudycollege/Course-costs-not-pulling-through-from-Shopify-to-Sanity-504631980b454d64b3b21d0c42dccdda?pvs=4

## 📝 Description

Prices for variants with only one study method/course option is not updating the price when the other is selected in the product form

image
image

Currently when there is only one option available the product form will not render the options. This is leading to a bug where the url params aren't including the correct options to match against the selected variant.

## 🚀 New behaviour

Rather than removing the option completely I have updated it to pass a hidden class, so the option still exists in the form but is just not visible to the user.

## 🧾 Final Checklist

## 📝 Additional Information


packages/osc-ecommerce/app/components/Forms/ProductForm/ProductForm.tsx
```diff
@@ -134,22 +134,21 @@ export const ProductForm = forwardRef<ElementRef<'div'>, ProductFormProps>(
...
-                                        }-${searchParamsWithDefaults.get(option.name)}`}
-                                        id={id}
-                                        option={option}
-                                        searchParamsWithDefaults={searchParamsWithDefaults}
-                                    />
-                                );
-                            }
-
-                            return <></>;
+                            // if (option.values.length >= 1) {
```
hi-im-si, line 137: remove commented code

packages/osc-ecommerce/app/components/Forms/ProductForm/ProductForm.tsx
```diff
@@ -134,22 +134,21 @@ export const ProductForm = forwardRef<ElementRef<'div'>, ProductFormProps>(
...
+                                    key={`${id}-${index}-${
+                                        option.name
+                                    }-${searchParamsWithDefaults.get(option.name)}`}
+                                    id={id}
+                                    option={option}
+                                    searchParamsWithDefaults={searchParamsWithDefaults}
+                                    className={option.values.length > 1 ? '' : 'u-hidden-from@all'}
+                                />
+                            );
+                            // }
```
hi-im-si, line 151: remove commented code

packages/osc-ecommerce/app/components/Forms/ProductForm/ProductForm.tsx
```diff
@@ -134,22 +134,21 @@ export const ProductForm = forwardRef<ElementRef<'div'>, ProductFormProps>(
...
+                                <Options
+                                    // Here we're passing the searchParamsWithDefaults as part of the key so the options remount when searchParams changes
+                                    // This is needed to keep the main form and the drawer forms in sync
+                                    key={`${id}-${index}-${
+                                        option.name
+                                    }-${searchParamsWithDefaults.get(option.name)}`}
+                                    id={id}
+                                    option={option}
+                                    searchParamsWithDefaults={searchParamsWithDefaults}
+                                    className={option.values.length > 1 ? '' : 'u-hidden-from@all'}
```
hi-im-si, line 148: i would imagine we would not need to create a class for this and simply apply the `hidden` attribute?
```suggestion
                                    hidden={option.values.length <= 1 ? true : undefined}
                                    
```

hi-im-si (changes requested):

packages/osc-ecommerce/app/components/Forms/ProductForm/ProductForm.tsx
stevenbister: Oops!

packages/osc-ecommerce/app/components/Forms/ProductForm/ProductForm.tsx
stevenbister: Yeah, that was my initial approach. But the inner component `RadioGroup` doesn't extend the html attributes for the `fieldset` `hidden` didn't work :/ I guess I could wrap the `Options` in a div and put it there but not keen on the extra markup for the sake of a `hidden` property :/

hi-im-si: ahh no worries makes sense.

stevenbister:
@hi-im-si Noticed I missed a bit of logic, so would still fail to update the cost if there is only one study pack e.g http://localhost:2000/courses/aat-accounting-level-2-and-3?Course+Options=Course+Material&Study+Options=Study+Pack. I've updated the condition to render the `Options` there too

PR #1416, opened 2023-08-16
Feature/blog posts

### LINKED ISSUE #1001: list all blog posts
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/1000

### Description

Posts to be pulled from the Blog type in Algolia

jkellerman:

- [x] Close issues
- [x] Request reviewers
- [x] Assign yourself
- [x] Add project
- [ ] Set project status
- [ ] Add milestone

---

- Closes #1001 

## 📝 Description

This branch adds blog posts for the blog index page. These posts include:

- An Island grid with the latest News post on the left card and the latest post for each other categories on the right cards
- A blog card for the latest Featured post
- A list of the latest blog cards that are not in the island grid or the latest Featured post

It uses Sanity groq queries to get the correct posts. Also a new 'featured' field has been added to posts in Sanity to declare whether a post is featured or not.

I have also created a new card paragraph component, which is used to help with the alignment of blog post cards in the grid for the latest posts. This may not be necessary, however I think it maintains a more clean and organised design.

A final note, the list of latest posts at the bottom of the blog page are currently being sliced to show the latest 4 posts. This is to be changed and will be use pagination, which I will complete in a new feature branch. 

N/A

## 🚀 New behaviour

N/A

## 🧾 Final Checklist

## 📝 Additional Information


packages/osc-ecommerce/app/types/sanity.ts
```diff
@@ -337,6 +343,7 @@ export interface heroSlide extends module {
 export interface heroModule extends module {
     carouselSettings: carouselModuleSettings;
     slides: heroSlide[];
+    body?: ModuleContent[];
```
stevenbister, line 346: I don't think we export the body directly on the hero do we? It lives inside the slides as far as I can tell.

If we do then I think the type would be better off as `contentModule`. That's what's being used in the slide content if not, then it should be `PortableTextBlock[];`

packages/osc-ecommerce/package.json
```diff
@@ -33,6 +33,7 @@
     "@radix-ui/react-slot": "^1.0.1",
     "@react-aria/i18n": "^3.7.0",
     "@react-aria/ssr": "^3.5.0",
+    "@react-aria/visually-hidden": "^3.8.3",
```
stevenbister, line 36: Like I said before we have our own visuallyhidden component in osc-ui. Can you uninstall this

packages/osc-ecommerce/app/routes/blog/index.tsx
```diff
@@ -127,6 +268,252 @@ export default function Index() {
...
+                                <CardImage>
+                                    <Image
+                                        src={getImage(newsData) as string}
+                                        alt={postCardData.image.alt}
+                                        width={postCardData.image.width}
+                                        height={postCardData.image.height}
+                                    />
+                                </CardImage>
+                                <CardInner>
+                                    <CardHeader>
```
stevenbister, line 283: I've noticed you've nested all of your card components within the card header. You'll need to move the body and the footer out for the card to behave properly. This is why everything looks misaligned. This is how they should look
image

packages/osc-ui/src/utils/truncate.ts
```diff
@@ -1,9 +1,18 @@
 /**
  * Truncate a string to a given length.
  */
-export const truncate = (str: string, maxLength: number = 343) => {
+export const truncate = (input: string | (() => string), maxLength: number = 343) => {
+    let str: string;
+
+    if (typeof input === 'function') {
```
stevenbister, line 7: We shouldn't be allowing functions to pass into this you can't run substring against anything but a string. Is there a reason for this?

stevenbister (changes requested):

packages/osc-ecommerce/app/routes/blog/index.tsx
```diff
@@ -98,10 +171,78 @@ export const meta: MetaFunction = ({ data, parentsData }) => {
...
+    };
+
+    const getImage = (data: postData) => {
+        const image = data.modules.find((item: heroModule) => item._type === 'module.hero');
+        if (image) {
+            return image?.slides[0].image?.src;
+        }
+    };
+
+    const getFeaturePostBgColour = (data: postData) => {
```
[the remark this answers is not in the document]
jkellerman: I looked into this and it appears that blogCard component in ecommerce is being used in the instantsearch, meaning the data structure of the card is geared towards algolia. I'm not sure if there is an easy way to tweak it so that it works with the content we get from sanitys so what i've done instead is create new components to break it all down, as mentioned in your below comments.

I've also created some utility functions which i've placed in utils folder for cards to get some of the data as they are used across multiple components. Couldn't think of another approach to this, taking into consideration how the data comes in from sanity.

packages/osc-ui/src/utils/truncate.ts
jkellerman: Yeah, I noticed that doing this was unnecessary, I've set it back to how it was and works fine.

packages/osc-ecommerce/app/routes/blog/index.tsx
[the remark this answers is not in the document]
stevenbister: There's one that's in `components/Cards/BlogCard.tsx` I'm pretty sure that one is set up for Sanity data?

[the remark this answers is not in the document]
stevenbister: Okay looking at that component a bit more I think it can be used in place of your featured card. You'll just need to pass through a few props such as `data.fullwidth` on line 39.

Could we tweak this component to handle different card variants maybe? I'm conscious that we are duplicating a lot of the cards and functionality within them where we might not need to?

packages/osc-ecommerce/app/components/Button/ButtonGroup.tsx
```diff
@@ -0,0 +1,26 @@
+import { ButtonGroup as BlogPostsButtons, Button } from 'osc-ui';
+import type { categoryModule } from '~/types/sanity';
+
+interface ButtonGroupProps {
+    data: categoryModule[];
```
stevenbister, line 5: Are we certain we'll always get data back? Might be worth making this an optional prop and setting a check in the component to handle null/undefined responses. Otherwise this component will throw an error when we release to prod without any categories set up

packages/osc-ecommerce/app/components/Cards/FeaturedCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+    CardTitle,
+    Image,
+    VisuallyHidden,
+} from 'osc-ui';
+import { formatDate } from 'osc-ui/src/utils/formatDate';
+import type { heroModule, postHeroModule } from '~/types/sanity';
+import { getAltText, getImage } from './utils';
+
+interface FeaturedCardProps {
+    data: postHeroModule;
```
stevenbister, line 18: Same as before I would make this optional and set some checks below for if it returns a falsey value. Likely in a CMS that the data could not be available

packages/osc-ecommerce/app/components/Cards/FeaturedCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+    };
+    const getFeaturePostTitleColour = (data: postHeroModule) => {
+        const heroModule = data.modules.find((item: heroModule) => item._type === 'module.hero');
+        if (heroModule) {
+            return heroModule?.slides[0].titleColor;
+        }
+    };
+    return (
+        <>
+            <BlogCard
```
stevenbister, line 38: Can we add the `blockLink` prop to this card so the whole thing is clickable? We're using it on the other places this card is usable so keeps things consistent

packages/osc-ecommerce/app/components/Cards/FeaturedCard.tsx
stevenbister: I'm not gonna mention this on the other cards but worth doing it there as well

packages/osc-ecommerce/app/components/Cards/utils.tsx
```diff
@@ -0,0 +1,29 @@
+import type { ModuleContent, heroModule, postContentModule, postHeroModule } from '~/types/sanity';
+
+export const getModuleContentBody = (data: postContentModule) => {
+    const contentModule = data.modules.find(
+        (item: ModuleContent) => item._type === 'module.content'
+    );
+
+    const body = contentModule?.body[0];
```
stevenbister, line 8: Getting a ts error 
```
Property 'body' does not exist on type 'ModuleContent'.
```

packages/osc-ecommerce/app/components/Cards/utils.tsx
```diff
@@ -0,0 +1,29 @@
+import type { ModuleContent, heroModule, postContentModule, postHeroModule } from '~/types/sanity';
+
+export const getModuleContentBody = (data: postContentModule) => {
+    const contentModule = data.modules.find(
```
stevenbister, line 4: Worth checking if data exists on all of these utils?

packages/osc-ecommerce/app/components/Cards/IslandGridCard.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { formatDate } from 'osc-ui/src/utils/formatDate';
+import type { postContentModule, postHeroModule } from '~/types/sanity';
+import { getAltText, getImage, getModuleContentBody } from './utils';
+
+interface IslandGridCardProps {
+    data: postHeroModule | postContentModule;
+    variant?: 'media-object';
+    isRounded?: boolean;
+    body?: boolean;
+    date?: boolean;
```
stevenbister, line 23: As these are booleans it would be better to prefix the prop with a verb such as `has` or `is` so in this case it would be `hasBody` or `hasDate`

packages/osc-ecommerce/app/routes/blog/index.tsx
```diff
@@ -99,33 +158,53 @@ export const meta: MetaFunction = ({ data, parentsData }) => {
...
-                                    className="u-color-tertiary"
-                                    size="lg"
-                                >
-                                    {category.title}
-                                </Button>
-                            ))}
-                        </ButtonGroup>
+                    <ButtonGroup data={categories} />
+
+                    <div className="o-container">
```
stevenbister, line 195: Can you have a copy of this in `PreviewSuspense` as well? 
For reference the preview suspense is for previewing the page before it's published so we want to mimic as much of is as possible

stevenbister (changes requested):

packages/osc-ecommerce/app/routes/blog/index.tsx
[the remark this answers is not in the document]
jkellerman: Okay, maybe you could clarify something for me. That BlogCard component that we already have is being used in our modules and is currently fetching the 3 popular posts at the top of the blogs page. If you look at the data it grabs everything in the 'reference' property. There is no reference property when querying these posts so isn't the best way to just create a separate component for these posts?

packages/osc-ecommerce/app/components/Cards/FeaturedCard.tsx
jkellerman: For featured card if it returns a falsey value, what do you think we should have as placeholder, just some text or some dummy content?

packages/osc-ecommerce/app/components/Button/ButtonGroup.tsx
stevenbister: Yeah I think it's worth doing. In a CMS with users you can never be sure they won't remove them for whatever reason so it's a good safety net

packages/osc-ecommerce/app/components/Cards/FeaturedCard.tsx
stevenbister: I think we just return null, if there's no data then the card shouldn't render

packages/osc-ecommerce/app/routes/blog/index.tsx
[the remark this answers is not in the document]
stevenbister: I think we could just check for the existence of the reference and change the values based on that? If you have a look at the `CourseCard` it's set to handle data from Sanity or Shopify by checking for the `store` property.
e.g.
```tsx
const { product } = props;
const store = product?.reference?.store;

if (!store && !product) return null;

const title = store?.title ?? product.title;
const slug = store?.slug?.current ?? product.handle;
```

I might be oversimplifying but could we do something similar here?

packages/osc-ecommerce/app/components/Cards/IslandGridCard.tsx
stevenbister: Don't forget to update the places these props are used, looks like these are still being used in `PostsGrid` etc.

packages/osc-ecommerce/app/routes/blog/index.tsx
stevenbister: I think your best bet would be to add `grid-template-rows: 250px 1fr` to the `c-card--blog` class looks like it does a good job of tightening this up

image

packages/osc-ecommerce/app/routes/blog/index.tsx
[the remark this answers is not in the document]
jkellerman: As discussed, there are quite a differences between featured card and the blogCard component, which would require a lot of tweaking and could affect some other components elsewhere so will keep these separate.

packages/osc-ecommerce/app/routes/blog/index.tsx
```diff
@@ -49,9 +101,35 @@ export const loader: LoaderFunction = async ({ request, params }) => {
...
 
+    const defaultCategory = {
+        title: 'All Articles',
+        _id: 'defaultCategory',
+    };
+
+    const categories = [defaultCategory, ...categoriesData];
+
+    const idsToRemoveFromAllPosts = [
+        newsData._id,
```
stevenbister, line 112: Noticed these don't exist in the production dataset atm so causing the app to crash if you have `SANITY_STUDIO_API_DATASET="production"` set in you env

```
TypeError: Cannot read properties of undefined (reading '_id')
    at loader14 (/Users/stevebister/SITES/osc/packages/osc-ecommerce/build/index.js:11560:14)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.callRouteLoaderRR (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/server-runtime/dist/data.js:52:16)
    at async callLoaderOrAction (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/router/router.ts:3626:16)
    at async Promise.all (index 0)
    at async loadRouteData (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/router/router.ts:2954:19)
    at async queryImpl (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/router/router.ts:2735:20)
    at async Object.queryRoute (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/router/router.ts:2675:18)
    at async handleDataRequestRR (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/server-runtime/dist/server.js:75:20)
    at async requestHandler (/Users/stevebister/SITES/osc/node_modules/@remix-run/node/node_modules/@remix-run/server-runtime/dist/server.js:49:18)
```

Probably worth adding some optional chaining to prevent this

packages/osc-ecommerce/app/components/Cards/IslandGridCard.tsx
```diff
@@ -0,0 +1,94 @@
...
+    isRounded?: boolean;
+    hasBody?: boolean;
+    hasDate?: boolean;
+}
+
+export const IslandGridCard = (props: IslandGridCardProps) => {
+    const { data, variant, isRounded, hasBody, hasDate } = props;
+
+    if (!data) {
+        return (
```
stevenbister, line 30: I think it would be better to return `null` rather than a message like this in case the live site has any issues. We probably don't want to display an error like this to customers

packages/osc-ecommerce/app/components/Button/ButtonGroup.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export const ButtonGroup = (props: ButtonGroupProps) => {
+    const { data } = props;
+
+    if (!data) {
+        return (
```
stevenbister, line 12: I think it would be better to return null rather than a message like this in case the live site has any issues. We probably don't want to display an error like this to customers

packages/osc-ecommerce/app/components/Cards/FeaturedCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+    };
+    const getFeaturePostTitleColour = (data: postHeroModule) => {
+        const heroModule = data.modules.find((item: heroModule) => item._type === 'module.hero');
+        if (heroModule) {
+            return heroModule?.slides[0].titleColor;
+        }
+    };
+
+    if (!data) {
+        return (
```
stevenbister, line 38: I think it would be better to return null rather than a message like this in case the live site has any issues. We probably don't want to display an error like this to customers

packages/osc-ui/src/components/Card/card.scss
```diff
@@ -294,6 +294,7 @@ $card-fs-3xl: $fs-3xl;
         &--blog {
             border-bottom: var(--card-border);
             color: var(--color-secondary);
+            grid-template-rows: 250px 1fr;
```
stevenbister, line 297: This will need to be wrapped in a media query, the image looks too big on mobile

image

Should look like

image

stevenbister (changes requested):

packages/osc-ecommerce/app/routes/blog/index.tsx
```diff
@@ -49,9 +101,39 @@ export const loader: LoaderFunction = async ({ request, params }) => {
...
+    const defaultCategory = {
+        title: 'All Articles',
+        _id: 'defaultCategory',
+    };
+
+    const categories = [defaultCategory, ...categoriesData];
+
+    let filteredAllPostsData;
+
+    if (newsData && careersAdviceData && studyTipsData && successStoriesData && featuredPostData) {
```
stevenbister, line 113: I see what you're trying to do here but ideally we'd want to have the option for posts to show if some of these options are missing. 
At the moment lets say Marketing decide not to feature any posts; then no posts will show at all.

Like I said in my last comment you could use optional chaining to handle this e.g
```ts
const idsToRemoveFromAllPosts = [
        newsData?._id,
        careersAdviceData?._id,
        studyTipsData?._id,
        successStoriesData?._id,
        featuredPostData?._id,
    ];
```
Or if that doesn't quite work/you don't like that this would also work
```ts
const idsToRemoveFromAllPosts = [
        newsData && newsData._id,
        careersAdviceData && careersAdviceData._id,
        studyTipsData && studyTipsData._id,
        successStoriesData && successStoriesData._id,
        featuredPostData && featuredPostData._id,
    ];
```
This way if one of these is missing then the others should still populate the array

stevenbister (changes requested):
Nice work man, one last thing and I think we're good

packages/osc-ecommerce/app/routes/blog/index.tsx
stevenbister: Oh just to follow on from this, I find it helpful when I'm creating things like this to switch between `staging` and `production` datasets so I can see the differences and possibly catch things we may have in set up in development but aren't in prod yet.

jkellerman: Good point thanks, have done this now.

Also thanks for the tip, on switching between staging or prod, very helpful.

PR #1455, opened 2023-08-21
Feature/blog pagination

### LINKED ISSUE #1003: Article filter

jkellerman:

- [ ] Close issues
- [ ] Request reviewers
- [ ] Assign yourself
- [ ] Add project
- [ ] Set project status
- [ ] Add milestone

---

- Closes #1003 

## 📝 Description
This adds the pagination the blogs page. On initial load, latest posts, featured post and all posts should be displayed. When the user clicks on any of the categories, they should be navigated to a new query param and all posts for that category should be shown.

Unfortunately, we can't use pagination in a groq query as Sanity doesn't seem to allow us to sort the truncated lists by date created (only by id). I have been contact in someone in the Sanity slack channel, however they do not seem to have a solution. Will keep an eye out for any updates. As it stands, paginated data fetches all the posts and then we're using the slice method to create pagination.

N/A

## 🚀 New behaviour

N/A

There seems to be issues with the flourishes in the hero section, they flash on inital render and then disappear. This is something that needs looking at.

## 🧾 Final Checklist

## 📝 Additional Information


packages/osc-ecommerce/app/routes/blog/index.tsx
```diff
@@ -58,28 +54,65 @@ export const loader: LoaderFunction = async ({ request, params }) => {
...
+
+        // removes latests posts from the paginated allposts list
+        const filteredAllPosts = allPosts.filter(
+            (post: postCategoryData) => !idsToRemoveFromAllPosts.includes(post._id)
+        );
+
+        allPosts = filteredAllPosts;
+    }
 
     if (!data?.page) {
```
mattwendzina, line 117: I'd probably move this to just after the query for the data so that it doesn't have to un-necssarily run more queries if the page data doesn't exist.

mattwendzina:
I think you need to use optimistic UI on this as it can take quite a bit of time for the page to reload when you change category, and this would be worse on a slow connection. Should be pretty straight forward to add in. If you have a look at the `blog-posts-v2` branch I created it's got it in there - just uses the `useNavigation()` hook to check for the loading state.

packages/osc-ecommerce/app/routes/blog/index.tsx
```diff
@@ -164,19 +186,55 @@ export default function Index() {
...
-    // TODO: Use pagination instead of slicing to fetch all posts
-    const allPostsData = filteredAllPostsData.slice(0, 4);
+    useEffect(() => {
+        setNumberLoaded(
+            Math.min(postsPerPage, allPosts ? allPosts.length : categoryPosts.page.length)
+        );
+    }, [allPosts, categoryPosts, postsPerPage]);
+
+    const allPostsListTotal = allPosts && allPosts.length;
+    const allPostsList = allPosts && allPosts.slice(0, numberLoaded);
```
stevenbister, line 211: Did you manage to get anywhere with filtering on the query? Loading all of the posts then filtering on the front end doesn't sit right with me. If we have a lot of posts to load then this could slow the page down quite significantly

packages/osc-ecommerce/app/routes/blog/index.tsx
```diff
@@ -185,26 +243,43 @@ export default function Index() {
...
-                            studyTipsData={studyTipsData}
-                            successStoriesData={successStoriesData}
-                        />
-
-                        {/* Featured Card */}
-                        <FeaturedCard data={featuredPostData} />
-
-                        {/* all posts */}
-                        <PostsGrid data={allPostsData} />
+                        {!categoryPosts && (
```
stevenbister, line 254: Should probably add an eslint rule but it's better to use ternaries than the `&&` in jsx. See https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

stevenbister (changes requested):

packages/osc-ecommerce/app/routes/blog/index.tsx
jkellerman: Yeah, I've spoken to Si and Matt about this whilst you we're away. I went back and forth with someone in the sanity slack channel who came back with the same solution I tried. Pagination works on the query but it doesn't allow us to sort them by date created only by id. I was passed on to the groq channel in sanity to get help with this but I haven't received anything back yet. 

Si said to do it this way for the time being so we're not waiting on them for an answer. I know it's not best practice but at the moment we're only dealing with a few hundred posts.

packages/osc-ecommerce/app/routes/blog/index.tsx
```diff
@@ -39,6 +66,64 @@ export const loader: LoaderFunction = async ({ request, params }) => {
...
+        latestStudyTipPost = await getSettingsData({
+            query: LATEST_STUDY_TIPS_POST,
+        });
+        latestSuccessStoriesPost = await getSettingsData({
+            query: LATEST_SUCCESS_STORIES_POST,
+        });
+        latestFeaturedPost = await getSettingsData({
+            query: LATEST_FEATURED_POST,
+        });
+        allPosts = await getSettingsData({
```
stevenbister, line 104: Have we considered wrapping these in a promise.all? Running them all in parallel might help with any loading speed issues 🤔 Unless each query is dependant on another?

stevenbister (changes requested):
This branch is a bit behind now so not sure how much has been sorted elsewhere but a few comments from the Rabbit that I agree with, so worth going through those.

Also the categories bar is left aligned, I think it should be centered?
image

stevenbister:
Closing as replaced by #1643

PR #1530, opened 2023-08-30
Feature/bloom connections feedback

jkellerman:

---

- Closes #

## 📝 Description

This feature branch makes a few changes to the bloom connections website based on feedback from marketing. These include:

1. Changing the background colour from yellow to off white
2. Changing all buttons on the website to have rounded edges
3. Gives CMS the option to select different button variants.
4. Creating a placeholder image for testimonial cards
5. Change the hover menu underline and other underlines on the website to fit the handwritten underline that is included in the brand guidelines

List of checks for approval:
1. The background colour is now an off white colour
2. Buttons on all pages have rounded edges
3. You are able to select from five different button variants in the CMS, all should have different colours. There should be no colours coming from ecommerce i.e. no gradient variants (some buttons are using capitals, some are not, will wait on feedback from marketing with which they prefer).
4. A placeholder image should be present if there are no images selected  in the CMS for testimonial cards.
5. The underlines in the footer and hover state in nav menu should a wavy underline using svg. 

N/A

## 🚀 New behaviour

N/A

## 🧾 Final Checklist

## 📝 Additional Information


packages/osc-recruitment/app/components/Nav.tsx
```diff
@@ -93,9 +94,10 @@ const RecursiveNavItemWrapper = (props: Props) => {
                 <NavLink
                     href={item?.internalLink?.slug ?? item?.externalLink}
                     isExternal={item?.target === 'External'}
-                    className="u-bc-secondary"
+                    className="u-bc-secondary is-no-divider"
```
stevenbister, line 97: Presumably this is removing the divider styles? If so do we need to include the `Divider` on line 100?

Using negatives as naming can be difficult to follow, and at the moment your css file is a double negative which makes it harder still `:not(.is-not-divider)`

It would be better to flip this and have something more like `has-divider` and then only set the required styles on that class and remove the `:not()`

packages/osc-ui/src/components/Navbar/navbar.scss
```diff
@@ -183,18 +183,21 @@
...
-                            linear-gradient(
-                                0deg,
-                                var(--color-secondary),
-                                var(--color-secondary)
-                            )
-                            no-repeat right bottom / 0 $bg-height;
                         line-height: $base-lh;
-                        transition: background-size $timing $ease-in-out;
+
+                        &:not(.is-no-divider) {
```
stevenbister, line 188: See my comment above about double negatives

packages/osc-ui/src/components/Select/select.scss
```diff
@@ -43,6 +43,10 @@
                 pointer-events: none;
             }
 
+            &.is-pill {
+                border-radius: 15px;
```
stevenbister, line 47: Would one of our `$rad` variables work here?

stevenbister (changes requested):
Looks good! Couple of comments, hopefully fairly minor

packages/osc-recruitment/app/components/Nav.tsx
jkellerman: I can see how this could be confusing with the double negative. Went through this with Si yesterday, I think we need the :not() pseudo class because we're using the navlink from ecommerce and nesting in the new divider component for Bloom. If we don't have a class on the navlink component for Bloom that indicates it has a divider then we will get the ecommerce underline on top the divider component that we need.

I'm thinking the best solution is to keep it as it is but change the name so could have `has-divider` instead of `is-not-divider`.

Issue #1008, opened 2023-06-08
Add Urgency Indicators


EnaamAlum:
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/1006

### Description

- [x] Header banner #1584 
- [x] Banners can be placed on any page #1601 
- [ ] Countdown/messaging on product pages/collections

PR #1601, opened 2023-09-08
Feature/discount banners pages

stevenbister:

- [ ] Close issues
- [x] Request reviewers
- [x] Assign yourself
- [x] Add project
- [x] Set project status
- [x] Add milestone

---

Task two on #1008 

## 📝 Description

Adds the ability in Sanity to add the discount banners to pages other than the product page
image

This will default to just the product page being selected 
image

And will allow users to globally set which page type this banner appears on. Except for the product page I've added this as part of the `Hero` component so will sit below that when it is active.
image

This could be more flexible if we set it up as a module that can be picked from the modules array, allowing the banner to be placed anywhere on the page, however this would mean the content editor would have to set this on every page manually. I think this way, although it is more restrictive on design, allows for a better UX.

N/a

## 🚀 New behaviour

- Adds checkbox list to allow the discount banner to be set on any page type
- Adds a utility function to work out whether the route id matches the passed value
- Updates callout banner component to have a full width prop

## 🧾 Final Checklist

## 📝 Additional Information

### Summary by CodeRabbit

- New Feature: Added a new prop `isFull` to the `DiscountBanner` and `CalloutBanner` components, allowing these banners to fill the width of their container.
- New Feature: Introduced a new field `showOnPages` in the `DISCOUNT_TAG_QUERY` GraphQL query and `SanityDiscountBanner` interface, enabling specification of pages where the discount banner should be displayed.
- New Feature: Implemented a new function `showBannerOnPage` that determines whether a discount banner should be shown on a specific page based on the `showOnPages` property of the banner.
- New Feature: Created a new module `DiscountBannerModule` that renders a `DiscountBanner` component based on the closest discount and current route ID.
- Refactor: Updated various components and routes to use the new `showBannerOnPage` function and `isFull` prop for better control over banner display and layout.


hi-im-si (changes requested):
Happy to have a conversation on this one but i personally feel the long strip banner is going back to the way the old site worked where everything is uniform and bar-like. I feel the banners would be more visual constraining to a container e.g.

image

Happy to discuss! I may share this with Marketing for their thoughts.

++++

On top of the above, it looks like there may need to be some spacing added to the callout as any regions below are squeezed together:
image

stevenbister:
> @stevenbister i do agree this would be better as a module. Would it be something which literally just feeds the active discount. I am thinking it just needs to be as is but allow the callout to be moved around the page and then pull in whatever is the active discount for the ticked page. I think course pages will be rigid but be good to implement this as more dynamic on the standard pages. i think marketing will need to do the leg work anyway. is this a big change @stevenbister ?

Nope :) I've updated as per our convo on Discord. So now there's a `DiscountBanner` module in the modules list
image

And that just pulls the latest banner onto the page wherever that is positioned. Product and collection pages then have it always fixed at the top

hi-im-si (approved):
🚀

PR #1608, opened 2023-09-11
Bugfix/collection page colouring

### LINKED ISSUE #1571: Collection page - colouring issue
### Link task to story

Story: https://github.com/Open-Study-College/osc/issues/1564

### Description

- retrieving the colour from the collections page and passing as a colour variable on to the elements.
- See Steven for guidance

jkellerman:

- [x] Close issues
- [x] Request reviewers
- [x] Assign yourself
- [x] Add project
- [x] Set project status
- [x] Add milestone

---

- Closes #1571 

## 📝 Description

Adds theme for collection pages.

The collections page colours currently default to primary colours.

## 🚀 New behaviour

The theme for collections page can now be changed in the CMS. The theme is applied to:

- Select bar
- Text and callout within the course card
- Pagination progress bar
- Slider range and slider thumb
- Checkbox

## 🧾 Final Checklist

## 📝 Additional Information

### Summary by CodeRabbit

- New Feature: Introduced `primaryTheme` and `secondaryTheme` props to various components including `InstantSearchCourseCard`, `Search`, `Hit`, `InfiniteHits`, `RefinementList`, `RefinementSlider`, and `SelectBar`. These new properties allow for dynamic theme customization across the application.
- New Feature: Updated `Checkbox` and `Slider` components in `osc-ui` package to accept a `theme` prop for custom styling based on different themes.
- Refactor: Removed hardcoded color values from several components and replaced them with dynamic theme-based colors. This change enhances the flexibility of these components by allowing customization of their appearance based on different themes.
- Chore: Updated queries and types to support the new `primaryTheme` and `secondaryTheme` properties.


packages/osc-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -57,7 +57,7 @@ export interface CheckboxProps extends ComponentPropsWithRef<typeof CheckboxPrim
     /**
      * Sets the custom styles, e.g. "Secondary", "Tertiary"
      */
-    variants?: Variants[];
+    variants?: Variants[] | string;
```
stevenbister, line 60: Is there a reason for updating this type and not working within what's already there?
This loosens and makes handling the type trickier as now we'll have to account for strings or arrays.

packages/osc-ui/src/components/Slider/Slider.tsx
```diff
@@ -35,7 +35,7 @@ export interface SliderProps extends ComponentPropsWithRef<typeof SliderPrimitiv
     /**
      * Sets the custom styles, e.g. "Secondary", "Tertiary"
      */
-    variants?: string[];
+    variants?: string[] | string;
```
stevenbister, line 38: Similar to my previous comment

packages/osc-ui/src/components/Slider/Slider.tsx
```diff
@@ -61,10 +61,14 @@ export const Slider = forwardRef<ElementRef<typeof SliderPrimitive.Root>, Slider
...
                 <SliderPrimitive.Track className="c-slider__track">
-                    <SliderPrimitive.Range className="c-slider__range" />
+                    <SliderPrimitive.Range className={`c-slider__range u-bg-color-${variants}`} />
                 </SliderPrimitive.Track>
                 {sliderValues?.map((_, i) => (
-                    <SliderPrimitive.SliderThumb className="c-slider__thumb" key={i}>
+                    <SliderPrimitive.SliderThumb
+                        className={`c-slider__thumb`}
+                        key={i}
+                        style={{ ['--thumb-theme' as string]: `var(--color-${variants})` }}
```
stevenbister, line 70: Rather than settings this directly on the thumb I think it would be better to lift this up to the `SliderPrimitive.Slider` and call it something like `--slider-theme-color` and we can use it for both the colour of the thumb and the track then. Keeps it all in one place if we ever need to update anything in the future

stevenbister (changes requested):
Looks good mate.

One thing I'd suggest changing is keeping the theming in line with how it works on the product pages for consistency.

Looking at the product form I created a hook called `usePageTheme` which spits out a pair of colours (gradient and related singular colour) e.g.
```tsx
const [primaryThemeColor, secondaryThemeColor] = usePageTheme();
```
If you use that instead you could remove the need for setting the text colour in Sanity and keep the content entry down (as marketing will need to go back into each collection now to set the value -- which i know they won't be keen on).

packages/osc-ui/src/components/Checkbox/Checkbox.tsx
jkellerman: TypeScript was complaining at me when I was passing variants as a string to the className I think, can't remember now but it's not doing it anymore, I've set it back to how it was.

packages/osc-ecommerce/app/components/InstantSearch/Components/Cards/InstantSearchCourseCard.tsx
```diff
@@ -140,7 +154,14 @@ export const InstantSearchCourseCard = (props: CourseCardProps) => {
                             />
                         </span>
                     </CardTitle>
-                    <CardTitle as="h3" subtitle isSmall isThemeable position="bottom">
+                    <CardTitle
+                        className={`u-color-${secondaryTheme ? secondaryTheme : 'primary'}`}
```
stevenbister, line 158: I think this needs to fallback to a gradient if no colour is set, they're coming through black/navy on the search page atm
image

Should look like this according to figma
image

stevenbister (changes requested):
Looking great! One more thing and I think we're there

packages/osc-ecommerce/app/components/InstantSearch/Components/Cards/InstantSearchCourseCard.tsx
jkellerman: Good spot! I've also noticed that the course pop over won't have a theme either and will default to white, wonder if it is worth have a fallback gradient colour for this too, or just keep it as is?

stevenbister: Yeah good shout, probs worth having a fallback for that as well

stevenbister (approved):

stevenbister (changes requested):
Getting this error when navigating to collections pages

image

Looks to be related to something with Trustpilot, not sure if this is the same issue you mentioned earlier?

jkellerman:
@stevenbister 

> Getting this error when navigating to collections pages
> 
> image
> 
> Looks to be related to something with Trustpilot, not sure if this is the same issue you mentioned earlier?

Yeah I think so, if this is the case then it must be something to do with main branch?

stevenbister:
Lol @jkellerman You edited my comment so I missed this 😂 But yeah that looks like it sorted it

jkellerman:
> Lol @jkellerman You edited my comment so I missed this 😂 But yeah that looks like it sorted it

Oops sorry haha

stevenbister (approved):

PR #1661, opened 2023-09-27
Feature/back to top button v2

### LINKED ISSUE #1298: Build "Back to Top" button
### Link task to story

Story: 

### Description

Should be sticky to the sign up form.
image

jkellerman:

- [ ] Close issues
- [ ] Request reviewers
- [ ] Assign yourself
- [ ] Add project
- [ ] Set project status
- [ ] Add milestone

---

- Closes #1298 

## 📝 Description

N/A

> Please describe the current behaviour that you are modifying.

N?A

> Please describe the behaviour or changes this PR adds.

## 🧾 Final Checklist

## 📝 Additional Information


packages/osc-ui/src/components/Button/button.scss
```diff
@@ -223,6 +223,34 @@
                 )
             )
         ),
+        octonary: (
```
stevenbister, line 226: I'm not super keen on having another variant for the buttons if we can help it. I think we'd be better using the primary button variant and adding some modifiers to remove the pseudo element and setting the position.

packages/osc-ui/src/components/Button/button.scss
```diff
@@ -444,6 +472,39 @@
                     }
                 }
 
+                @if $class == "octonary" {
+                    position: absolute;
+                    top: -6.75rem;
+                    left: 0;
+                    max-width: 244px;
+
+                    @include mq($mq-mob-m) {
```
stevenbister, line 481: I think there's a mistake in the design file and having the gap between the button and the footer looks a bit weird. I reckon it should be attached to the footer like this
image

There's also a lot of media queries here, I don't think you need all of these 🤔 Just set the top to be a negative of the height of the button e.g. `top: -46px;` or whatever it needs to be

packages/osc-ui/src/styles/components/_forms.scss
```diff
@@ -15,9 +15,17 @@
...
                 justify-content: center;
+                padding-block: var(--space-scale-7xl);
                 background: var(--color-gradient-secondary-270);
                 box-shadow: none;
                 color: var(--color-tertiary);
+
+                @include mq($mq-desk) {
+                    .c-btn {
+                        align-self: flex-start;
+                        max-width: 135px;
```
stevenbister, line 26: I don't think we need this? The difference in size is minimal if it's not there

packages/osc-ecommerce/app/components/Footer/Footer.tsx
```diff
@@ -33,19 +34,46 @@ export const SiteFooter = (props: SiteFooterProps) => {
     // ! Once solution in place we can update this to use design token files instead
     const mq = {
         tab: 768,
+        deskMed: 1280,
```
stevenbister, line 37: As we're adding a new option here it would be a good opportunity to replace this with design tokens

`import { mediaQueries as mq } from 'osc-design-tokens'`

stevenbister (changes requested):

packages/osc-ui/src/styles/components/_forms.scss
jkellerman: This is to follow the design for the  signup button in the newsletter, something Si brought up. It prevents the button from spanning across the whole newsletter.

packages/osc-ui/src/components/Button/button.scss
jkellerman: I tried to avoid the media queries but I was following the design and the button is within the page margin (o-container)so because of this, the button travels vertically as the screen sizes change hence the several media queries I used to change the position.

If you guys are happy to have the button attached to the footer and at the edge of the page then this shouldn't be a problem and I think it looks fine.

packages/osc-ecommerce/app/components/Footer/Footer.tsx
jkellerman: Can we set this as another task? Got quite a bit to get through before I go away.

stevenbister: I don't think it needs to be an entirely separate task, it should only take a minute. It's just taking this object out and putting that import, I'm not asking for it to happen in every file just this one. You might have to change your use of deskMed on line 42 to desk-med but otherwise it shouldn't need any extra work

packages/osc-ui/src/styles/components/_forms.scss
stevenbister: That't what the align-self is doing. If you disable the max width it won't span the full width

packages/osc-ui/src/components/Button/button.scss
stevenbister: The right property should keep it within the container, but yeah fixing it to the top of the footer looks better imo

packages/osc-ecommerce/app/components/Footer/Footer.tsx
```diff
@@ -31,21 +33,43 @@ export const SiteFooter = (props: SiteFooterProps) => {
...
+    const [showOnGreaterThanDeskMed, setShowOnGreaterThanDeskMed] = useState(false);
 
     // We need this useEffect to set the showOnTab state only when the window object exists
     // Otherwise we will receive an SSR warning telling us the markup differs from the server
     useEffect(() => {
         setShowOnGreaterThanTab(isGreaterThanTab);
     }, [isGreaterThanTab]);
 
+    useEffect(() => {
+        setShowOnGreaterThanDeskMed(isGreaterThanDeskMed);
```
stevenbister, line 49: I don't think we need this? As far as I can tell it's not doing anything?

packages/osc-ui/src/components/Button/button.scss
```diff
@@ -718,6 +718,26 @@
         }
     }
 
+    .back-to-top {
```
stevenbister, line 721: If we're taking this approach this needs to be a modifier class at the moment this doesn't match up with BEMIT methodology.

Along these lines, I think we should be using the primary variant, rather than the primary-gradient and using this modifier to remove the pseudo element etc. to keep things more consistent. 

The gradient variants are designed to fit in with Ohana more than ecommerce, so the hover colours etc. will be different. E.g. you're hover colour isn't being applied due to the specificity of the gradient variants
image

stevenbister (changes requested):

packages/osc-ecommerce/app/components/Footer/Footer.tsx
jkellerman: Yeah you're right, it was to make the button smaller in mobile but the difference isn't enough for this to be worth doing, I will remove.

stevenbister (approved):
Looks great! :D

